Issue #387, opened 2014-06-29
Why metadata is not set on create?


sysout:
I am using the method provided by issue #325 
![screen shot 2014-06-28 at 10 33 56 pm](https://cloud.githubusercontent.com/assets/877905/3422382/10cbbd10-ff35-11e3-90ee-e424290513e3.png)
I add a parent_id metadata to solve has_many association problem. Only PackageContent will store parent_id.

```
class PackageContent < ActiveRecord::Base
  has_paper_trail :meta => { :parent_id  => :package_id}
end
```

But look at Row 12, parent_id is not set.

Issue #385, opened 2014-06-20
Paper_trail creates a first version with the custom meta attribute `approved` set to false even when it should be true


duarme:
The Content model has a bunch of attributes, but one of them is `approved` defined by this code in its migration:

```
add_column :contents, :approved, :boolean, null: false, default: false
```

And in its ContentVersion migration:

```
class CreateContentVersions < ActiveRecord::Migration
  def self.up
    create_table :content_versions do |t|
      t.string   :item_type, :null => false
      t.integer  :item_id,   :null => false
      t.string   :event,     :null => false
      t.boolean  :approved, null: false, default: false # <= this one
      t.string   :whodunnit
      t.text     :object
      t.text     :object_changes
      t.datetime :created_at
    end
    add_index :content_versions, [:item_type, :item_id]
  end
# ...
```

Of course, Content also has paper_trail:

```
# content.rb
has_paper_trail class_name: 'ContentVersion', meta: { approved: :approved }
```

This code creates and saves a content object:

```
content = Content.create!({
      author_id: user.id,
      content: Lorem.sentence,
      approved: true
    })
```

I expect it's first version to have `approved = true`, but, checking it via rails console it's easy to find out that `content.versions.first.approved` is `false`, while it should be `true`:

```
#!
- !ruby/object:ContentVersion
  attributes:
    id: 447
    item_type: Measure
    item_id: 42
    event: create
    approved: false
    whodunnit: '1'
    object: 
    object_changes: "--- !ruby/hash:[...]"
    created_at: 2014-06-20 10:08:59.263012000 Z
```

Why is that?

giangnguyennet:
I got confused too. At first I thought it was a bug that made the meta value always 1 step behind the current state. But now I understand that truly how it works. Thanks.

Issue #185, opened 2012-12-24
Can not store meta data on create operation, but creating being update operation in model.


rokibulhassan:
I have something like following in my model but data not set to hotel_id and agent_user_id while creating the object. Can not figure it out. I have hotel_id and agent_user_id in my version table. I also use `attr_accessible :hotel_id, :agent_user_id` in paper_trail.rb

``` ruby
class Hotel < ActiveRecord::Base
  has_paper_trail :ignore => [:user_id],  :meta => {:hotel_id => :id, :agent_user_id => :user_id}
  belongs_to :user
end
```

what is happening here?

PR #458, opened 2015-01-06
store attribute metadata as it exists at creation, instead of nil

rhec:
When the initial paper trail version is created any attribute metadata will be nil instead of the values that exist, because #{attribute}_was will always return nil for new models. This looks like an explicit decision, given there is a test, but I don't understand the logic behind it. 
For background, in my case I'm building a multi-tenant application and I'm adding the tenant id attribute to my paper trail version metadata. It's important for us to be able to query the version history restricted by tenant, and with the current behavior it's impossible to match create events with a tenant without reifying the actual objects.
Any other use case for storing attribute metadata I can think of also would benefit from storing the metadata that exists at create, rather than storing nil, but maybe I'm missing something?

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -403,8 +403,9 @@ module PaperTrail
             if v.respond_to?(:call)
               v.call(self)
             elsif v.is_a?(Symbol) && respond_to?(v)
-              # if it is an attribute that is changing, be sure to grab the current version
-              if has_attribute?(v) && send("#{v}_changed?".to_sym)
+              # if it is an attribute that is changing in an existing object, 
+              # be sure to grab the current version
+              if has_attribute?(v) && send("#{v}_changed?".to_sym) && data[:event] != 'create'
                 send("#{v}_was".to_sym)
               else
                 send(v)
--- a/test/unit/model_test.rb
+++ b/test/unit/model_test.rb
@@ -800,8 +800,8 @@ class HasPaperTrailModelTest < ActiveSupport::TestCase
         assert_equal @article.action_data_provider_method, @article.versions.last.action
       end
 
-      should 'store dynamic meta data based on an attribute of the item prior to creation' do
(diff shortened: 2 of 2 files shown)


batter:
So as discussed in some previous issues (#185, #385, #387), this is the expected behavior, as all of the aspects of a model (including metadata) are meant to show what the model looked like PRIOR to the event driving the generation of the version. That being said, I'm guessing that lots of users (I've done this in the past as well) are often looking to have the metadata populated even on `create` events, so maybe it just makes sense to merge this.

Although it begs the question: How important are  `create` events?  I usually just skip them since they can't be `reified` anyways..

rhec:
For me the value of create events is to have a complete history available in paper trail without having to reify anything or check existing objects. If you don't store create events the only way to get the date a model was created is to check created_at on an existing model or reify a version for deleted models. This method also only works if your model has a created_at attribute, which isn't a given.

batter:
Cheers

Billiam:
Isn't this a significant mismatch in behavior for new records vs update records?

When metadata may be nil for created records, this matches the `created` behavior for Version `objects` as well, where new versions represent the state of the previous, rather than current data.

It seems like this behavior breaks a few possible expectations:

1. That metadata matches version behavior. With this change, this is true for updates, where both the metadata and object data refer to the state prior to the change, but no longer the case for created events.
2. That metadata can be used for queries across versions. Per the example shown in the readme, let's say that you wanted to find all Article versions with a particular `author_id`, and had an Article which was created, and then had a change in author.   
A query for `PaperTrail::Version.where(:author_id => author_id)` will return two results back for this article (rather than one), one for the initial created version, and one for the change event, both with identical metadata.
3. That metadata matches the current version. This is an incorrect expectation based on open issues, and documentation, but it *is* the apparent behavior when creating a new record.

Would a new issue be better for discussion, or has this already been hashed out?

jaredbeck:
These all seem like valid concerns, @Billiam. Ben, can address these concerns?

PR #460, opened 2015-01-08
add recent history

nikolai-b:
This is very minor but I keep having to use the order and limit scopes.  I thought it might be useful for others.

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -116,6 +116,12 @@ module PaperTrail
       def object_changes_col_is_json?
         @object_changes_col_is_json ||= columns_hash['object_changes'].try(:type) == :json
       end
+
+      # Returns ActiveRecord::Relation [] of recent versions
+      def recent_history(num_versions = 10)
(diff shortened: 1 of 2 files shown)


batter:
So it's just a shortcut for `limit` essentially?

nikolai-b:
its very minor, I wasn't sure if others might use it but as I'd written it I thought I'd put in a pr incase.

batter:
Thank you for the PR, I don't think there's enough functionality being added here as it's essentially an alias for the `limit` method, which is provided out of the box by ActiveRecord.  Looking at the PR makes me curious though, we've never had a default sort order on the `PaperTrail::Version` class, even though we do have a default sort order for the association accessor. Do we think it makes sense to use the `timestamp_sort_order` as the default sort order?

My take was always just to let the user specify if they wanted to, but let AR do it's default, but I think it's a valid question.

nikolai-b:
So adding a default would make sense for me, I always want to see most recent versions first. For someone else though, if they want the order `:asc` then currently they can just `order(created_at: :asc)` but if you add a default then they have to `reorder(created_at: :asc)`.  Guess it depends on what most users want.

Issue #420, opened 2014-09-10
where_object works only for update event


panSarin:
Since it search only in "object" field, there is no way that it find 1st version (when event == create cause it always have object => nil). There is a reason of that , or it is TODO - i mean searching also in object_changes field?

PR #462, opened 2015-01-09
Versioning a model that has a polymorphic belongs_to association

kaapa:
New association versioning feature is broken for polymorphic belongs_to associations as reflection has no way to point to the dynamically set [associated model class](https://github.com/kaapa/paper_trail/compare/polymorphic_belongs_to_association?expand=1#diff-7fc294d950a7311cf6ad815a91f90ac2L376).

These commits add a failing test case and a "fix" by disabling the feature for polymorphic associations.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -372,7 +372,7 @@ module PaperTrail
       # saves associations if the join table for `VersionAssociation` exists
       def save_associations(version)
         return unless PaperTrail::VersionAssociation.table_exists?
-        self.class.reflect_on_all_associations(:belongs_to).each do |assoc|
+        self.class.reflect_on_all_associations(:belongs_to).reject(&:polymorphic?).each do |assoc|
           if assoc.klass.paper_trail_enabled_for_model?
             PaperTrail::VersionAssociation.create(
               :version_id => version.id,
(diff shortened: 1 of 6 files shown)


batter:
Hmm.... Any ideas of how we might be able to make this work instead of just ignoring those associations? @bli?

kaapa:
Unfortunately I don't think I'm able to take this much further at this time or in the immediate future as I'm a bit time constrained right now... :(

Perhaps this could be merged as an intermediate fix en route to full support for polymorphic associations?

PS. At a glance Travis test failure seems unrelated to this patch. Is that correct? Also perhaps a mention about polymorphic association behavior in the readme's Associations section should be added whether or not this patch gets merged.

batter:
Travis is failing because I have a few tests I created to test my feature implementation for #420 (`where_object_changes` method implementation) which seem to have inconsistent return values on certain versions of Ruby, so I need to fix those myself. Going to try to do that later today, but anyways, you are correct that the failures are unrelated to this.

batter:
Just to follow up @mhuggins was able to fix the polymorphic association versioning for even `belongs_to` associations with the work he did in #476. Thanks for the PR though.

mhuggins:
@batter I didn't consider writing a test that covers polymorphic `belongs_to` associations, only `has_many` associations.  It's probably worth adding a test that covers that though, and perhaps `has_one` associations as well!

PR #468, opened 2015-01-21
Describe use of JSON columns for JSON serializer

kjohnston:
I'm quite excited about being able to serialize the object and object_changes as JSON, but found that the README didn't highlight the fact that you could use JSON PostgreSQL columns to store the data, so I'm proposing we add it here.


batter:
Good call on this. Didn't realize it wasn't documented. Thanks!

batter:
Actually, you don't even need to use the JSON serializer in order to do this. The gem should be smart enough to just store the raw hash data into the JSON columns out of the box / without any adjustments.

PR #471, opened 2015-01-30
Fix typo in readme

qsymmachus:
':post_version_id_seq' should be ':post_versions_id_seq'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,7 +570,7 @@ If you are using Postgres, you should also define the sequence that your custom 
 ```ruby
 class PostVersion < PaperTrail::Version
   self.table_name = :post_versions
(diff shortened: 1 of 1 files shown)


batter:
Cheers!

batter:
Actually I don't think it matters what the `sequence_name` is, as long as it's unique, right?

qsymmachus:
You're right, a sequence can be named whatever you want. But if I'm not mistaken, when you create a postgres table with a primary key (the `serial` type), it also creates a sequence with the name `<table_name>_id_seq`. So by convention we would expect a sequence for the table `post_versions` to be called `post_versions_id_seq`.

PR #473, opened 2015-01-30
Fix typo in README

akenger:
There is a typo in the master branch readme.  3.0.6 will not get the features documented in the page.  This is very confusing.


batter:
Makes sense, cheers

PR #476, opened 2015-02-01
Fix saving of polymorphic associations

### LINKED ISSUE #474: Issue tracking history on polymorphic associations (4.0.0.beta2)
Hi there!  I'm trying to use the paper_trail gem with custom version tables/abstract `PaperTrail::Version` model as well as tracking version associations.  My relevant classes and modules look like this:

```ruby

PaperTrail::Version.module_eval do
  self.abstract_class = true
end

# app/models/concerns/taggable.rb
module Taggable
  extend ActiveSupport::Concern

  included do
    has_many :taggings, as: :taggable, inverse_of: :taggable, dependent: :destroy
    has_many :tags, through: :taggings
  end
end

# app/models/question.rb
class Question < ActiveRecord::Base
  include Taggable

  has_paper_trail class_name: 'QuestionVersion'
end

# app/models/tagging.rb
class Tagging < ActiveRecord::Base
  belongs_to :taggable, polymorphic: true, inverse_of: :taggings
  belongs_to :tag, inverse_of: :taggings, counter_cache: true

  has_paper_trail class_name: 'TaggingVersion'
end

# app/models/tag.rb
class Tag < ActiveRecord::Base
  has_many :taggings, inverse_of: :tag, dependent: :destroy
  has_many :taggables, through: :taggings

  has_paper_trail class_name: 'TagVersion'
end

# app/models/question_version.rb
class QuestionVersion < PaperTrail::Version
  self.table_name = :question_versions
end

# app/models/tagging_version.rb
class TaggingVersion < PaperTrail::Version
  self.table_name = :tagging_versions
end

# app/models/tag_version.rb
class TagVersion < PaperTrail::Version
  self.table_name = :tag_versions
end
```

For reference, my migrations are defined as:

```ruby
class CreateQuestionVersions < ActiveRecord::Migration
  def change
    create_table :question_versions do |t|
      t.string   :item_type,     null: false
      t.integer  :item_id,       null: false
      t.string   :event,         null: false
      t.integer  :transaction_id
      t.string   :whodunnit
      t.text     :object
      t.text     :object_changes
      t.datetime :created_at,    null: false
    end

    add_index :question_versions, [:item_type, :item_id]
    add_index :question_versions, [:transaction_id]
  end
end

class CreateTagVersions < ActiveRecord::Migration
  def change
    create_table :tag_versions do |t|
      t.string   :item_type,     null: false
      t.integer  :item_id,       null: false
      t.string   :event,         null: false
      t.integer  :transaction_id
      t.string   :whodunnit
      t.text     :object
      t.text     :object_changes
      t.datetime :created_at,    null: false
    end

    add_index :tag_versions, [:item_type, :item_id]
    add_index :tag_versions, [:transaction_id]
  end
end

class CreateTaggingVersions < ActiveRecord::Migration
  def change
    create_table :tagging_versions do |t|
      t.string   :item_type,     null: false
      t.integer  :item_id,       null: false
      t.string   :event,         null: false
      t.integer  :transaction_id
      t.string   :whodunnit
      t.text     :object
      t.datetime :created_at,    null: false
    end

    add_index :tagging_versions, [:item_type, :item_id]
    add_index :tagging_versions, [:transaction_id]
  end
end

class CreateVersionAssociations < ActiveRecord::Migration
  def change
    create_table :version_associations do |t|
      t.integer  :version_id
      t.string   :foreign_key_name, null: false
      t.integer  :foreign_key_id
    end

    add_index :version_associations, [:version_id]
    add_index :version_associations, [:foreign_key_name, :foreign_key_id], name: 'index_version_associations_on_foreign_key'
  end
end
```

The error pops up in the [`save_associations` method of `has_paper_trail.rb`](https://github.com/airblade/paper_trail/blob/v4.0.0.beta2/lib/paper_trail/has_paper_trail.rb#L376).  The call to `assoc.klass` is returning `Taggable`, which from my model definitions is supposed to be a polymorphic association.  Instead, it's resolving to my `Taggable` module, and attempting to call `paper_trail_enabled_for_model?` on it, which is not defined.

It seems that instead of trying to classify the polymorphic association via `assoc.klass` (which returns `Taggable`), it would make more sense to call `send(assoc.name).class` (which in this case would return `Question`).  Unfortunately, I'm not sure if this will fix everything since the following call would end up storing the `foreign_key_name` as `taggable_id` instead of `question_id` with no reference to the `foreign_type` of `taggable_type`:

```ruby
PaperTrail::VersionAssociation.create(
  :version_id => version.id,
  :foreign_key_name => assoc.foreign_key,
  :foreign_key_id => self.send(assoc.foreign_key)
)
```

Any thoughts or input?  I could very well just be doing something wrong, or perhaps there is no working path for polymorphic associations at this time.

mhuggins:
Fixes #474 to address issue with saving polymorphic associations.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -373,15 +373,20 @@ module PaperTrail
       def save_associations(version)
         return unless PaperTrail::VersionAssociation.table_exists?
         self.class.reflect_on_all_associations(:belongs_to).each do |assoc|
-          associated_record = send(assoc.name)
-
-          if associated_record && associated_record.class.paper_trail_enabled_for_model?
-            PaperTrail::VersionAssociation.create(
+          assoc_version_args = {
               :version_id => version.id,
-              :foreign_key_name => assoc.foreign_key,
-              :foreign_key_id => associated_record.id
-            )
+              :foreign_key_name => assoc.foreign_key
+          }
+
+          if assoc.polymorphic?
+            if (associated_record = send(assoc.name)).class.paper_trail_enabled_for_model?
+              assoc_version_args.merge!(:foreign_key_id => associated_record.id)
+            end
+          elsif assoc.klass.paper_trail_enabled_for_model?
+            assoc_version_args.merge!(:foreign_key_id => send(assoc.foreign_key))
           end
+
+          PaperTrail::VersionAssociation.create(assoc_version_args) if assoc_version_args.has_key?(:foreign_key_id)
         end
       end
 
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -373,11 +373,13 @@ module PaperTrail
       def save_associations(version)
         return unless PaperTrail::VersionAssociation.table_exists?
         self.class.reflect_on_all_associations(:belongs_to).each do |assoc|
-          if assoc.klass.paper_trail_enabled_for_model?
+          associated_record = send(assoc.name)
+
+          if associated_record && associated_record.class.paper_trail_enabled_for_model?
             PaperTrail::VersionAssociation.create(
               :version_id => version.id,
               :foreign_key_name => assoc.foreign_key,
-              :foreign_key_id => self.send(assoc.foreign_key)
+              :foreign_key_id => associated_record.id
             )
           end
         end
@@ -403,7 +405,7 @@ module PaperTrail
             if v.respond_to?(:call)
               v.call(self)
             elsif v.is_a?(Symbol) && respond_to?(v)
-              # if it is an attribute that is changing in an existing object, 
+              # if it is an attribute that is changing in an existing object,
               # be sure to grab the current version
               if has_attribute?(v) && send("#{v}_changed?".to_sym) && data[:event] != 'create'
                 send("#{v}_was".to_sym)
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -378,7 +378,7 @@ module PaperTrail
               :foreign_key_name => assoc.foreign_key
           }
 
-          if assoc.polymorphic?
+          if assoc.options[:polymorphic]
             if (associated_record = send(assoc.name)).class.paper_trail_enabled_for_model?
               assoc_version_args.merge!(:foreign_key_id => associated_record.id)
             end
--- a/test/dummy/app/models/whatchamajigger.rb
+++ b/test/dummy/app/models/whatchamajigger.rb
@@ -1,4 +1,4 @@
 class Whatchamajigger < ActiveRecord::Base
   has_paper_trail
-  belongs_to :owner, polymorphic: true
+  belongs_to :owner, :polymorphic => true
 end
--- a/test/dummy/app/models/whatchamajigger.rb
+++ b/test/dummy/app/models/whatchamajigger.rb
@@ -1,0 +1,4 @@
+class Whatchamajigger < ActiveRecord::Base
+  has_paper_trail
+  belongs_to :owner, polymorphic: true
+end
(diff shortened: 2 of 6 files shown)


batter:
This looks very promising, however, there are still a few tweaks I'd like to see made...

I'd still like to see [this `if` statement kept](https://github.com/airblade/paper_trail/pull/476/files#diff-7fc294d950a7311cf6ad815a91f90ac2L376), to prevent unnecessary association loading when possible and a corresponding class is not being tracked by PaperTrail in the first place.

My proposal is to split the logic for that method into 2 different possible routes, one for `polymorphic` associations, one for normal associations, like so:

```ruby
def save_associations(version)
  return unless PaperTrail::VersionAssociation.table_exists?
  self.class.reflect_on_all_associations(:belongs_to).each do |assoc|
    assoc_version_args = {
      :version_id       => version.id,
      :foreign_key_name => assoc.foreign_key
    }

    if assoc.options[:polymorphic] == true &&
      (associated_record = send(assoc.name)).class.paper_trail_enabled_for_model?
        assoc_version_args.merge!(:foreign_key_id => associated_record.id)
    elsif assoc.klass.paper_trail_enabled_for_model?
      assoc_version_args.merge!(:foreign_key_id => self.send(assoc.foreign_key))
    end
    PaperTrail::VersionAssociation.create(assoc_version_args)
  end
end
```

This optimizes performance by avoiding the call to load the association unless it is truly necessary (as is the case with Polymorphic associations). Thoughts?

daniel-rikowski:
I'm affected by #474, too, so I just chime in:

@batter Your proposal partly cancels out @mhuggins code:

If the association is polymorphic but the associated model - which is now found correctly - has no paper trail, the `elsif` branch kicks in and again calls the broken `klass` method. If the association is polymorphic, the `klass` method must never be called.

Also the code would create `VersionAssociation` instances without a `foreign_key_id` (even if all checks fail `create` is still called) 

Here is my proposal:

``` ruby
assoc_version_args = {
    :version_id => version.id,
    :foreign_key_name => assoc.foreign_key
}

if assoc.options[:polymorphic]
  if (associated_record = send(assoc.name)).class.paper_trail_enabled_for_model?
    assoc_version_args.merge!(:foreign_key_id => associated_record.id)
  end
elsif assoc.klass.paper_trail_enabled_for_model?
  assoc_version_args.merge!(:foreign_key_id => self.send(assoc.foreign_key))
end

PaperTrail::VersionAssociation.create(assoc_version_args) if assoc_version_args.key?(:foreign_key_id)
```
It's working, but could use some refactoring. Perhaps using `next` would make it less convoluted...

batter:
@daniel-rikowski - Thanks for chiming in, you're spot on with your code assessment(s).  I nominate your code suggestion for usage.

@mhuggins, do you want to modify your PR or do you want me to modify during the merge process?

mhuggins:
Thanks to you both for the feedback!  The concern makes sense, and the `klass` callout is a good catch.  I've gone ahead and updated the PR with this change.  Please let me know if you need me to squash the commits or anything else before this goes in! :)

mhuggins:
It looks like this is failing due to use of the new hash syntax used by the ffaker gem.  They literally JUST pushed a change out today [as seen here](https://rubygems.org/gems/ffaker) making use of the new hash syntax.  I think the fix would require changing the following:

```ruby
# add_development_dependency 'ffaker',  '>= 1.15'
# add_development_dependency 'ffaker',  '>= 1.15', '<= 1.31.0'
```

batter:
Good to know. I've actually been considering dropping Ruby18 support here since I question how much it's actually being utilized, but since compatibility is usually just a case of using the hash rocket syntax, it hasn't been difficult to maintain compatibility.. Anyways, I will push a commit up to master this afternoon to limit the 'ffaker' dependency to avoid test failures.

mhuggins:
Sounds good.  Just a heads up that the owner of ffaker said this is intentional since 1.8.7 is no longer supported.

batter:
You can rebase your branch off current master and/or merge master into it if you want to see it pass on Travis, either way, I'll try to merge it before EOD

PR #480, opened 2015-02-03
Avoid looking up item when reifying with dup: true

TMaYaD:
When `dup: true` is set on reify we don't need to look up the item. This helps in avoiding unnecessary lookup when showing lists of reified versions


batter:
Good call on this, thank you!

TMaYaD:
My pleasure.

PR #484, opened 2015-02-08
Add support for PostgreSQL jsonb columns

mattbeedle:
Postgres now has binary JSON columns


batter:
And it accepts the same raw input?  [Docs](http://www.postgresql.org/docs/9.4/static/datatype-json.html) say:

> They accept **almost** identical sets of values as input

Just want to make sure we don't need to do any special serialization for inserts here, sounds like we don't.

Issue #483, opened 2015-02-06
Custom classes & Rails' console reload!


lacco:
I just updated to PaperTrail master/ 4.0.0.beta2, added ```PaperTrail::Rails::Engine.eager_load!``` to ```config/initializers/papertrail.rb```, and now I get the following error when using Rails' console in development:

    > PostVersion.first
    => #<PostVersion id: 7 ...
    > reload!
    > PostVersion.first
    PG::UndefinedTable: ERROR:  relation "versions" does not exist

batter:
I'm actually not familiar with the `reload!

batter:
I'm actually not familiar with the `reload!` method. What do you use it for?

Also, does it return `true` or `false` from the `reload!` method? And did you try dropping in `PaperTrail::Rails::Engine.eager_load!` after you executed `reload!`?

lacco:
I am using ```reload!``` from time to time to play around with local changes and test them in the console immediately without reopening it (in my current case I was debugging performance issues in my application with papertrail's ```reify``` method).

Calling ```PaperTrail::Rails::Engine.eager_load!``` after ```reload!``` didn't help. What does help is to resetting ```self.abstract_class = true``` as done in my initializer. I have no clue why this is causing in issue - as far as I know ```reload!``` only reloads the application's code and not the code from gems...

> PostVersion.first
=> #<PostVersion id: 7 ...
> reload!
=> true
> PostVersion.first
PG::UndefinedTable: ERROR:  relation "versions" does not exist  
> PaperTrail::Rails::Engine.eager_load!
 => ["../lib/paper_trail/frameworks/active_record/models"] 
> PostVersion.first
PG::UndefinedTable: ERROR:  relation "versions" does not exist  
> PaperTrail::Version.module_eval do; self.abstract_class = true; end
=> #<PostVersion id: 7 ...

lacco:
I just saw this error outside of the Rails console during a page load: I opened a page that loads a ```PostVersion``` in development mode and then edited a file (e.g. adding some code to a controller). When I now reload the page, I get the ```PG::UndefinedTable: ERROR:  relation "versions" does not exist``` exception.

Could somebody try to reproduce my issues with ```reload!```? If I am the only one experiencing this, I'll dive into my code and try to find out the reason for this behavior.

lacco:
[...]

PR #492, opened 2015-03-03
Fix issues with Rails 4.2.x when not using STI for version models

### LINKED ISSUE #488: Issues with custom versions table names in 4.0.0.beta.2
Occasionally In normal development environment, and regularly using `spring` it seems that PaperTrail 4.0.0.beta.2 looses track of the proper table name for version models. I have many version models in my rather large application. Each with its own table and custom class. `PaperTraill::Version` has been modified with  `self.abstract_class = true` and is only used as such. Since upgrading to 4.0.0.beta.2 if I reload the application from the console with `reload!` things go haywire. The fact that `reload!` can consistently reproduce the issue points some to why spring just wont work at all. This seems to be some type of preload/autoload issue. I'm pretty sure this would not manifest in production environments where code is not reloaded, but I could be wrong. Here is an example of the problem in action:

```shell
$ be rails console                                                                  ✹
Loading development environment (Rails 3.2.21)
[1] pry(main)> StoryTagVersion.count
   (0.3ms)  SELECT COUNT(*) FROM `dk4_dev_versions`.`story_tag_versions` 
=> 1
[2] pry(main)> reload!
Reloading...
   (0.5ms)  SELECT DATABASE() as db
   (0.4ms)  SELECT DATABASE() as db
   (0.5ms)  SELECT DATABASE() as db
   (0.4ms)  SELECT DATABASE() as db
   (0.4ms)  SELECT DATABASE() as db
   (0.3ms)  SELECT DATABASE() as db
=> true
[3] pry(main)> StoryTagVersion.count
ActiveRecord::StatementInvalid: Mysql2::Error: Table 'dk4_dev_versions.versions' doesn't exist: SHOW FULL FIELDS FROM `versions`
from .bundle/bundle/ruby/1.9.1/gems/activerecord-3.2.21/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:245:in `query
```

TobiasBales:
Fixes #488, #483

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,18 +578,7 @@ Alternatively you could store certain metadata for one type of version, and othe
 
 If you only use custom version classes and don't use PaperTrail's built-in one, on Rails `>= 3.2` you must:
 
-- either declare the `PaperTrail::Version` class to be abstract like this (in an initializer):
-
-```ruby
-# config/initializers/paper_trail.rb
-
-# the following line is required for PaperTrail >= 4.0.0 with Rails
-PaperTrail::Rails::Engine.eager_load!
-
-PaperTrail::Version.module_eval do
-  self.abstract_class = true
-end
-```
+- either use `PaperTrail::AbstractVersion` 
 
 - or create a `versions` table in the database so Rails can instantiate the `PaperTrail::Version` superclass.
 
--- a/lib/paper_trail/frameworks/active_record/models/paper_trail/abstract_version.rb
+++ b/lib/paper_trail/frameworks/active_record/models/paper_trail/abstract_version.rb
@@ -1,0 +1,18 @@
+require 'paper_trail/version_concern'
+
+module PaperTrail
+  # When using PaperTrail with multiple tables without STI
+  # PaperTrail::Versions needs to be abstract or you have to create
+  # an unused versions table since Rails now reloads gemfiles as well
+  # as application code and the preferred method was to use an initializer
+  # which isn't reloaded resulting in PaperTrail::Version not being abstract.
+  #
+  # For details see #488 and #483
+  #
+  # So if you want an abstract base class and not use sti use AbstractVersion as
+  # your parent class.
+  class AbstractVersion < ::ActiveRecord::Base
+    include PaperTrail::VersionConcern
+    self.abstract_class = true
+  end
+end


bramswenson:
:+1: As it turns out I'm still suffering this issue with non-STI setup, but only when using reloading features ie spring.

bramswenson:
Oh, this is not a Rails version issue btw as my project is still Rails 3.2.x and suffers the same issue exactly.

bramswenson:
Trying to understand why this fixes the issue... @TobiasBales Do you think it was the timing of `PaperTrail::Version.module_eval` that caused rails to not see it as "abstract" until to late on code reload?

TobiasBales:
Not really a timing thing.
Aparently now the paper_clip source get's reloaded in the development environment when saving a file (not sure what exactly changed that) which means the class gets discarded from the cache and rebuild from the paper_trail sources.
The initializer is not interpreted again since they are one time only, no module_eval, no abstract class -> exceptions.

The other option is to just use

    require 'paper_trail/version_concern'
    class MyVersionVersion < ActiveRecord::Base
     end

instead of an inheritance chain (which is kind of pointless with the VersionConcern anyway)
and just define your own versions which derive from ActiveRecord::Base

batter:
So this is only an issue w/ PaperTrail 4.x?  I'm guessing the cause is the loading of the version class via a `Rails::Engine` when possible (see #347), as that is what necessitates adding the `PaperTrail::Rails::Engine.eager_load!` to a custom initializer with the new version.

I don't even know how useful it is anymore but it sounds like maybe there should be a config option to allow users to select whether they want to eager load or lazily load the model. Fixing one issue has broken another :frown:

@TobiasBales - Won't you still need the `PaperTrail::Rails::Engine.eager_load!` on your initializer, even with these changes?

batter:
Anyone have an answer for me?

ngelx:
It works for me. Rails 4.2, ruby 2.1.2

@batter - Initialiser is not needed at all. Custom version should be something like:

```
class SomeCustomVersion < PaperTrail::AbstractVersion
  # custom behaviour, e.g:
  self.table_name = :some_custom_versions
end
```

Thanks both for your help and dedication.

johnnymo87:
I'm on Rails 4.1.5, Ruby 2.1.0, and this did *not* work for me, although I don't know why. Instead, I had to define ...
```
if Rails.env.development?
  PaperTrail::Version.module_eval do
    self.abstract_class = true
  end
end
```
... over every custom versions class.

ngelx:
@johnnymo87 sorry, but i must ask. have you changed the gemfile according?
```
gem 'paper_trail', github: 'TobiasBales/paper_trail', branch: 'fix-rails-4.2.0'
```
I'm not sure if that fix is for rails 4.2 and above, but I think not.

johnnymo87:
@ngelx Yes. Just as a sanity-check, I also did `bundle open paper_trail`, and inspected that Tobias's new code was in fact there.

But I tried the other thing he mentioned in here ...
```
require 'paper_trail/version_concern'

module PaperTrail
  class MyCustomVersion < ::ActiveRecord::Base
    self.table_name = :my_custom_versions
    self.sequence_name = :my_custom_versions_id_seq
  end
end
```
... and this worked as well.

jaredbeck:
Thanks @TobiasBales, this PR would address the particular problem of the `abstract_class` flag. However, I think this is a symptom of a larger problem, the practice of using an initializer to reopen `PaperTrail::Version`.  As you say,

> .. Rails now reloads gemfiles as well as application code .. the preferred method was to use an initializer which isn't reloaded ..

If whatever we put in the initializer will be lost when we [`reload!`][1], then the real problem is the initializer.  We must address the real problem, not just this particular symptom (the `abstract_class` flag).

Instead of the initializer, applications that need to reopen `PaperTrail::Version` should be encouraged to have their own base class that includes `VersionConcern`, just as you have done here Tobias.  To clarify, you would put your `AbstractVersion` in your own codebase.

I just tried this out with [a new rails app][2] and `reload!` seems to work:

```ruby
# app/models/abstract_version.rb
class AbstractVersion < ::ActiveRecord::Base
  self.abstract_class = true
end

# app/models/banana_version.rb
class BananaVersion < AbstractVersion
  self.table_name = :banana_versions
end

# app/models/banana.rb
class Banana < ::ActiveRecord::Base
  has_paper_trail class_name: 'BananaVersion'
end

# in console
Loading development environment (Rails 4.2.2)
Banana.create(grams: 125) 
=> #<Banana id: 7, grams: 125>
reload!
Reloading...
=> true
Banana.create(grams: 125)
=> #<Banana id: 8, grams: 125>
```

Tobias and Ben, please take a look at [my complete experiment][2] and let me know if I've missed anything.

If that all sounds good, I recommend that we close this PR and update the documentation, replacing any mention of an initializer with the suggestions above.

[1]: http://api.rubyonrails.org/classes/Rails/ConsoleMethods.html#method-i-reload-21
[2]: https://github.com/jaredbeck/paper_trail_alternative_to_pr_492

batter:
@jaredbeck - Thanks for taking a detailed look at this.  I will pull down the repo you created and play with it ASAP when I get chance, and this is great to know because I believe a lot of the issues people are seeing when they are playing around with `4.0.0.rc1` stem from the fact that there seems to be an issue with the custom initializer changes they make to their versions class not working as expected or not "sticking" so to speak.

My question is, what about users that don't want to use a custom version class of their own, but simply want to modify or add to the way that the standard `PaperTrail::Version` class behaves? (See #546, #526, etc).  Does it make more sense to tell them to put their custom class changes in `app/models/paper_trail/version.rb`?  Am I understanding correctly that the `PaperTrail::Version` class is being unloaded from memory and then reloaded into memory as part of the `Gemfile` reloading that is occurring in rails by default now?  And does this behavior occur only when the server is being run in `development` or `test` mode?

batter:
@jaredbeck - Exciting news. I was just playing around with your dummy app, trying to figure out a way to get methods and adjustments to `PaperTrail::Version` to stick, even after a call to `reload!`.

Originally I attempted this:

```ruby

module PaperTrail
    def self.foo
      'test'
    end
  end
end

$ rails c
Loading development environment in sandbox (Rails 4.2.2)
> PaperTrail::Version.foo
=> 'test'
> reload!
Reloading...
=> true
> PaperTrail::Version.foo
NoMethodError: undefined method `foo' for #<Class:0x007fb8c6807290>...
```

But then I got it to work by trying this:
```ruby
module PaperTrail
  module VersionConcern
    module ClassMethods
      def foo(date)
        'test'
      end
    end
  end
end

$ rails c
Loading development environment in sandbox (Rails 4.2.2)
> PaperTrail::Version.foo
=> 'test'
> reload!
Reloading...
=> true
> PaperTrail::Version.foo
=> 'test'
```

So perhaps this is the key to allowing users to openly modify their classes. The downside to this is that it impacts every class that includes `PaperTrail::VersionConcern`, but I don't think that's such a huge problem because users can always override what they don't want / need on their subclasses.  Does this make sense or should we be looking for an alternative solution?

brennovich:
It is on of the `ActiveSupport::Concern` magics, handle dependencies https://github.com/airblade/paper_trail/blob/master/lib/paper_trail/version_concern.rb#L5 :)

jaredbeck:
> My question is, what about users that don't want to use a custom version class of their own, but simply want to modify or add to the way that the standard PaperTrail::Version class behaves?

Yes, I think we should continue to support that in PaperTrail 4.x.  I'd like to have a discussion about deprecating `PaperTrail::Version` and requiring users to have their own class (like carrierwave, devise, sidekiq, cancan), but we should postpone that discussion to PaperTrail 5.0.

> Does it make more sense to tell them to put their custom class changes in app/models/paper_trail/version.rb? 
>

That sounds good, `app/models` makes more sense for an AR model than `config/initializers` :)

> Am I understanding correctly that the PaperTrail::Version class is being unloaded from memory and then reloaded into memory as part of the Gemfile reloading that is occurring in rails by default now? And does this behavior occur only when the server is being run in development or test mode?

Yes, that's my understanding as well.  I suspect this behavior is controlled by rails' `cache_classes` flag.

> .. I still don't fully understand when reload! is called in the rails lifecycle.

I believe `reload!` is only available in the rails console, though I suspect the underlying mechanism is the same as the `cache_classes` flag.

> I got it to work by [reopening ClassMethods in the initializer]

I have two concerns:

1. Will this technique (reopening `ClassMethods` in the initializer) work for the `abstract_class` flag?
2. If we continue to recommend reopening *anything* in an initializer, it will be tempting for users to put code outside `ClassMethods` and continue to run into this issue.  This can be mitigated with documentation, but it still feels like we're leading people in the wrong direction.

jaredbeck:
> Will this technique (reopening ClassMethods in the initializer) work for the abstract_class flag?

Ben, I don't see a way to set `abstract_class` by reopening `VersionConcern`.  It would have to go in the `included do` block, but `ActiveSupport::Concern` does not allow multiple `included do` blocks.

> Instead of the initializer, applications that need to reopen PaperTrail::Version should be encouraged to [instead] have their own base class that includes VersionConcern, just as you have done here Tobias. To clarify, you would put your AbstractVersion in your own codebase.

Have you, Tobias, or anyone else had a chance to try this technique?  Please let us know.

jaredbeck:
Thanks @TobiasBales for the PR, and everyone who joined the discussion.  Tobias, your solution would work for the particular problem of the `abstract_class` flag, but the larger problem is the technique of reopening `PaperTrail::Version` in an initializer.  Going forward, we're planning to recommend that people reopen or redefine `PaperTrail::Version` in their `app/models`.  Here are the proposed recommendations:

https://github.com/airblade/paper_trail/pull/557

Please take a minute to check it out and we can continue this discussion there.  Thanks!

Issue #481, opened 2015-02-04
Manual versioning?


marnen:
I have a use case where I don't want versions to be created on every update (at least for the time being), but only when I call `touch_with_version` on the object. I would think that `has_paper_trail on: []` would make this happen, but it doesn't seem to work: I get automatic versions on every update. What am I doing wrong?

FWIW, I'm using paper_trail 3.0.6 with Rails 4.2 and Ruby 2.2.0.

marnen:
Ah, I see what's going on. `on: []` is treated as if the `:on` key weren't there. I'll submit a pull request to fix this.

marnen:
Hmm, this may be harder than I thought; see my branch at https://github.com/marnen/paper_trail/tree/handle-on-nothing-correctly. I was able to get `on: []` to turn off all automatic versioning, but now `touch_with_version` doesn't create a version. Help?

batter:
Why not just do `has_paper_trail on: :destroy`?

marnen:
How will that help?

batter:
It will only generate versions when a record is destroy, and if you want to workaround that / eliminate those records, you could put something like `after_destroy { versions.last.delete }` as a callback on the record or something. I know it's not perfect, but I think it will server your purposes.

With `4.0.0` I'm going to work on something that allows for easier manual versioning and / or versioning upon association changes only.

marnen:
That's not my use case. I don't want to generate versions when a record is destroyed; rather, as I said above, I want to record snapshots of a record's state at particular times of my choosing (not necessarily coincident with an update).  So really what I need is `on: []` and some sort of `create_version!` method -- either that or custom versioning events. 

(Actually, default settings and `touch_with_version` would work, but would give me far more versions than I need, at least at this stage of the project.)

marnen:
Pull request #500 should implement what I'm talking about.

Issue #463, opened 2015-01-13
creating versions when only associations change


nyon:
At first, I'd like to thank you for the great work you and all other contributers have spend on this gem.

I noticed, that no versions entry is generated when only the association data is changing. Is this the expected behaviour, or am I doing something wrong? I tried to write a little workaround which actually works now, but I would like to know, if there is some better way to achieve what I want to:

    before_save :check_change
    
    def check_change
      if Proc.new { |p| p.articles.map { |e| e.send(:changed_notably?) }.any? }
        self.updated_at = DateTime.now
      end
    end

Otherwise ```my_object.versions``` returns only the main model changes.
Is there some option I can pass to ```has_paper_trail``` to reproduce this functionality?

Best regards.

theworkerant:
:+1: I've got the same thing going on with one of my associations. Other nearly identical associations call `update_attribute` on the parent model for other reasons, but they work.

batter:
This is a dup of #461, and yes, that is the expected behavior.

I assume you are using the `4.0.0beta1` release?  For the time being, you can invoke `touch_with_version` to generate a new version on demand, but by next release (hopefully later today) I'm hoping to adjust it so you can just call `touch` and have a callback handle the version generating.

theworkerant:
Cool, for simplicity I'm looking to update versions manually in my app. If I wanted to do that could, how could I prevent auto-versioning? I've tried `has_paper_trail :on => [:update], :only => []` but that doesn't seem to do it. Will the work with `touch_with_version` make this possible so we can say `:on => [:touch]`?

batter:
Not yet. I'm debating what to do with the `touch` method.  There was an issue about this #329, and originally I had planned to just invoke the `after_touch` callback to have the normal `touch` method generate versions.  That being said, then that eliminates the option of invoking `touch` without generating a version.

Another alternative is to use something like `alias_method_chain` to override the default behavior for `touch` so it does generate a version, but then still have the method to generate without a version available named `touch_without_version`.  Still not sure how to handle that.  What do you think?

theworkerant:
I can't say I'm an expert on the ins and outs of `paper_trail` yet, but I feel overriding `touch` might be more confusing than not. I don't see/use `touch` very often, but I feel like it comes up in circumstances where you want to avoid things like the save callback cycle and have supreme control (could be wrong).

[...]

PR #500, opened 2015-03-18
Deal with `on: []`

marnen:
See #481 for more discussion. This pull request enables `on: []` to disable all automatic versioning, yet still forces `touch_with_version` to write a version.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -70,14 +70,15 @@ module PaperTrail
             :order      => self.paper_trail_version_class.timestamp_sort_order
         end
 
-        options_on = Array(options[:on]) unless options[:on].nil? # so that a single symbol can be passed in without wrapping it in an `Array`
-        after_create  :record_create, :if => :save_version? if options_on.nil? || options_on.include?(:create)
-        if options_on.nil? || options_on.include?(:update)
+        options[:on] ||= [:create, :update, :destroy]
+        options_on = Array(options[:on]) # so that a single symbol can be passed in without wrapping it in an `Array`
+        after_create  :record_create, :if => :save_version? if options_on.include?(:create)
+        if options_on.include?(:update)
           before_save   :reset_timestamp_attrs_for_update_if_needed!, :on => :update
           after_update  :record_update, :if => :save_version?
           after_update  :clear_version_instance!
         end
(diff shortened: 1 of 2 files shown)


batter:
I think this looks great! I also think the work done here will make it easier to knock out #463, which is something I've been struggling with, and one of the few issues I want to make sure is addressed before the `4.0.0` release. Cheers!

lib/paper_trail/has_paper_trail.rb
```diff
@@ -269,6 +270,7 @@ def touch_with_version(name = nil)
 
         attributes.each { |column| write_attribute(column, current_time) }
         save!
+        record_update(true) if self.class.paper_trail_options[:on] == []
```
stian-fredrikstad: In my code I need to do touch_with_version because something in my metadata on the version has changed, not the object itself, but this line stops me from actually doing a record_update since my model has option[:on] set to nil, which will default it to [:create, :update, :destroy]. So is tocuh_with_version only meant to be used on models without automatic versions on update/create/destroy?

batter: "something in my metadata on the version has changed", then shouldn't you just be updating the version record directly instead of generating a new version?  I'm not sure I'm following you.  `touch_with_version` can be called regardless of what your `on` option is.

This line is here to ensure that in cases where `PaperTrail` is essentially set to not do recordings on changes, users can still manually generate versions via `touch_with_version`.  The following line (`save!(:validate => false)`) should generate versions if you have `options[:on]` set to `nil`, or if `:update` is in the collection.  Hope this helps.

jaredbeck: Unrelated to your discussion, I think I found a bug.  It sounds like `touch_with_version` should always create a version, but I think I found (and fixed) a situation where it doesn't: https://github.com/airblade/paper_trail/pull/563

stian-fredrikstad: @batter Yes, I see that ```save!(:validate => false)``` should be enough, but because my record doesn't change, it seems like active record wonth fire after_update, and hence the version will no be created.
```
Foo.last.touch_with_version
Foo Load (0.4ms)  SELECT  "foos".* FROM "foos"  ORDER BY "foos"."id" DESC LIMIT 1
   (0.1ms)  begin transaction
   (0.1ms)  commit transaction
 => true 
```
But this seems like it will work with the PR from @jaredbeck . Thank you

PR #501, opened 2015-03-19
Fixing bug with `PaperTrail::VersionAssociation.table_exists?` check

TMaYaD:
Not a real pull request. Wanted to start a dialogue about the issue.
The lambda always returns true to the best of my knowledge because it creates a proc and a proc is not false. What's the Original Author's(or should I say, Original Fixer) intent behind the change. Is there another way to fix it?

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -9,7 +9,7 @@ module PaperTrail
 
       # Need to inspect inside of a Proc so that tests pass even when DB is not initialized
       # such as when we run on Travis (there won't be a db in `test/dummy/db/`)
-      if lambda { PaperTrail::VersionAssociation.table_exists? }
+      if (lambda { PaperTrail::VersionAssociation.table_exists? }).call
         has_many :version_associations, :dependent => :destroy
       end
 


batter:
You are correct, now I feel silly for even having put that there / not noticing before... I'll try to work on it tonight.

batter:
@TMaYaD - Can you please try pulling down ca09a8c and seeing if that fixes your issue?  You were spot on that the check for the DB table should never have gone inside of a lambda.  For some reason I had it in my head that it would be lazily evaluated, but I think I was getting it confused with how on certain callback declarations you can pass `:if => Proc.new { # code here }` and it gets lazily evaluated (especially for dates / times, etc).

TMaYaD:
In that case the evaluation is moved from the time to callback is defined to the time the callback is triggered. While if doesn't do that directly, we can do something similar but I couldn't figure out what's the 'later' context you wanted the condition to defer till.

TMaYaD:
Made some changes straight on github on my fork. They may need a lot of tweaking to get it properly setup. But IMHO, this is the best approach for this issue.
https://github.com/airblade/paper_trail/compare/master...TMaYaD:master

batter:
Hmm thanks for the suggestions @TMaYaD - I think thats a good start, I'll take another crack at it soon and try to use some of those suggestions.

PR #505, opened 2015-03-20
if belongs_to: returns a nil object it crashes

robmathews:
turns out that if you have a model with belongs_to, a polymorphic association, and nil, bad things happen.

this PR fixes that.


batter:
Thanks for noticing that and making a PR! Also love that you added a test to cover this case.

PR #507, opened 2015-03-23
Add flag to skip saving object_changes when column exists

tconst:
For certain models especially with large TEXT columns I don't want to store object_changes but still need object_changes stored on other models.  I set the key name to :save_changes but can rename this to whatever you think is best.


batter:
Thanks for the PR, merged

PR #512, opened 2015-04-06
Fixed: the redefined methods problem

proAlexandr:
I have tried to fix problem with redefined attribute methods in a model.
Now, I get problem if I redefine an attribute getter and setter. For example:

```
def price=(value)
  write_attribute(:price, value / 60)
end

def price
  read_attribute(:price) * 60 
end
```

Version model use new getter (`price`) for store a value, but when it restore the value, version don't use my new setter.

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -185,7 +185,7 @@ module PaperTrail
         # Set all the attributes in this version on the model
         attrs.each do |k, v|
           if model.has_attribute?(k)
-            model[k.to_sym] = v
+            model.send "#{k}=", v
           else
             logger.warn "Attribute #{k} does not exist on #{item_type} (Version id: #{id})."
           end


batter:
You shouldn't need the custom setter to be invoked because the getter will read the value out properly. We store the "unserialized" value, and thus that's what gets assigned to the model when reifying, which seems like it should be correct unless I'm mistaken.

If I'm wrong then please provide a test demonstrating what you're trying to address here.  As you can see, we already have a test in the suite mimicking the type of thing you're referring to, and it passes with the current master but fails with your "patch" here..

batter:
[Here is an expanded view of all the code surrounding those tests / when `write_attribute` was implemented in the first place](https://github.com/airblade/paper_trail/commit/fce1c908ef33d5675dc471c8eddaaecdfbc8c50f)

proAlexandr:
I get a wrong behaviour when I use [multilang-hstore](https://github.com/bithavoc/multilang-hstore)
It uses overwriting default accessors and serialization simultaneously. 
I am going to write a test soon.

batter:
@proAlexandr, here is my proposed fix: f5f310bfdc94be42405c23055e916d1a4813a963

It's on the [`vattr_support` branch](https://github.com/airblade/paper_trail/tree/vattr_support), so can you try bundling from that branch and seeing if it fixes your test case please?

batter:
Thanks for the PR, and for bringing this to my attention. I made a slight modification so the `reify` method now handles both normal ActiveRecord attributes and custom defined virtual attributes and setters properly.

PR #513, opened 2015-04-06
Use Postgres containment operator where available

mikeraimondi:
`where_object` and `where_object_changes` are currently broken in master when using PostgreSQL JSON and JSONB for object serialization. This pull request fixes the problem in an ugly but performant fashion.

Note that Arel is unlikely to support the Postgres containment operator anytime soon, so I don't think we can rely upon it here. And since JSON and JSONB are Postgres-only for the foreseeable future, it might be OK to use database-specific syntax.

Also, I think there may be a SQL-injection security hole here. Thoughts?

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -1,4 +1,5 @@
 require 'active_support/concern'
+require 'json'
 
 module PaperTrail
   module VersionConcern
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -86,12 +86,17 @@ module PaperTrail
       # identically-named method in the serializer being used.
       def where_object(args = {})
         raise ArgumentError, 'expected to receive a Hash' unless args.is_a?(Hash)
-        arel_field = arel_table[:object]
 
-        where_conditions = args.map do |field, value|
-          PaperTrail.serializer.where_object_condition(arel_field, field, value)
-        end.reduce do |condition1, condition2|
-          condition1.and(condition2)
+        if object_col_is_json?
+          where_conditions = "object @> '#{args.to_json}'::#{columns_hash['object'].type}"
+        else
+          arel_field = arel_table[:object]
+
+          where_conditions = args.map do |field, value|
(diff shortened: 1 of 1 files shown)


batter:
Thanks for the PR, looks like a smart solution for the time being.

Your last statement you mean the code you are proposing for the Postgres JSON && JSONB columns here has a possible SQL-injection security hole?  Not sure there's much we can do about that. Did you have any thoughts?

mikeraimondi:
I *think* we're protected because we know we're dealing with a Ruby hash by the time we build the SQL query. Any string keys will be escaped.

Likewise, I *think* `to_json` will escape potentially malicious SQL strings, but a quick search fails to turn up anything.

I haven't verified either of these assertions with tests or official documentation, hence the caveat.

lib/paper_trail/version_concern.rb
```diff
@@ -1,4 +1,5 @@
 require 'active_support/concern'
+require 'json'
```
batter: This is unnecessary because the `to_json` method is already provided by [this requiring of `active_support/json`](https://github.com/airblade/paper_trail/blob/master/lib/paper_trail/serializers/json.rb#L1)

batter:
Any chance we could add some tests for this functionality? I'm happy to assist but I have no experience dealing with real queries against these types of columns.  In the test suite, we can wrap any associated tests in an `if ENV['DB'] == 'postgres'` blocks to allow the tests to only run against Postgres (which will happen automatically on Travis CI).

batter:
I partially ask this because I've never seen the `@>` operator you are utilizing here. I don't even see it in the [Postgres docs](http://www.postgresql.org/docs/9.3/static/functions-json.html)..

mikeraimondi:
They're in the [9.4 version of those docs](http://www.postgresql.org/docs/9.4/static/functions-json.html). Hm. The containment operator is JSONB only, so I'll have to to update the conditional.

I'll add some tests as well.

mikeraimondi:
Closed in favor of #518.

gmhawash:
Any plans on merging this PR?

PR #531, opened 2015-05-05
Fix no ::YAML::ENGINE

y00rb:
in 2.2.1, doesn't have ::YAML::ENGINE.
::YAML is Psych class. so adding in condition.

### DIFF
--- a/lib/paper_trail/serializers/yaml.rb
+++ b/lib/paper_trail/serializers/yaml.rb
@@ -23,7 +23,7 @@ module PaperTrail
       # in the serialized object_changes
       def where_object_changes_condition(arel_field, field, value)
         # Need to check first (before) and secondary (after) fields
-        if defined?(::YAML::ENGINE) && ::YAML::ENGINE.yamler == 'psych'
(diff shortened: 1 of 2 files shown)


lib/paper_trail/serializers/yaml.rb
```diff
@@ -23,7 +23,7 @@ def where_object_condition(arel_field, field, value)
       # in the serialized object_changes
       def where_object_changes_condition(arel_field, field, value)
         # Need to check first (before) and secondary (after) fields
-        if defined?(::YAML::ENGINE) && ::YAML::ENGINE.yamler == 'psych'
+        if defined?(::YAML::ENGINE) && ::YAML::ENGINE.yamler == 'psych' || ::YAML == Psych
```
batter: Can you qualify Psych with `::Psych`? I think that would make the Travis errors stop failing.

y00rb:
@batter What's ruby version setting in Travis?

y00rb:
@batter  fixed

batter:
Good call, thanks for the PR!

PR #540, opened 2015-05-26
Update hash syntax

st0012:
I just updated hash syntax in readme and generator's templates for better style, but if you think it's unnecessary, just close it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ PaperTrail stores the pre-change version of the model, unlike some other auditin
 # Add has_paper_trail to Widget model.
 
 >> widget.versions                             # []
->> widget.update_attributes :name => 'Wotsit'
+>> widget.update_attributes name: 'Wotsit'
 >> widget.versions.last.reify.name             # 'Doobly'
 >> widget.versions.last.event                  # 'update'
 ```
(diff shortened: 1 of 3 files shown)


batter:
Thanks for the PR. The gem is still Ruby18 compatible, which is why the old syntax is used on examples and in the migration generator. I've been considering deprecating Ruby18 support for a while but haven't done it yet.

st0012:
I see

jaredbeck:
> I've been considering deprecating Ruby18 support for a while but haven't done it yet.

Ben, how about dropping 1.8 support in 5.0?  If we do drop 1.8 support, I'd suggest a major version bump.  Semver isn't really clear on major bump vs. minor bump, but there have been some good discussions about it recently.  Haml, for example, is going to bump their major version when they drop 1.8 support.

- https://github.com/haml/haml/issues/761
- https://github.com/mojombo/semver/issues/148

jaredbeck:
Our master branch now represents PaperTrail 5.0. As a major version bump, now is a good time to drop ruby 1.8 support.  Stan, can you please rebase this?  Thanks!

st0012:
I think I'll open another branch to do this again, cause there are many conflicts.

jaredbeck:
> I think I'll open another branch to do this again, cause there are many conflicts.

OK.  I'll close this one.  Thanks.

Issue #539, opened 2015-05-25
`PaperTrail.enabled` is not Thread safe.


vimutter:
At least `enabled=` and `enabled?` should be implemented with Thread local store instead of class variable.

```ruby
module PaperTrail
  # Switches PaperTrail on or off.
  def self.enabled=(value)
    Thread.current[:paper_trail_disabled] = !value
  end

  # Returns `true` if PaperTrail is on, `false` otherwise.
  # PaperTrail is enabled by default.
  def self.enabled?
    !Thread.current[:paper_trail_disabled]
  end
end
```

batter:
Closed by a1587724172a804e2307202aeb4fcefe78d6eff1

PR #541, opened 2015-05-26
Make Config.enabled backed by a thread variable #539

stormsilver:


batter:
Perhaps it makes more sense to just store it in the [`PaperTrail#paper_trail_store`](https://github.com/airblade/paper_trail/blob/v4.0.0.rc1/lib/paper_trail.rb#L122), which is backed by the thread-safe [`RequestStore` library](https://github.com/steveklabnik/request_store)?

stormsilver:
@batter I looked at that initially. I haven't seen `RequestStore` before but it looked like it really wanted a request to work properly. I'll take another swing through it and see about using the `paper_trail_store`.

stormsilver:
@batter Yep, as I looked into RequestStore it's obviously just a wrapper around sticking a hash with a name onto the thread. I replaced `Thread` with `PaperTrail.paper_trail_store` and it seemed to work okay - let me know if that's what you were thinking or not.

Issue #504, opened 2015-03-20
Yaml getting stripped


brandonhilkert:
We're stuck at version `1.6.5` because of Rails 2...I know it's probably no very well supported any more, but I was hoping someone could offer some insight. We're on Ruby 1.9.3 and have the following specified in the app:

```ruby
  YAML::ENGINE.yamler= 'syck'
```

`reify` works on some of our models, but not others. I tracked it down to the fact that, when the `object` gets in to the database, it looks like this:

```
--- \naccount_id: 1\ncached_tag_ids: \ncompany_id: \ncompany_name_d: \"\"\nconverted_at: \ncreated_at: 2014-12-12 14:28:47 Z\nwork_postal_code: \nwork_state:"
```

If from a Rails console, I run `person.attributes.to_yaml` (the same thing the get seems to be doing, I get:

```
--- \naccount_id: 1\ncached_tag_ids: \ncompany_id: \ncompany_name_d: \"\"\nconverted_at: \ncreated_at: 2014-12-12 14:28:47 Z\nwork_postal_code: \nwork_state:  \n"
```

Notice the ending ` \n` that's missing from the object in the database. I've experienced the same thing on 2 different models and it doesn't appear to be related to the number of characters, it seems to exclude the ending `\n` for each causing them to no be parsed when loaded back in.

Any idea why this would happen?

batter:
Looks to me like you have YAML that is parsable by newer versions of the YAML library (re: Psych) but not older ones (Syck).  Older versions of YAML have consistency issues as far as I know but I don't know what to tell you in terms of getting it to work with psych, but you may be able to workaround by setting it like to Psych like so:

```ruby
YAML::ENGINE.yamler= 'psych'
```

I was able to get both of those yaml dumpouts to load via calls to `YAML.load` when the YAMLER was set to `Psych`. Did you give that a try?

batter:
[This article may help educate you on the differences between the Psych and Syck libraries](http://devblog.arnebrasseur.net/2014-02-yaml-syck-vs-psych)

batter:
The inconsistencies and deficiencies I've witnessed in the `YAML` libraries over the years have been enough to tempt me to switch to `JSON` as the default serializer for this gem, the only problem being that it would immediately break everyone's applications who were using older versions of PaperTrail and didn't realize they needed to either switch their default serializer back to YAML or convert their serialized data to `JSON`.

brandonhilkert:
[...]

Issue #388, opened 2014-06-30
Reify on an ActiveRecord BigDecimal attribute incorrectly restores the wrong value


conanjen:
Currently using PaperTrail 3.0.1, and have a model with a decimal column. PaperTrail incorrectly serializes the decimal column as 18, and so all future reify calls restore the incorrect decimal value.

(As an example, the yaml output of an object with the bigdecimal attribute looks something like `18:0.725E1` for 7.25, but the version gets saved as 18.0.

Fairly certain this is related to how it's being encoded / decoded by the version object, however not really sure if this is due to papertrail or something else in my project. Gemfile.lock provided below as well.

```
GIT
  remote: git://github.com/gregbell/active_admin.git
  revision: 4f445b51c22b12af2cdde57fe2ce9835c32ef88e
  specs:
    activeadmin (1.0.0.pre)
      arbre (~> 1.0)
      bourbon
      coffee-rails
      formtastic (~> 2.3.0.rc3)
      inherited_resources (~> 1.3)
      jquery-rails
      jquery-ui-rails
      kaminari (~> 0.15)
      rails (>= 3.2, < 4.2)
      ransack (~> 1.0)
      sass-rails

GEM
  remote: https://rubygems.org/
  specs:
    actionmailer (4.1.0)
      actionpack (= 4.1.0)
      actionview (= 4.1.0)
      mail (~> 2.5.4)
    actionpack (4.1.0)
      actionview (= 4.1.0)
      activesupport (= 4.1.0)
      rack (~> 1.5.2)
      rack-test (~> 0.6.2)
    actionview (4.1.0)
      activesupport (= 4.1.0)
      builder (~> 3.1)
      erubis (~> 2.7.0)
    activemodel (4.1.0)
      activesupport (= 4.1.0)
      builder (~> 3.1)
    activerecord (4.1.0)
      activemodel (= 4.1.0)
      activesupport (= 4.1.0)
      arel (~> 5.0.0)
    activesupport (4.1.0)
      i18n (~> 0.6, >= 0.6.9)
      json (~> 1.7, >= 1.7.7)
      minitest (~> 5.1)
      thread_safe (~> 0.1)
      tzinfo (~> 1.1)
    acts-as-taggable-on (3.1.1)
      actionpack (>= 3, < 5)
      activerecord (>= 3, < 5)
      activesupport (>= 3, < 5)
    arbre (1.0.1)
      activesupport (>= 3.0.0)
    arel (5.0.1.20140414130214)
[...]

Issue #194, opened 2013-01-22
A JSON serializer should be included by default


batter:
Here is roughly what it would look like:

``` ruby
# lib/paper_trail/serializers/json.rb
require 'active_support/json'

module PaperTrail
  module Serializers
    class Json
      def self.load(string)
        ActiveSupport::JSON.decode string
      end

      def self.dump(hash)
        ActiveSupport::JSON.encode hash
      end
    end
  end
end
```

Or

``` ruby
# lib/paper_trail/serializers/json.rb
require 'json'

module PaperTrail
  module Serializers
    class Json
      def self.load(string)
        JSON.parse string
      end

      def self.dump(hash)
        hash.to_json
      end
    end
  end
end
```

@dwbutler - Can you elaborate a bit more on what you mean when you say "It's also more portable (for example, it will automagically pick up [MultiJSON](https://github.com/intridea/multi_json).)"?  I don't see ActiveSupport::JSON listed as one of the [supported JSON engines for MultiJSON](https://github.com/intridea/multi_json#supported-json-engines) (JSON is though).  Just curious as to what you think the pro's and con's are of using ActiveSupport::JSON vs. the default JSON gem?

Perhaps it makes sense to just bundle MultiJSON into the library, although, I'm a bit weary of doing that when this will not be the default parser for PaperTrail (at least for now).

PR #551, opened 2015-06-22
In MySQL, use longtext instead of text

### LINKED ISSUE #248: Big objects
currently migration is like the following:

```ruby
class CreateVersions < ActiveRecord::Migration
  def self.up
    create_table :versions do |t|
      t.string   :item_type, :null => false
      t.integer  :item_id,   :null => false
      t.string   :event,     :null => false
      t.string   :whodunnit
      t.text     :object
      t.datetime :created_at
    end
    add_index :versions, [:item_type, :item_id]
  end

  def self.down
    remove_index :versions, [:item_type, :item_id]
    drop_table :versions
  end
end
```

This approach is good enough for objects that serialized is smaller then 64kb (on mysql), as this is the maximum size "TEXT" can hold, according to http://www.electrictoolbox.com/maximum-length-mysql-text-field-types/

I've searched on how to force ActiveRecord migration to create "MEDIUMTEXT" or "LONGTEXT". It seems that all you have to do is define a limit that is bigger enough.

See http://stackoverflow.com/questions/4443477/rails-3-migration-with-longtext for more information.

I don't know what is the best default versions should have. It seems like a good idea to have a bigger size as default to prevent headaches from people who figure out late that they lost data because it wasn't saved (as the object was bigger then expected). Also worth noting that it increases only 2 bytes per entry according to: http://dev.mysql.com/doc/refman/5.0/en/storage-requirements.html#id733892

This might be the reason #245 happened.

jaredbeck:
See discussion in https://github.com/airblade/paper_trail/issues/248

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,8 @@ If you depend on the `RSpec` or `Cucumber` helpers, you will need to [manually l
     to `PaperTrail::Serializers::YAML`.
   - Both `PaperTrail.config` and `PaperTrail.configure` are now identical, and will both return the `PaperTrail::Config`
     instance and also yield it if a block is provided.
+  - [#248](https://github.com/airblade/paper_trail/issues/248) - The migrations
+    created by the generator now use `longtext` instead of `text` in MySQL.
 
 ## 3.0.8
 
--- a/lib/generators/paper_trail/templates/add_object_changes_to_versions.rb
+++ b/lib/generators/paper_trail/templates/add_object_changes_to_versions.rb
@@ -1,5 +1,10 @@
 class AddObjectChangesToVersions < ActiveRecord::Migration
+
+  # The largest text column available in all supported RDBMS.
+  # See `create_versions.rb` for details.
+  TEXT_BYTES = 1_073_741_823
+
   def change
-    add_column :versions, :object_changes, :text
+    add_column :versions, :object_changes, :text, limit: TEXT_BYTES
   end
 end
--- a/lib/generators/paper_trail/templates/create_versions.rb
+++ b/lib/generators/paper_trail/templates/create_versions.rb
@@ -1,11 +1,18 @@
 class CreateVersions < ActiveRecord::Migration
+
+  # The largest text column available in all supported RDBMS is
+  # 1024^3 - 1 bytes, roughly one gibibyte.  We specify a size
+  # so that MySQL will use `longtext` instead of `text`.  Otherwise,
+  # when serializing very large objects, `text` might not be big enough.
+  TEXT_BYTES = 1_073_741_823
+
   def change
     create_table :versions do |t|
       t.string   :item_type, :null => false
       t.integer  :item_id,   :null => false
       t.string   :event,     :null => false
       t.string   :whodunnit
-      t.text     :object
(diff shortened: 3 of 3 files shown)


batter:
Seems fine to me, and could potentially even be backported to v3.x, however, I wonder if we should just add this as a suggestion instead of a default?  In [your comment on 248](https://github.com/airblade/paper_trail/issues/248#issuecomment-114282740), you mention that:

> `text` uses two bytes to store its length, `mediumtext` uses three, and `longtext` uses four. Otherwise I'm not seeing any downside.

I don't think those are big enough size differences to prevent this from merging, although for an app with millions of rows, perhaps this would have big ramifications.  Users that know what they are doing can always take off the limit declaration. Then again they could also add it prior to this. So it's kind of a tossup, but I'll let you use your judgement as to what you feel makes the most sense.

I will say that I have been wondering for a while whether it makes sense to switch the default serializer to `PaperTrail::Serializers::JSON` (instead of `PaperTrail::Serializers::YAML`), because JSON syntax is more consistent in it's parsing (not as much of an issue anymore, but when it was unclear whether you'd get `syck` or `psych` depending on which Ruby version you were using for a while, it did cause some compatibility issues), and also dumps into shorter and more concise syntax than YAML (it's also faster as far as I know).  This would probably prevent some of these column limit size issues that people like @brodock were running up against. That being said, it would also be a breaking change for anyone who didn't read the documentation and realize they needed to manually swap in the old YAML serializer if they have legacy data (or update all their old data to JSON... perhaps we could provide a script / mechanism).

There is some appropriate / notable discussion about the pros / cons of JSON vs YAML in #194 & a bit of commentary on #504. There was also a user submitted script for updating legacy YAML data to JSON that we might be able to leverage in #388..  Since it's a non backwards compatible change, if we do decide to go this route, it would probably need to be part of a major version release.

jaredbeck:
> I wonder if we should just add this as a suggestion instead of a default?

I'm undecided.  On one hand, it's really easy to increase the size of the object column later if you hit the 64k limit that Gabriel hit in https://github.com/airblade/paper_trail/issues/248.  On the other hand, as Gabriel pointed out, 

> It seems like a good idea to have a bigger size as default to prevent headaches from people who figure out late that they lost data because it wasn't saved

That's hard to argue with, and we're only talking about 2-4 bytes per row.  Even with millions of rows, that's still only megabytes.

If you leave it up to me, I'll probably merge this, but I could be persuaded either way. :)

> I will say that I have been wondering for a while whether it makes sense to switch the default serializer to PaperTrail::Serializers::JSON (instead of PaperTrail::Serializers::YAML)

Interesting.  While I've worked with both formats for years, I can't say I've given them enough scrutiny to say which would be better for PaperTrail.  Of course, if you're using a modern postgres that supports json columns, it'd be a shame not to use that with PaperTrail.

batter:
> That's hard to argue with, and we're only talking about 2-4 bytes per row. Even with millions of rows, that's still only megabytes.

That's an excellent point, I say we merge it.

carsonreinke:
It should be noted that MySQL will silently trim off excess bytes with these BLOB fields, making the format unreadable after.  Making the default bigger is probably the best idea.

PR #556, opened 2015-06-28
Control the order of `set_paper_trail_whodunnit` callback

jaredbeck:
Fixes https://github.com/airblade/paper_trail/issues/301

Stop automatically adding the `set_paper_trail_whodunnit` before_filter.  This gives people control over the order of this callback.  People will have to do a little more thinking, but it makes the library less magical, easier to understand.  It's a tradeoff.

I tried to write a test for the warning (`warn_about_not_setting_whodunnit`) but couldn't get it passing reliably on Travis.  Help with that would be appreciated.  I also wasn't sure if we should use `Kernel#warn` or `ActiveSupport::Deprecation.warn`.

I'd like to postpone this change to 5.0, especially given that there's already a workaround for https://github.com/airblade/paper_trail/issues/301.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+## 5.0.0 (Unreleased)
+
+### Changed
+
+- If you are tracking who is responsible for changes with `whodunnit`, be aware
+  that PaperTrail no longer adds the `set_paper_trail_whodunnit` before_filter
+  for you. Please add this before_filter to your ApplicationController to
+  continue recording whodunnit. See the readme for an example.
+
+### Added
+
+None
+
+### Fixed
+
+None
+
 ## 4.1.0 (Unreleased)
 
 ### Changed
@@ -6,7 +23,6 @@
   [#588](https://github.com/airblade/paper_trail/issues/588) -
   `PaperTrail.timestamp_field` should be assigned to match the `updated_at`
   value when a version is generated for new versions.
-  
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -729,13 +729,28 @@ PaperTrail::Version.delete_all ["created_at < ?", 1.week.ago]
 
 ## Finding Out Who Was Responsible For A Change
 
-If your `ApplicationController` has a `current_user` method, PaperTrail will
-attempt to store the value returned by `current_user.id` in the version's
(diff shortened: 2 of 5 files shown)


seanlinsley:
I would argue that this should be an option, not the default. IMO, magic is only bad if it's undocumented.

jaredbeck:
> I would argue that this should be an option, not the default. IMO, magic is only bad if it's undocumented.

I'd be open to the idea of an option.  That would be better than the existing workaround for https://github.com/airblade/paper_trail/issues/301.

What other gems add controller callbacks?

- Devise offers a bunch of callbacks, like `authenticate_user!`, but it doesn't add them automatically.  Are there any other callbacks that it does add automatically?
- CanCan offers a class method (`load_and_authorize_resource`) to add its callbacks, instead of adding the callbacks automatically.

Are there any gems that automatically add callbacks?  If so, how do they handle callback ordering?

batter:
:+1: Looks good to me, and hopefully it will clear up the confusion surrounding this and cut down on people having issues with order of operations of assignment

jaredbeck:
> :+1: Looks good to me

Thanks for reviewing, Ben.  I've rebased, but I don't want to merge yet, because this is a breaking change, so I'd like to wait for 5.0.  Should we start 4.0-stable and 4.1-stable branches, so that master can represent 5.0?

jaredbeck:
Merged; `master` now represents 5.0.

> Should we start 4.0-stable and 4.1-stable branches ..?

Branches created.

batter:
Thanks for the re-organization @jaredbeck ... Do you think it makes more sense to have `4.1` be the current `master` (at least until a `4.1.x` is released), and have a `5.0-stable` branch for merging stuff upstream?

jaredbeck:
> Do you think it makes more sense to have 4.1 be the current master (at least until a 4.1.x is released), and have a 5.0-stable branch for merging stuff upstream?

It makes more sense to me if master always represents the bleeding edge, but I can't explain why, it's just what I'm used to.

It looks like most ruby projects use master to represent the "next" version. ([rails branches][1], [rspec branches][2]), but I don't know if there is a convention for when there are two unreleased versions.

[1]: https://github.com/rails/rails/branches
[2]: https://github.com/rspec/rspec/branches

PR #557, opened 2015-06-29
Docs: Do not recommend initializer

jaredbeck:
An alternative to https://github.com/airblade/paper_trail/pull/492

@batter @TobiasBales

### DIFF
--- a/README.md
+++ b/README.md
@@ -646,25 +646,19 @@ class PostVersion < PaperTrail::Version
 end
 ```
 
-If you only use custom version classes and don't use PaperTrail's built-in one,
-on Rails `>= 3.2` you must:
-
-- either declare the `PaperTrail::Version` class to be abstract like this (in an
-  initializer):
+If you only use custom version classes and don't have a `versions` table, you
+must let ActiveRecord know that the `PaperTrail::Version` class is an
+`abstract_class`.
 
 ```ruby
-# config/initializers/paper_trail.rb
-
-# the following line is required for PaperTrail >= 4.0.0 with Rails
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -646,26 +646,19 @@ class PostVersion < PaperTrail::Version
...
-
-# the following line is required for PaperTrail >= 4.0.0 with Rails
-PaperTrail::Rails::Engine.eager_load!
-
-PaperTrail::Version.module_eval do
-  self.abstract_class = true
+# app/models/paper_trail/version.rb
+module PaperTrail
+  class Version < ActiveRecord::Base
+    include VersionConcern
```
batter: I think we want to be explicit here and have this line read `include PaperTrail::VersionConcern` just to make sure there is no confusion as to where that module comes from.

README.md
```diff
@@ -646,26 +646,19 @@ class PostVersion < PaperTrail::Version
 end
 ```
 
-If you only use custom version classes and don't use PaperTrail's built-in one,
-on Rails `>= 3.2` you must:
-
-- either declare the `PaperTrail::Version` class to be abstract like this (in an
-  initializer):
+If you only use custom version classes and don't have a `versions` table, you
+must let ActiveRecord know that `PaperTrail::Version` is an `abstract_class`.
```
batter: I kind of liked the phrasing of "know that the `PaperTrail::Version` class is an `abstract_class`", just to be explicit about saying up front that `PaperTrail::Version` is a class to begin with (even though I understand it's implied)

jaredbeck:
Thanks Ben, I made the changes you recommended.

Issue #558, opened 2015-07-01
`has_paper_trail class_name:`  does not resolve modules.


tobypinder:
I expected the `class_name` parameter to work as methods like `#has_many` etc do in ActiveRecord  when it comes to modules -

      module AppName
        module Foo
          class Baz
            has_paper_trail class_name: 'BazVersion'
          end
        end

       [...]

        module Bar
          class Baz
            has_paper_trail class_name: 'BazVersion'
          end
        end
      end

Unfortunately it doesn't seem PaperTrail does this at present - instead I have to do something like

      module AppName
        module Foo
          class Baz
            has_paper_trail class_name: 'Foo::BazVersion'
          end
        end

       [...]

        module Bar
          class Baz
            has_paper_trail class_name: 'Bar::BazVersion'
          end
        end
      end

ActiveRecord would see the string with no namespace, and try the current namespace first, then it's parent and so on before giving up. It seems PaperTrail by comparison requires an "absolute" class name. Not a massive deal, but the former construction would allow me (and presumably others) to extract PaperTrail statements into concerns along with most of my other associations (as the two namespaces in my system have very similar models and associations)

jaredbeck:
> It seems PaperTrail .. requires an "absolute" class name. 
> .. the former construction would allow me .. to extract .. concerns

Thanks for writing that up, Toby. If you'd like to submit a PR that updates the documentation, we can probably get that into PT 4.0, and if you'd like to submit a PR that supports "relative" constant lookup, we could consider it for PT 4.1 or 5.0.  Are either of those something you'd be interested in working on?  Thanks.

tobypinder:
I'd certainly be comfortable changing the documentation to make this difference clear. I have no idea how AR implements this but I'll certainly try to take a look at it if I find some time at the weekend.

batter:
As far as I know, we literally pass that argument over to `ActiveRecord`'s `has_many` association declaration directly as it is passed to `has_paper_trail`, [here's the source](https://github.com/airblade/paper_trail/blob/v4.0.0.rc1/lib/paper_trail/has_paper_trail.rb#L70). Because we are literally just passing it over, I'm not sure what we would change to accomplish this. If you (or anyone else) has an idea, please make a pull request.

jaredbeck:
Now that we've improved the documentation with #559 I'm going to close this.  As Ben said, if anyone has an idea of how to improve the behavior of `class_name` please open a PR.  Thanks!

batter:
[...]

PR #559, opened 2015-07-02
Document limitations of `class_name`

tobypinder:
Make it clear that PaperTrail doesn't currently support relative `class_name`s for modules - an absolute module path is required. (See #558)


jaredbeck:
Thanks Toby!

PR #561, opened 2015-07-02
Add test demonstrating namespace

jaredbeck:
Demonstrates changes to docs in
https://github.com/airblade/paper_trail/pull/559

### DIFF
--- a/spec/models/kitchen/banana_spec.rb
+++ b/spec/models/kitchen/banana_spec.rb
@@ -1,0 +1,14 @@
+require 'rails_helper'
+
+module Kitchen
+  describe Banana, :type => :model do
+    it { is_expected.to be_versioned }
+
+    describe '#versions' do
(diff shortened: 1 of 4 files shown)


batter:
FWIW, we may be able to roll this back, since I believe this is already covered [by this bit of code](https://github.com/airblade/paper_trail/blob/master/spec/support/alt_db_init.rb#L32) and [this bit of code](https://github.com/airblade/paper_trail/blob/master/spec/support/alt_db_init.rb#L51)

jaredbeck:
Thanks, I didn't notice those.  I'm ok with reverting, but one possible reason to keep it would be that it uses sub-directories (`models/kitchen`, `versions/kitchen`).  Using actual sub-directories gives us test coverage for autoloading.  Do you think that provides enough value, or should we revert?

batter:
I think it's fine, doesn't hurt to have them, and it's a good point about the autoloading coverage. It doesn't hurt to have more coverage.

PR #566, opened 2015-07-13
Remove deprecated: paper_trail_on, paper_trail_off

jaredbeck:
I noticed the changelog for version 3.0.1 says these were to be removed in 4.0. :scissors: :scissors:


batter:
Hey @jaredbeck, yes is on the ToDo before the final 4.0.0 release. I was saving it as a "finishing touch" to address immediately prior to the release in case we decided to make one additional `rc` release prior to the official one.

jaredbeck:
> .. on the ToDo before the final 4.0.0 release. I was saving it as a "finishing touch" to address immediately prior to the release in case we decided to make one additional rc release prior to the official one.

I see what you're saying, and I could see leaving this out of a beta maybe, but In my experience the benefit of a release candidate is that people get a chance to test something that is as close to the final release as possible.

PR #567, opened 2015-07-13
Various doc updates

jaredbeck:
Mostly organizing the changelog, but also changes a deprecation message.

### DIFF
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -22,9 +22,11 @@ module PaperTrail
 
     def serialized_attributes=(value)
       if ::ActiveRecord::VERSION::MAJOR >= 5
-        warn("DEPRECATED: ActiveRecord 5.0 deprecated `serialized_attributes` "  +
+        ::ActiveSupport::Deprecation.warn(
+          "ActiveRecord 5.0 deprecated `serialized_attributes` "  +
           "without replacement, so this PaperTrail config setting does " +
-          "nothing with this version, and is always turned off")
+          "nothing with this version, and is always turned off"
+        )
       end
       @serialized_attributes = value
     end
@@ -33,12 +35,12 @@ module PaperTrail
       @track_associations ||= PaperTrail::VersionAssociation.table_exists?
     end
     alias_method :track_associations?, :track_associations
-    
+
     # Indicates whether PaperTrail is on or off.
     def enabled
       PaperTrail.paper_trail_store[:paper_trail_enabled].nil? || PaperTrail.paper_trail_store[:paper_trail_enabled]
     end
-    
+
     def enabled= enable
       PaperTrail.paper_trail_store[:paper_trail_enabled] = enable
     end
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -192,7 +192,7 @@ module PaperTrail
       end
 
       def originator
-        warn "DEPRECATED: use `paper_trail_originator` instead of `originator`. Support for `originator` will be removed in PaperTrail 4.0"
+        warn "DEPRECATED: use paper_trail_originator instead of originator."
         self.paper_trail_originator
       end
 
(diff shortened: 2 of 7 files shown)


lib/paper_trail/config.rb
```diff
@@ -22,9 +22,11 @@ def initialize
 
     def serialized_attributes=(value)
       if ::ActiveRecord::VERSION::MAJOR >= 5
-        warn("DEPRECATED: ActiveRecord 5.0 deprecated `serialized_attributes` "  +
+        ::ActiveSupport::Deprecation.warn(
```
batter: Nifty trick, I didn't realize this class existed. I like how it prints more details about the source of the warning too.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -192,7 +192,7 @@ def paper_trail_originator
       end
 
       def originator
-        warn "DEPRECATED: use `paper_trail_originator` instead of `originator`. Support for `originator` will be removed in PaperTrail 4.0"
+        ActiveSupport::Deprecation.warn "Use paper_trail_originator instead of originator."
```
batter: Let's fully qualify these constants by prepending `::` just in case we later add an `ActiveSupport` module within the gem.  (`::ActiveSupport::Deprecation`).

jaredbeck: > Let's fully qualify these constants by prepending :: just in case we later add an ActiveSupport module within the gem. (::ActiveSupport::Deprecation).

Roger wilco.  I was wondering about that, after seeing fully-qualified references elsewhere.

batter:
+1

jaredbeck:
Thanks, I'm going to look through the rest of the changelog entry for 4.0 for other items that could better be classified as "Changed" or "Fixed".

md5:
I notice in the comments for 5a03a1b7bb6f68bd0f510c12d298024a0c3fb05a that using an initializer is "no longer recommend[ed]". What approach should be used instead?

I still see an initializer being recommended in some cases, for example here: https://github.com/airblade/paper_trail/blob/2cec324c1e86480fb12854436d322ef800cc66d5/lib/paper_trail/config.rb#L43-L48

jaredbeck:
Thanks Mike, that's a good point. IIRC, we don't recommend re-opening the `::PaperTrail::Version` class in an initializer, which used to be a common practice. However, we actively recommend using an initializer to set `track_associations = false`. So, we definitely need to clarify these docs. Can you open a new issue please?

md5:
@jaredbeck I'm just circling back on this and I don't think there's anything to be done after https://github.com/airblade/paper_trail/commit/5a03a1b7bb6f68bd0f510c12d298024a0c3fb05a since the docs no longer recommend monkey patching in an initializer.

Or did you want an issue to update the docs to active recommend setting `track_associations = false`? It seems like updating the scaffolding to generate an initializer with `track_associations = false` might be a good idea.

jaredbeck:
> Or did you want an issue to update the docs to actively recommend setting track_associations = false? It seems like updating the scaffolding to generate an initializer with track_associations = false might be a good idea.

That's a great idea. Can you please take the lead on that? Thanks.

md5:
@jaredbeck I'm strapped for time right now, but I've opened #830

md5:
Note, I've also run into the fact that the `whodunnit` setting for Rake tasks that got moved to the wiki doesn't work if you use the Rails 5 convention of calling `rake` tasks via the `rails` command. I've made this update: https://github.com/airblade/paper_trail/wiki/Setting-whodunnit-in-the-rails-console/_compare/840c0238e5d2bada60bd366d1a0b5dde1bfc2e96...dd7d9ffe11b2cdb51c574a92f104674a33b99ff4

jaredbeck:
> .. the Rails 5 convention of calling rake tasks via the rails command.

Huh. First I've heard of that. What is their reason?

> I've made this update [to the wiki]

Thanks

md5:
See https://github.com/rails/rails/issues/18878 and https://github.com/rails/rails/pull/22288

md5:
I think it's more that Rails-bundled tasks should be called through the `rails` command.

PR #570, opened 2015-07-16
Respect `:skip` re: timestamp-only updates

### LINKED ISSUE #569: Ignoring and skipping attributes not working again?
Hi, I saw that this was an issue in the past.  In testing ignore and skips, it doesn't pass. I have a couple of simple skip and ignore fields .

```
class Case < Base
  has_paper_trail skip: [
                        :last_case_comment_at,
                        ],
                ignore: [
                        :created_at,
                        ]
```
And a simple test to see if they are indeed skipped and ignored.
```
    it 'should not track changes to skipped attributes' do
      count = @case.versions.length

      @case.update_attributes!(last_case_comment_at: Time.now)
      expect(@case.versions.length).to eql count
    end

    it 'should not track changes to ignored attributes' do
      count = @case.versions.length

      @case.update_attributes!(created_at: Time.now)
      expect(@case.versions.length).to eql count
    end
```
But for some reason the skips and ignores are being ignored.
```
 1) Case versioning should not track changes to skipped attributes
     Failure/Error: expect(@case.versions.length).to eql count
       
       expected: 1
            got: 2
       
       (compared using eql?)

 2) Case versioning should not track changes to ignored attributes
     Failure/Error: expect(@case.versions.length).to eql count
       
       expected: 1
            got: 2
       
       (compared using eql?)
```
I changed the ```count``` in my example to ```@case.versions.last.changeset``` to try to debug.
```
   it 'should not track changes to skipped attributes' do
      count = @case.versions.length

      @case.update_attributes!(last_case_comment_at: Time.now)
      expect(@case.versions.length).to eql @case.versions.last.changeset
   end
```
and saw that it was indeed creating another version.
```
  1) Case versioning should not track changes to skipped attributes
     Failure/Error: expect(@case.versions.length).to eql @case.versions.last.changeset
       
       expected: {"gender"=>["Female", "Female"], "birthdate"=>["1946-03-09", "1946-03-09"], "first_name"=>["Nolan", "Nolan"], "middle_name"=>["Jensen", "Jensen"], "last_name"=>["Walker", "Walker"]}
            got: 2
       
       (compared using eql?)
```
What gives?

jaredbeck:
When attributes are ignored, either by `:ignore` or by `:skip`,
then during updates, timestamp attributes like `updated_at` should
not be considered notable changes.

[Fixes #569]

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -469,16 +469,25 @@ module PaperTrail
         attrs
       end
 
-      # This method is invoked in order to determine whether it is appropriate to generate a new version instance.
-      # Because we are now using `after_(create/update/etc)` callbacks, we need to go out of our way to
-      # ensure that during updates timestamp attributes are not acknowledged as a notable changes
-      # to raise false positives when attributes are ignored.
+      # This method determines whether it is appropriate to generate a new
+      # version instance. A timestamp-only update (e.g. only `updated_at`
+      # changed) is considered notable unless an ignored attribute was also
+      # changed.
       def changed_notably?
-        if self.paper_trail_options[:ignore].any? && (changed & self.paper_trail_options[:ignore]).any?
-          (notably_changed - timestamp_attributes_for_update_in_model.map(&:to_s)).any?
+        if ignored_attr_has_changed?
+          timestamps = timestamp_attributes_for_update_in_model.map(&:to_s)
+          (notably_changed - timestamps).any?
         else
           notably_changed.any?
         end
+      end
+
+      # An attributed is "ignored" if it is listed in the `:ignore` option
+      # and/or the `:skip` option.  Returns true if an ignored attribute
+      # has changed.
+      def ignored_attr_has_changed?
+        ignored = self.paper_trail_options[:ignore] + self.paper_trail_options[:skip]
+        ignored.any? && (changed & ignored).any?
       end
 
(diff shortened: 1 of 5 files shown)


batter:
I'm looking at the code, it doesn't look like any of the logic has changed... I've re-read it a few times and I'm scratching my head. Is [this the difference here](https://github.com/airblade/paper_trail/pull/570/files#diff-7fc294d950a7311cf6ad815a91f90ac2R489)? The `+ self.paper_trail_options[:skip]`?

batter:
Thanks for looking into this btw!

jaredbeck:
> I'm looking at the code, it doesn't look like any of the logic has changed... I've re-read it a few times and I'm scratching my head. Is this the difference here? The + self.paper_trail_options[:skip]?

Yes, that's the key difference.  For the purposes of whether a change is "notable", we should respect both `:ignore` *and* `:skip`, but without this PR, we do not respect `:skip`.

batter:
+1, good catch on this one! Based on README description of how `skip` should function this seems right.

jaredbeck:
Cool, thanks for reviewing.  Do you want to cut a new release candidate with this and https://github.com/airblade/paper_trail/pull/566 ?

batter:
I think we probably should cut a new one since there have been a lot of changes to `master` since `rc1`, but I'm inclined to wait until the final release to merge in #566, because the warnings say they will be removed for 4.0.0 final.

PR #571, opened 2015-07-17
Relax version constraint on request_store

jaredbeck:
.. to allow `request_store` 1.2.

AFAICT they don't have a changelog, so we have to review commits:

https://github.com/steveklabnik/request_store/commits/master

The only new feature appears to be:

https://github.com/steveklabnik/request_store/pull/37


batter:
Looks good to me, doesn't appear like a drastic change, nor one that will really impact how we are utilizing RequestStore. I say merge it assuming the tests suite passes.

PR #579, opened 2015-07-31
Possibility to preserve skipped attributes on reify

McRip:
Hello, i added the possibility to preserve skipped attributes on reify instead of setting them all to nil.
I added an option to the reify method to enable this behaviour. A Test is attached to this PR.

Thanks for merging.

### DIFF
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -13,77 +13,108 @@
 
 ActiveRecord::Schema.define(version: 20110208155312) do
 
-  create_table "animals", force: true do |t|
+  create_table "animals", force: :cascade do |t|
     t.string "name"
     t.string "species"
   end
 
-  create_table "articles", force: true do |t|
+  create_table "articles", force: :cascade do |t|
     t.string "title"
     t.string "content"
     t.string "abstract"
     t.string "file_upload"
   end
 
-  create_table "authorships", force: true do |t|
+  create_table "authorships", force: :cascade do |t|
     t.integer "book_id"
     t.integer "person_id"
   end
 
-  create_table "books", force: true do |t|
+  create_table "banana_versions", force: :cascade do |t|
+    t.string   "item_type",  null: false
+    t.integer  "item_id",    null: false
+    t.string   "event",      null: false
+    t.string   "whodunnit"
+    t.text     "object"
+    t.datetime "created_at"
+  end
+
+  add_index "banana_versions", ["item_type", "item_id"], name: "index_banana_versions_on_item_type_and_item_id"
+
+  create_table "bananas", force: :cascade do |t|
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
+
+  create_table "books", force: :cascade do |t|
     t.string "title"
   end
 
-  create_table "customers", force: true do |t|
+  create_table "boolits", force: :cascade do |t|
+    t.string  "name"
+    t.boolean "scoped", default: true
+  end
+
+  create_table "customers", force: :cascade do |t|
     t.string "name"
   end
 
-  create_table "documents", force: true do |t|
+  create_table "documents", force: :cascade do |t|
     t.string "name"
   end
 
-  create_table "editors", force: true do |t|
+  create_table "editors", force: :cascade do |t|
     t.string "name"
   end
 
-  create_table "editorships", force: true do |t|
+  create_table "editorships", force: :cascade do |t|
     t.integer "book_id"
     t.integer "editor_id"
   end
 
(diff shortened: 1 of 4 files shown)


McRip:
I actually have no clue why this fails:

1) JsonVersion Methods Class #where_object_changes valid arguments should be able to locate versions according to their `object_changes` contents
     Failure/Error: expect(fruit.versions.where_object_changes(:color => color)).to eq(fruit.versions[1..2])
       
       expected: [#<JsonVersion id: 5, item_type: "Fruit", item_id: 2, event: "update", whodunnit: nil, object: {"id"=>2, "name"=>"pomegranate", "color"=>nil}, object_changes: {"name"=>["pomegranate", "coconut"], "color"=>[nil, "navy"]}, created_at: "2015-07-31 09:23:59">, #<JsonVersion id: 6, item_type: "Fruit", item_id: 2, event: "update", whodunnit: nil, object: {"id"=>2, "name"=>"coconut", "color"=>"navy"}, object_changes: {"name"=>["coconut", "orange"]}, created_at: "2015-07-31 09:23:59">]
            got: #<ActiveRecord::AssociationRelation [#<JsonVersion id: 5, item_type: "Fruit", item_id: 2, event: "update", whodunnit: nil, object: {"id"=>2, "name"=>"pomegranate", "color"=>nil}, object_changes: {"name"=>["pomegranate", "coconut"], "color"=>[nil, "navy"]}, created_at: "2015-07-31 09:23:59">]>
       
       (compared using ==)
       
       Diff:
       @@ -1,4 +1,4 @@
       -[#<JsonVersion:0x575ddcc9
       +[#<JsonVersion:0x288bd61a
          id: 5,
          item_type: "Fruit",
          item_id: 2,
       @@ -6,14 +6,5 @@
          whodunnit: nil,
          object: {"id"=>2, "name"=>"pomegranate", "color"=>nil},
          object_changes: {"name"=>["pomegranate", "coconut"], "color"=>[nil, "navy"]},
       -  created_at: Fri, 31 Jul 2015 09:23:59 UTC +00:00>,
       - #<JsonVersion:0x379f2abf
       -  id: 6,
       -  item_type: "Fruit",
       -  item_id: 2,
       -  event: "update",
       -  whodunnit: nil,
       -  object: {"id"=>2, "name"=>"coconut", "color"=>"navy"},
       -  object_changes: {"name"=>["coconut", "orange"]},
          created_at: Fri, 31 Jul 2015 09:23:59 UTC +00:00>]
     # ./spec/models/json_version_spec.rb:67:in `(root)'

jaredbeck:
Don't worry about the JsonVersion test, it's flakey.  I ran it again and it passed.

jaredbeck:
Can you squash this into one commit, please?

test/dummy/db/schema.rb
```diff
@@ -96,31 +127,38 @@
...
     t.integer "length"
   end
 
-  create_table "things", force: true do |t|
+  create_table "things", force: :cascade do |t|
     t.string "name"
   end
 
-  create_table "translations", force: true do |t|
+  create_table "translations", force: :cascade do |t|
```
jaredbeck: Why did the `force` option change from `true` to `:cascade` on every table? Is that the new recommended value, or is it something you did intentionally? From the docs:

```
:force
Set to true to drop the table before creating it. Set to :cascade to drop dependent objects as well. Defaults to false.
```

It seems like with `force: true` the order of table creation wouldn't matter, but with `:cascade` it would, so `true` seems preferable to me.

jaredbeck:
This seems like a useful feature, and it doesn't increase the complexity of `reify` too much.  

Also, I like this as an option to `reify`, as opposed to e.g. model-specific config.

As far as the API is concerned, you have:

```diff
+    # :preserve             `false` default behaviour
+    #                       `true` preserve attributes of reified object instead of setting unknown attributes to nil
 ```

but I think I'd prefer something like:

```diff
+    # :unversioned_attribute    `:nil` - (default) attributes undefined in version record are set to nil in reified record
+    #                           `:preserve` - attributes undefined in version record are not modified
 ```

My thinking is, this allows for a third option in the future.

test/dummy/db/schema.rb
McRip: It was changed during db:migrate. This has changed since Rails 4.2. The release notes say:

```SchemaDumper uses force: :cascade on create_table. This makes it possible to reload a schema when foreign keys are in place.```

McRip:
Hello Jared,
thanks for your feedback. I changed the syntax and merged all the commits into one.

jaredbeck:
Merged. Thanks Moritz!

PR #580, opened 2015-08-03
Organize 4.0.0 change list

jaredbeck:
Keeping all changes in a single list means users only have to look
in one place when they upgrade.

In this case, the "Removed" item was really obvious, but in general
a single list is a good practice, I think.

A single list is also recommended by http://keepachangelog.com/
which I'm a big fan of.

[ci skip]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ candidates.
   https://github.com/airblade/paper_trail/pull/492.
 - If you depend on the `RSpec` or `Cucumber` helpers, you must
   [require them in your test helper](https://github.com/airblade/paper_trail#testing).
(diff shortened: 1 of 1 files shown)


jaredbeck:
Ben, just wanted to run this by you in case you had strong feelings about "Removed" vs. "Changed"

batter:
I don't actually have a strong preference one way or the other. I guess I just wanted to put the "Removed" piece in bold to help make it stand out in case any users ran into this.  Last time something was removed from PaperTrail there was a lot of questions filed in the issues tracker about what happened with it if I remember correctly. Then again, I didn't really start following good semantic versioning practices until prep for 4.0.0

PR #587, opened 2015-08-08
Prevent global variable pollution

grosser:
atm if controller A sets whodunnit and then some other controller runs that does not include papertrail the whodunnit state lingers around, this also makes testing ugly since the whodunnit will randomly change and be lest in a bad state -> unrelated tests can fail.

These helpers should also be useful in general for background jobs or any task that does not want to change up global state.

@jaredbeck

### DIFF
--- a/lib/paper_trail/frameworks/rails/controller.rb
+++ b/lib/paper_trail/frameworks/rails/controller.rb
@@ -4,7 +4,7 @@ module PaperTrail
 
       def self.included(base)
         base.before_filter :set_paper_trail_enabled_for_controller
-        base.before_filter :set_paper_trail_whodunnit, :set_paper_trail_controller_info
+        base.around_filter :set_paper_trail_whodunnit, :set_paper_trail_controller_info
       end
 
       protected
@@ -61,14 +61,22 @@ module PaperTrail
       end
 
       # Tells PaperTrail who is responsible for any changes that occur.
-      def set_paper_trail_whodunnit
-        ::PaperTrail.whodunnit = user_for_paper_trail if ::PaperTrail.enabled_for_controller?
+      def set_paper_trail_whodunnit(&block)
+        if ::PaperTrail.enabled_for_controller?
+          ::PaperTrail.with_whodunnit(user_for_paper_trail, &block)
+        else
+          yield
+        end
       end
 
       # Tells PaperTrail any information from the controller you want to store
       # alongside any changes that occur.
-      def set_paper_trail_controller_info
-        ::PaperTrail.controller_info = info_for_paper_trail if ::PaperTrail.enabled_for_controller?
+      def set_paper_trail_controller_info(&block)
+        if ::PaperTrail.enabled_for_controller?
+          ::PaperTrail.with_controller_info(info_for_paper_trail, &block)
+        else
+          yield
+        end
       end
 
     end
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -85,6 +85,14 @@ module PaperTrail
     paper_trail_store[:whodunnit]
   end
 
+  # whodunnit for a given block of code
+  def self.with_whodunnit(value)
+    old, self.whodunnit = whodunnit, value
+    yield
+  ensure
+    self.whodunnit = old
+  end
+
   # Sets any information from the controller that you want PaperTrail to
   # store.  By default this is set automatically by a before filter.
   def self.controller_info=(value)
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -107,6 +107,14 @@ module PaperTrail
     paper_trail_store[:controller_info]
   end
 
+  # controller_info for a given block of code
+  def self.with_controller_info(info)
+    old, self.controller_info = controller_info, value
+    yield
+  ensure
+    self.controller_info = old
+  end
+
   # Getter and Setter for PaperTrail Serializer
   def self.serializer=(value)
     PaperTrail.config.serializer = value
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -24,6 +24,14 @@ module PaperTrail
     !!PaperTrail.config.enabled
   end
 
+  # Disable or enable for a given block of code
+  def self.with_enabled(value)
+    old, self.enabled = enabled?, value
+    yield
+  ensure
+    self.enabled = old
+  end
+
   # ActiveRecord 5 drops support for serialized attributes; for previous
   # versions of ActiveRecord it is supported, we have a config option
   # to enable it within PaperTrail.
--- a/test/paper_trail_test.rb
+++ b/test/paper_trail_test.rb
@@ -24,6 +24,14 @@ class PaperTrailTest < ActiveSupport::TestCase
     assert PaperTrail.enabled?
   end
 
+  test 'with_whodunnit' do
+    refute PaperTrail.whodunnit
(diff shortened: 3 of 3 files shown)


lib/paper_trail/frameworks/rails/controller.rb
```diff
@@ -61,14 +61,23 @@ def set_paper_trail_enabled_for_controller
...
+        if ::PaperTrail.enabled_for_controller?
+          ::PaperTrail.with_whodunnit(user_for_paper_trail, &block)
+        else
+          yield
+        end
       end
 
       # Tells PaperTrail any information from the controller you want to store
       # alongside any changes that occur.
       def set_paper_trail_controller_info
```
jaredbeck: To fix the failing tests, 

```
NameError: undefined local variable or method `block'
```

maybe you need to add `&block` to the arguments here?

jaredbeck:
Thanks for the PR! Let me see if I understand the motivation:

> if controller A sets whodunnit and then some other controller runs that does not include papertrail the whodunnit state lingers around, this also makes testing ugly

So, at the beginning of each controller test, you expected `whodunnit` etc. to be `nil`, but it had a value from a previous test?

If I'm understanding you correctly, I think I'd prefer to do something like this, clearing those variables when `enabled_for_controller?` is false.

```diff
diff --git a/lib/paper_trail/frameworks/rails/controller.rb b/lib/paper_trail/frameworks/rails/controller.rb
index e1569f0..9b4a215 100644
--- a/lib/paper_trail/frameworks/rails/controller.rb
+++ b/lib/paper_trail/frameworks/rails/controller.rb
@@ -62,13 +62,17 @@ module PaperTrail
 
       # Tells PaperTrail who is responsible for any changes that occur.
       def set_paper_trail_whodunnit
-        ::PaperTrail.whodunnit = user_for_paper_trail if ::PaperTrail.enabled_for_controller?
+        ::PaperTrail.whodunnit = ::PaperTrail.enabled_for_controller? ?
+          user_for_paper_trail :
+          nil
       end
 
       # Tells PaperTrail any information from the controller you want to store
       # alongside any changes that occur.
       def set_paper_trail_controller_info
-        ::PaperTrail.controller_info = info_for_paper_trail if ::PaperTrail.enabled_for_controller?
+        ::PaperTrail.controller_info = ::PaperTrail.enabled_for_controller? ?
+          info_for_paper_trail :
+          {}
       end
 
     end
```

This approach, if it meets your needs, adds far less complexity.

But, I don't really understand why this is an issue if we're using [request_store][1]. Maybe request_store doesn't work in controller tests?

[1]: https://github.com/steveklabnik/request_store

lib/paper_trail/frameworks/rails/controller.rb
grosser: thx, fixed

grosser:
I'm coming from an app where global state was always left behind, so I don't trust it / want to avoid messing it up :)

I guess using request_store solves that ... but now you added another gem to do something that is trivial to do in ruby ... also encouraged users to use a polluting pattern without offering any alternatives.

So if you think adding another dependency is less complexity then using around then maybe leave the controller extensions as they are, but add the with_xxx helpers ?
I'm using my own around_filter anyway since I did not realize there was this extensions.

jaredbeck:
> I'm coming from an app where global state was always left behind, so I don't trust it / want to avoid messing it up :)
>

I agree, I don't like this way of passing data from controller to model.  I'm open to alternatives.

It may be worth clarifying that PaperTrail doesn't use global variables, not the true global variables that begin with a dollar sign. The variables you are referring to are in a `RequestStore`.  They are not global.  (Unrelated to this PR, there is one class variable though, `@@config`, so that's not thread-safe.)
 
> I guess using request_store solves that ... but now you added another gem to do something that is trivial to do in ruby ... 

I think the author of request_store would argue that it's not trivial :)

> So if you think adding another dependency is less complexity then using around ..

Yes, in most situations I would rather add a dependency than add code because it is less code for me to maintain.

> .. then maybe leave the controller extensions as they are, but add the with_xxx helpers ?

I like how they're written, and each one on its own is fairly simple.  My only concern would be if we add them and then you end up being the only person who uses them.  But, I may just not understand the use case well enough.  What is a situation someone would use one of these methods in?  How would you describe these new features in the documentation?  Pretend you're writing a paragraph explaining them in the readme, if that helps.

grosser:
The problem this gem tries to address (counters / Thread[:current] ||= xxx)
is not the problem PaperTrail has, so I still think using around would be
simpler / 1 less dependency / more idiomatic / easier to reuse for workers
etc / a good example for others (clean up your state kids!).

If you insist on keeping it the readme could be:
To audit a block of code and not leave behind state (workers / migrations /
internal batch processing) use `PaperTrail.with_whodunnit('The user'){  ...
your code ... }`

batter:
> atm if controller A sets whodunnit and then some other controller runs that does not include papertrail the whodunnit state lingers around

PaperTrail doesn't get included into a controller FWIW, there are a couple helper methods made available in the source (such as the one that attempts to assign the `whodunnit` value, but it will get called on every controller run to my knowledge, I don't believe that there is an option to "opt out" so to speak. Also, there should only be one controller invoked per request as far as I can tell, so I'm not following what the issue is here exactly.  It doesn't feel like there's a lot of "pollution" so to speak.

There is also the [`whodunnit` method on `PaperTrail::Model::InstanceMethods`](https://github.com/airblade/paper_trail/blob/master/lib/paper_trail/has_paper_trail.rb#L280) which seems to accomplish mostly the same thing your PR does, so does that solve your issues?

grosser:
Atm I'm using a custom around_filter + setting whodunnit in my workers, but all with a with_whodunnit helper I added, since I don't want to keep state floating around by accident.

The request-state thingy also works, I just don't like it very much ... also if I want to use whodunnit in my workers I'd need the with_whodunnit helper.

jaredbeck:
> There is also the whodunnit method on PaperTrail::Model::InstanceMethods which seems to accomplish mostly the same thing your PR does, so does that solve your issues?

That's neat, I didn't know about that.  By comparison, Michael's proposed methods are defined on `::PaperTrail`, and he is proposing similar methods for `enabled` and `controller_info`.

What do you (both) think about combining Michael's proposed methods?

```ruby
# Temporarily overwrites variables like whodunnit within the provided block.
def self.with(whodunnit, controller_info, enabled)
  raise ArgumentError, 'Expected a block' unless block_given?
  old_whodunnit = PaperTrail.whodunnit
  old_controller_info = PaperTrail.controller_info
  old_enabled = PaperTrail.enabled
  PaperTrail.whodunnit = whodunnit
  PaperTrail.controller_info = controller_info
  PaperTrail.enabled = enabled
  yield self
ensure
  PaperTrail.whodunnit = old_whodunnit
  PaperTrail.controller_info = old_controller_info
  PaperTrail.enabled = old_enabled
end
```

A single method would be easier to use if one wanted to override multiple variables.

I still think there are more fundamental problems here: how we pass data from controller to model, and the fact that `PaperTrail.config` is backed by a class variable, which is not thread-safe, but I don't mind providing a workaround like the above, temporarily, until we can fix the more fundamental problems.

grosser:
So far I only needed with_whodunnit and with_enabled.
If they get combined then I'd suggest an options hash:

```Ruby
def with(options)
  if options.key?(:whodunnit)
   ....
  end
```

but that is pretty messy code too ... might be simpler to just do `with_xxx { with_yyy }`

PR #592, opened 2015-08-24
Add uuid support in version associations

msbrigna:


lib/paper_trail/version_concern.rb
```diff
@@ -375,14 +375,12 @@ def reify_has_many_directly(associations, model, options = {})
...
-          where("foreign_key_name = ?", assoc.foreign_key).
-          where("foreign_key_id = ?", model.id).
-          where("#{version_table_name}.item_type = ?", assoc.class_name).
-          where("created_at >= ? OR transaction_id = ?", options[:version_at], transaction_id).
-          group("item_id").to_sql
-        versions = model.class.paper_trail_version_class.where("id IN (#{version_id_subquery})").inject({}) do |acc, v|
+        version_id_subquery = "SELECT v1.id from #{model.class.version_association_name.table_name} v1 where exists (
+select distinct on (item_id) version_id, min(created_at) from #{model.class.version_association_name.table_name} v2 where v1.id = v2.version_id
+and foreign_key_name = '#{assoc.foreign_key}' and foreign_key_id = '#{model.id}' and #{version_table_name}.item_type = #{assoc.class_name} and #{version_table_name}.created_at >= '#{options[:version_at]}'
+OR transaction_id = #{transaction_id.blank? ? "null" : transaction_id} group by item_id, id)"
```
jaredbeck: Please format SQL.  A heredoc can be good for this:

```
my_sql = <-SQL
  select *
  from foo
SQL
```

Also, please hard-wrap lines at 80 characters where possible.

Also, we don't want to duplicate `model.class.version_association_name.table_name`.  You could extract a local variable to reduce this duplication.

batter: Agreed. My preference would be to continue using `Arel` for building these queries so that it is more readable as well. The raw SQL syntax with all the string interpolation / ruby value injection is very difficult to parse.  What happens when this query needs to be tweaked?

jaredbeck:
Tests are failing with:

```
ActiveRecord::StatementInvalid: Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'ON (item_id) version_id, MIN(created_at)
```

See Travis CI.

msbrigna:
Sorry, but I notice that the association table has lot of problem with polymorphic associations and with STI models. I will think about a more flexible implementation to handle them all.

Issue #590, opened 2015-08-18
Has Many Through Association via has_many: and belongs_to: on the join table


allavena:
I'm reopening #530 (https://github.com/airblade/paper_trail/issues/530) as I'm facing the same issue and don't believe the reference https://github.com/airblade/paper_trail/issues/466) addresses it (Or if it does, I've completely failed to understand)

```ruby
class Document < ActiveRecord::Base
  has_many :sections
  has_many :paragraphs, through: :sections
end
 
class Section < ActiveRecord::Base
  belongs_to :document
  has_many :paragraphs
end
 
class Paragraph < ActiveRecord::Base
  belongs_to :section
end
```

Paper_Trail 4.0.0 Beta doesn't cater this situation. It looks for foreign key of association Paragraph i.e, section_id in the through model i.e, Section which will not be available in the above mentioned case.

https://github.com/airblade/paper_trail/issues/466 mentions adding an additional column to 'versions' table. However, the data is already there. With the existing code, a query on section can already return its paragraphs and revisions. 

[If anything, may I suggest a note in the Readme that this case is not supported?]

Thanks
André

jaredbeck:
> Paper_Trail 4.0.0 Beta doesn't cater this situation. 

Currently, `paper_trail 4.0.0` is the latest.  If you are still using the beta, please upgrade.

> It looks for foreign key of association Paragraph i.e, section_id in the through model i.e, Section which will not be available in the above mentioned case.

What PaperTrail methods are you using?  `reify`?  You've given some ActiveRecord models, but I don't see anything from PaperTrail in your example.

allavena:
Bad copy/paste, sorry. I'm using paper trail 4.0.0, although same result with using master from git. 

    document = FactoryGirl.create(:document_with_sections_and_paragraphs)
    puts 'Updating some paragraph'
    document.sections[0].paragraphs[0].update_attributes(title: 'New paragraph 1 title')
    puts 'Updating a section name'
    document.sections[0].update_attributes(title: 'My new Section')
    puts 'Updating document itself'
    document.update_attributes(title: 'New document Title')
    document.sections[0].paragraphs[0].update_attributes(title: 'Very New paragraph 1 title')
    puts 'Previous version of document'
    pp document.versions.last.reify(:has_many => true)

This last line fails, in the rails console in development mode, with a 
NoMethodError: undefined method `section_id' for #<Section:0x00000006c6ad58>

[...]

PR #596, opened 2015-08-28
fix issue in #reify_has_many_through

theRealNG:
Fix for https://github.com/airblade/paper_trail/issues/590

The assoc.foreign_key holds true only when the association is as follows

```ruby
Book
has_many :authorships
has_many :authors, :through => :authorships

Authorships
belongs_to :book
belongs_to :author

Author
has_many :authorships
has_many :books, :through => :authorships
```

In this case the assoc.foreign_key and assoc.association_foreign_key are the same

But in the case of the following associations

```ruby
Document
has_many :sections
has_many :paragraphs, :through => :sections

Section
belongs_to :document
has_many :paragraphs

Paragraph
belongs_to :section
```

the assoc.foreign_key and assoc.association_foreign_key are different and assoc.association_foreign_key gets the correct list of associated foreign_keys for the model.

For Ref: http://www.rubydoc.info/docs/rails/3.0.0/ActiveRecord%2FReflection%2FAssociationReflection%3Aassociation_foreign_key

EDIT: 
When a nested has_many relation is present, using the "reify_has_manys" method to reify the has_many through associations.

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -417,7 +417,14 @@ module PaperTrail
       associations.each do |assoc|
         next unless assoc.klass.paper_trail_enabled_for_model?
         through_collection = model.send(assoc.options[:through])
-        collection_keys = through_collection.map { |through_model| through_model.send(assoc.foreign_key) }
+
+        # if the association is a has_many association again, then call reify_has_manys for each through_collection
+        if !assoc.source_reflection.belongs_to? && through_collection.present?
+          through_collection.each { |through_model| reify_has_manys(through_model,options) }
+          next
+        end
+
+        collection_keys = through_collection.map { |through_model| through_model.send(assoc.association_foreign_key)}
 
         version_id_subquery = assoc.klass.paper_trail_version_class.
           select("MIN(id)").
--- a/test/unit/model_test.rb
+++ b/test/unit/model_test.rb
@@ -1902,4 +1902,169 @@ class HasPaperTrailModelTransactionalTest < ActiveSupport::TestCase
       end
     end
   end
+
+  context "Models with nested has_many through relationships" do
+    setup { @chapter = Chapter.create(:name => 'ch_1') }
+
+    context "before any associations are created" do
+      setup do
+        @chapter.update_attributes(:name => "ch_2")
+        @ch_1 = @chapter.versions.last.reify(:has_many => true)
+      end
+
+      should "reify the record when reify is called" do
+        assert_equal "ch_1", @ch_1.name
+      end
+    end
+
+    context "after the first has_many through relationship is created" do
+      setup do
+        @chapter.update_attributes :name => "ch_2"
+        Timecop.travel 1.second.since
+        @chapter.sections.create :name => "section 1"
+        Timecop.travel 1.second.since
+        @chapter.sections.first.update_attributes :name => "section 2"
+        Timecop.travel 1.second.since
+        @chapter.update_attributes :name => "ch_3"
+        Timecop.travel 1.second.since
+        @chapter.sections.first.update_attributes :name => "section 3"
+      end
+
+      context "when reify is called" do
+        setup do
+          @chapter_2 = @chapter.versions.last.reify(:has_many => true)
+        end
+
+        should "show the value of the base record as it was before" do
+          assert_equal "ch_2", @chapter_2.name
+        end
+
+        should "show the value of the associated record as it was before the base record was updated" do
+          assert_equal ['section 2'], @chapter_2.sections.map(&:name)
+        end
+
+        context "to the version before the relationship was created" do
+          setup { @chapter_1 = @chapter.versions.second.reify(:has_many => true) }
+
+          should "not return any associated records" do
+            assert_equal 0, @chapter_1.sections.size
+          end
+        end
+
+        context "to the version before the associated record has been destroyed" do
+          setup do
+            @chapter.update_attributes :name => 'ch_3'
+            Timecop.travel 1.second.since
+            @chapter.sections.destroy_all
+            Timecop.travel 1.second.since
+            @chapter_3 = @chapter.versions.last.reify(:has_many => true)
+          end
+
+          should "return the associated record" do
+            assert_equal ['section 2'], @chapter_3.sections.map(&:name)
+          end
+        end
+
+        context "to the version after the associated record has been destroyed" do
+          setup do
+            @chapter.sections.destroy_all
+            Timecop.travel 1.second.since
+            @chapter.update_attributes :name => 'ch_4'
+            Timecop.travel 1.second.since
+            @chapter_4 = @chapter.versions.last.reify(:has_many => true)
+          end
+
+          should "return the associated record" do
+            assert_equal 0, @chapter_4.sections.size
+          end
+        end
+      end
+
+      context "after the nested has_many through relationship is created" do
+        setup do
+          @section = @chapter.sections.first
+          @paragraph = @section.paragraphs.create :name => 'para1'
+        end
+
+        context "reify the associations" do
+          setup do
+            Timecop.travel 1.second.since
+            @initial_section_name = @section.name
+            @initial_paragraph_name = @paragraph.name
+            @chapter.update_attributes :name => 'ch_5'
+            Timecop.travel 1.second.since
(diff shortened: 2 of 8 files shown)


jaredbeck:
Thanks for the PR!  Can you add a test that demonstrates the issue?

jaredbeck:
> Can you add a test that demonstrates the issue?

Just to clarify, we know that #590 is a confirmed bug, but I'd still be nice to have a test.  Thanks.

theRealNG:
Ya sure. Will work on it.

jaredbeck:
You should be able to use the failing test I already wrote (https://github.com/jaredbeck/pt_issue_590/blob/de54892426966d146a13c8c2382735826aa76660/spec/model/document_spec.rb) as a starting point.

test/unit/model_test.rb
```diff
@@ -1902,4 +1902,152 @@ class HasPaperTrailModelTransactionalTest < ActiveSupport::TestCase
...
+        @chapter.sections.create :name => "section 1"
+        Timecop.travel 1.second.since
+        @chapter.sections.first.update_attributes :name => "section 2"
+        Timecop.travel 1.second.since
+        @chapter.update_attributes :name => "ch_3"
+        Timecop.travel 1.second.since
+        @chapter.sections.first.update_attributes :name => "section 3"
+      end
+
+      context "when reified is called" do
```
jaredbeck: The method is called `reify`, not `reified`.  Please update the context string here.

test/unit/model_test.rb
```diff
@@ -1902,4 +1902,152 @@ class HasPaperTrailModelTransactionalTest < ActiveSupport::TestCase
...
+        setup do
+          @chapter_1 = @chapter.versions.last.reify(:has_many => true)
+        end
+
+        should "show the value of the base record as it was before" do
+          assert_equal "ch_2", @chapter_1.name
```
jaredbeck: If the chapter name is "ch_2", then the instance variable here should be named `@chapter_2`, not `@chapter_1` as you have.

lib/paper_trail/version_concern.rb
```diff
@@ -417,7 +417,14 @@ def reify_has_many_through(associations, model, options = {})
       associations.each do |assoc|
         next unless assoc.klass.paper_trail_enabled_for_model?
         through_collection = model.send(assoc.options[:through])
-        collection_keys = through_collection.map { |through_model| through_model.send(assoc.foreign_key) }
+
+        # if the association is a has_many association again, then call reify_has_manys for each through_collection
+        if !assoc.source_reflection.belongs_to? && through_collection.present?
+          through_collection.each { |through_model| reify_has_manys(through_model,options) }
+          return
```
jaredbeck: Does this return `nil`?  That seems different from the existing return type, which is `associations`, right? We probably don't really care about the return value, so maybe we should just always return `nil`?

jaredbeck:
@theRealNG this is looking good.  I appreciate how thorough you were with the tests.  My comments are mostly about style. I'm trying my best to understand the meat of the change, but I'd really like to get Ben's opinion as he is more familiar with the reify functions.

CC: @batter

test/unit/model_test.rb
batter: Agreed, the way you are modifying the attributes here vs the variable names in usage is a bit confusing. There are other spots int he test suite where we have similar confusing attribute modification vs variable names, specifically within this file, but that doesn't mean we should continue to follow that pattern.

lib/paper_trail/version_concern.rb
batter: Should this really be returning at this point?

theRealNG: Yes, So that the rest of the code is not executed. 
We will be getting the required records by calling "reify_has_manys".

theRealNG:
Bug: 
We are not handling nested has_many associations.

Fix:
For a nested has_many associations we can call "has_manys" method for each of the nested associations so that we handle any levels of nesting. 

Explanation with example :- 
* A chapter has many sections and a section has many paragraphs.
* So when we call "reify_has_many_directly" on the chapter object, we get a list of sections that are associated with the chapter. 
* Since the association between section and paragraphs is also a has_many association, we can call reify_has_many_directly on each of the sections we got from the above result to get all the paragraphs associated to that section and in turn associated to the chapter. 

But i'm calling "reify_has_manys" on each section so that we can handle any levels of nesting.

lib/paper_trail/version_concern.rb
batter: I guess my question / thought is, shouldn't the method still perform the same behavior it does prior to the return in these cases? In the normal / existing method [flow we assign the new collection of updated issues to the association definition](https://github.com/theRealNG/paper_trail/blob/issue%23590/lib/paper_trail/version_concern.rb#L458) and return it in one call... In this case is it not necessary to perform that since the collection is already assigned to the association, or am I missing something here?

theRealNG: The collection is assigned to the association in the "reify_has_many_directly" method also.

lib/paper_trail/version_concern.rb
```diff
@@ -417,7 +417,14 @@ def reify_has_many_through(associations, model, options = {})
...
+        end
+
+        collection_keys = through_collection.map { |through_model| through_model.send(assoc.association_foreign_key)}
```
jaredbeck: If I revert the change to this one line, by calling `assoc.foreign_key` instead of `assoc.association_foreign_key`, all the tests still pass.  So, is it necessary to change this line?

theRealNG: In the example of Document, Sections and paragraphs:
The through_model is a object of Section.

What we are trying to do is get the list of all the paragraphs associated to the section by calling:
section.paragraph_ids

If we do Section.reflect_on_association(:paragraphs).foreign_key, it returns 'section_id' and if we do Section.reflect_on_association(:paragraphs).association_foreign_key, it returns 'paragraph_id'. So we should be using 'association_foreign_key' instead of 'foreign_key'.

This was not an issue so far because in a 'belongs_to' association assoc.association_foreign_key and assoc.foreign_key are the same.

jaredbeck: That all makes sense, thanks.

lib/paper_trail/version_concern.rb
batter: I believe we want to use `next` instead of `return` here, because the code is executing through [a loop at this point and we don't want to prematurely return out of the method and break the loop](https://github.com/theRealNG/paper_trail/blob/issue%23590/lib/paper_trail/version_concern.rb#L417)

jaredbeck: > I believe we want to use next instead of return here, because the code is executing through a loop at this point and we don't want to prematurely return out of the method and break the loop

I agree.  Here's a test that fails with `return`, passes with `next`: https://github.com/theRealNG/paper_trail/pull/1

theRealNG: Yup @batter you are right. Thanks @jaredbeck

jaredbeck:
Ben, I think this is ready to merge, now that it includes https://github.com/theRealNG/paper_trail/pull/1

What do you think?

theRealNG:
Do you want me to squash the commits into a single commit?

jaredbeck:
> Do you want me to squash the commits into a single commit?

Sure, thanks.

batter:
:+1: Nice work, thanks to both of you.

My take is it's fine to leave it as a split commit as long as the merge has no fast forward but I don't care

batter:
@jaredbeck Did the test we wrote ever end up working out of curiosity?

jaredbeck:
> @jaredbeck Did the test we wrote ever end up working out of curiosity?

Yes, see https://github.com/theRealNG/paper_trail/pull/1

batter:
So what is our merge process now for these types of things @jaredbeck?

Merge to `master`, then cherry-pick / backport to `4.1-stable`, `4.1-stable`?

jaredbeck:
> So what is our merge process now for these types of things? 
> Merge to master, then cherry-pick / backport to [4.0-stable], 4.1-stable?

Yes, that's how I picture it working. I think it's a common strategy; it's what rails seems to do. One thing I'm not sure of is how to run CI for a backport. I guess we could do a PR for each backport..

batter:
ok that sounds good, just wanted to be clear about how you were envisioning it

PR #614, opened 2015-09-09
Enable control over the order of AR callbacks

Ninigi:
Ok, this is embarrassing, but it was actually easier for me to delete the fork and redo my changes than trying to get a grip on how to rework all those commits [in this PR] (https://github.com/airblade/paper_trail/pull/607) ... I left the PR as a reference, but I guess all you need to review is this.


Ninigi:
and messed it up again... *sigh* It's too late today and this is becoming quite frustrating... Guess I'll have to look into it tomorrow once again.

jaredbeck:
> sigh It's too late today and this is becoming quite frustrating

:( take a break.  When you come back, you'll get it :+1:

Ninigi:
OK I think this should be reviewable now... I left everything as it was wherever possible, so the changes should be obvious now.
Was good that I looked into it once again though, because there were some non-critical misstakes I removed before reopening this PR.

On a side-note, there was a case when a single test would fail on Travis-CI. I looked through the code and just couldn't find what could have caused it (was an "undefined method for nil-class" error), when I couldn't find anything I just reran the build and it passed without me changing anything. Any ideas where that came from?

jaredbeck:
Thank you for "minimizing" this PR by moving that other refactoring to a separate PR.

A few things I like:

- It makes callbacks more explicit and thus easier to understand.
- It gives the user control over the order of callbacks, making integrating with other libraries easier.
- It doesn't appear to break backwards compatibility.

A few things I don't like:

- The introduction of state, specifically the new `@_set_up` instance variable.  Is there a way we can do this without state?
- The `paper_trail_after_destroy` and `paper_trail_before_destroy` methods. I don't want a [Cambrian explosion][1] of methods for every conceivable combination of options.
- The handling of options.  I think I'd prefer if `has_paper_trail` remain mandatory, and to require that it come before the callbacks.  This may help to eliminate the proposed `@_set_up` state.

[1]: https://en.wikipedia.org/wiki/Cambrian_explosion

Ninigi:
Thanks for the feedback (I was really looking forward to this)
> The introduction of state, specifically the new @_set_up instance variable. Is there a way we can do this without state?
To be honest, I did not really like this either, I did it to work around the problem of breaking the original has_paper_trail way of doing it
> The paper_trail_after_destroy and paper_trail_before_destroy methods. I don't want a Cambrian explosion of methods for every conceivable combination of options.
No problem, I just added this to have a more explicit way of adding the callback instead of adding another "do this + option". If you don't like it, scratch it :)
> The handling of options. I think I'd prefer if has_paper_trail remain mandatory, and to require that it come before the callbacks. This may help to eliminate the proposed @_set_up state.
Love and hate.
Hate: I tried to do it without adding a state, but I found myself with either clattering the models with reduntant stuff, or missing some of the important setups.
Love: I chose the easiest way to do this, but I already planed on doing a revision and refine it, even if you had just accepted this PR, so it is nice to get the "go for a better implemantion, because we basically like it". Just asume "has_paper_trail" to be mandatory makes everything a lot easier too, I guess I did not see the forest because of all the trees here ( = german proverb saying "I did not see the obvious, because I was involved with the detail").

Since I felt like a complete idiot with the last PR and the way it went (it seems everyone but me knows how to easily sqash commits the way he wants), for the changes you want, should I do a new PR, update this one with a commit or do something else?

jaredbeck:
> .. should callbacks not take any options at all?

Good question.  Yes, I think the easiest thing for now would be to keep all of the options where they are, passed to `has_paper_trail`.  The only option passed to a callback would be the `:before` or `:after` passed to `paper_trail_destroy`.

> .. should I do a new PR, update this one with a commit or do something else?

Doesn't matter, whatever is easiest for you.

Ninigi:
OK, I am having a hard time trying to keep the traditional way of using has_paper_trail and the "call it explicitely" way... No matter how you look at it, there is no way around introducing some kind of "hack" to not break older versions.
Right now I am looking into a way to delete the set callbacks if callback-methods are used, but it is really hard. Give me some time for this, but I don't know if you would like my solution.

jaredbeck:
> Right now I am looking into a way to delete the set callbacks if callback-methods are used, but it is really hard.

Yeah, that sounds tough.  What about something like:

```ruby
has_paper_trail on: []
paper_trail_create
paper_trail_update
paper_trail_destroy
```

I'm assuming that passing `on: []` will prevent `has_paper_trail` from installing any callbacks.

jaredbeck:
Also, what do you think about naming the methods:

```
paper_trail_on_create
paper_trail_on_update
paper_trail_on_destroy
```

Adding the word "on" makes it sound more like a callback, and it's not much more typing.

Ninigi:
> Also, what do you think about naming the methods:
> ```Ruby
> paper_trail_on_create
> paper_trail_on_update
> paper_trail_on_destroy
> ```

I like that. Actually my very first shot at this was with methods named "record_create" etc until I realised there methods with those names already in use internally ;)

> ```Ruby
> has_paper_trail on: []
> paper_trail_create
> paper_trail_update
> paper_trail_destroy
> ```
> I'm assuming that passing on: [] will prevent has_paper_trail from installing any callbacks.

Yes it would, but it would make the usage a bit awkward, don't you think? I will put a little more time into trying to find a way to make the traditional has_paper_trail work with this. If I can't make it work, I'll do it with the empty :on option and maybe tackle this problem in another iteration.

Thanks for your feedback.

Ninigi:
Ok, so the way I did it will set all callbacks on `has_paper_trail` and then clear the callback_chain of those callbacks when you first call `paper_trail_on_*`.
I looked into the ActiveSupport code for this, they use the CallbackChain#delete method internally to do exactyl what I was trying to do with it and do so as long as the CallbackChain has been there, so it should be backwards compatible.
The methods I wrote for that could be removed if you decided to use has_paper_trail together with one of the paper_trail_on_* methods (meaning using has_paper_trail alone would not be allowed). They will have to stay if using has_paper_trail alone should set up all callbacks.

The way it would work now is 
```Ruby
class ModelTrackingCreateEvent < ActiveRecord::Base
  # Will track create only
  has_paper_trail :ignore => ['something'], ...
  paper_trail_on_create # no options possible
end

class ModelTrackingAllEvents
  has_paper_trail :ignore => ['something'], ...
end
```

jaredbeck:
>> Yeah, that sounds tough. What about something like:
>> 
>> ```
>> has_paper_trail on: []
>> paper_trail_create
>> paper_trail_update
>> paper_trail_destroy
>> ```
>
> it would make the usage a bit awkward, don't you think?
> .. the way I did it will set all callbacks on has_paper_trail and then clear the callback_chain of those callbacks when you first call paper_trail_on_*.

I don't like the complexity that this introduces.  I'm OK with a slightly harder-to-use API if it makes the library a lot easier to maintain.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -8,6 +9,7 @@ def self.included(base)
     end
 
     module ClassMethods
+      include Callbacks
```
jaredbeck: I'd rather keep the contents of `Callbacks` here in `ClassMethods`. I generally agree with Steve Klabnik's article "[Mixins: A refactoring anti-pattern](http://blog.steveklabnik.com/posts/2012-05-07-mixins--a-refactoring-anti-pattern)".

lib/paper_trail/callbacks.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    # Record version after "create" event
+    def paper_trail_on_create
+      cleanup_callback_chain
+
+      after_create :record_create,
+                   :if => :save_version?
+    end
+
+    private
```
jaredbeck: Per the [ruby style guide](https://github.com/bbatsov/ruby-style-guide) "Indent .. private methods as much as the method definitions they apply to".

lib/paper_trail/has_paper_trail.rb
```diff
@@ -46,6 +48,29 @@ module ClassMethods
...
+        # up the callback-methods is important to make them compatible with the
+        # traditional has_paper_trail method.
+        # This can move to setup_model_for_paper_trail, as soon as the
+        # cleanup_callback and setup_callbacks_from_options method is no longer
+        # needed.
+        class_attribute :paper_trail_options
+        self.paper_trail_options = {}
+
+        # TODO: Add a deprecation warning / info to use callback methods instead
+        # of the :on option?
```
jaredbeck: I doubt we'll be removing the `:on` option anytime soon.  Please remove comments that refer to the `:on` option being removed.

spec/support/callback_modifier.rb
```diff
@@ -0,0 +1,28 @@
...
+
+class CreateModifier < CallbackModifier
+  paper_trail_on_create
+end
+
+class DefaultModifier < CallbackModifier
+  # Because of the way I set up the destroy method for testing
+  # has_paper_trail has to be initialized in this model seperately
+  has_paper_trail
+end
```
jaredbeck: Why did you choose to put these classes in `spec/support` instead of in `test/dummy/app/models` like their parent `callback_modifier.rb`?

test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
```diff
@@ -277,5 +282,7 @@ def self.down
     remove_index :version_associations, :column => [:version_id]
     remove_index :version_associations, :name => 'index_version_associations_on_foreign_key'
     drop_table :version_associations
+    drop_table :filter_modifier
```
jaredbeck: I think this like is probably an oversight. We can remove it, right?

Ninigi: Ohhh.. Yeah that one is very old. No one will miss that line ;)

spec/support/callback_modifier.rb
Ninigi: Because for me they are more like helper-classes wrapped around the model, it didn't feel right to put them there with all the "normal" activerecord models.

Ninigi:
All right, I'll just admit defeat on the callback-chain thing ;) I have a way of keeping the library maintainable in mind, but I feel like I am trying to force this instead of accepting it is an unnecessary addition and just do it the easiest way, like you suggested.

I moved the methods back into `lib/has_paper_trail.rb`. I disagree that having all methods in a file adds to maintainability or readability if it inflates it to 500 lines of code and more, but I think you are right that those methods belong into the `ClassMethods` module.

lib/paper_trail/callbacks.rb
Ninigi: Ah, haha sorry... I am using the [rails-style](https://github.com/bbatsov/rubocop/issues/1342) as a default for my rubocop and I am so used to it, I don't think about it.

Anyways, there are no more private methods now.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ None
 
 ### Added
 
+- Added callback-methods `paper_trail_update` `paper_trail_create` `paper_trail_destroy`
+  instead of has_paper_trail
+  [#593](https://github.com/airblade/paper_trail/pull/607)
```
jaredbeck: - Should read `paper_trail_on_update`.
- PR number in link should be 614, right?

README.md
```diff
@@ -330,6 +330,25 @@ a.versions.size                           # 3
 a.versions.last.event                     # 'update'
 ```
 
+You can also use the corresponding callback-methods seperately instead of using
+the :on option. If you choose to use the callback-methods, PaperTrail will only
+track the according events - so `paper_trail_on_create` is basically the same as
+`has_paper_trail :on => :create`.
```
jaredbeck: I'd put a heading like "Controlling the Order of AR Callbacks" above this paragraph, as (for me) that's the point of this new feature.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -87,26 +100,46 @@ def has_paper_trail(options = {})
...
+        options_on.each do |option|
+          send "paper_trail_on_#{option}"
+        end
+
+        paper_trail_options[:on] = options_on
+      end
+
+      # Record version before or after "destroy" event
+      def paper_trail_on_destroy(recording_order = 'after')
+        unless %(after before).include?(recording_order.to_s)
```
jaredbeck: Should be `%w[after before]`.

```
%(a b) #=> "a b"
%w[a b] #=> ["a", "b"]
```

The `%w` is the important part. The square brackets are a stylistic choice, reminding us that it returns an array.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -87,26 +100,46 @@ def has_paper_trail(options = {})
...
+          fail ArgumentError, 'recording order can only be "after" or "before"'
```
jaredbeck: Nice use of `fail` here, and I appreciate the validation of arguments.

spec/spec_helper.rb
```diff
@@ -37,6 +37,8 @@
     # `true` in RSpec 4.
     mocks.verify_partial_doubles = true
   end
+  config.filter_run :focus
+  config.run_all_when_everything_filtered = true
```
jaredbeck: Please don't change the spec_helper as part of this PR.  If this is a change that matters to you, please make a separate PR for it.

spec/support/callback_modifier.rb
jaredbeck: I see your point, but I'd still rather keep all the test models in one place.

test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
jaredbeck: Please don't forget to delete this line.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -87,26 +100,46 @@ def has_paper_trail(options = {})
...
         after_rollback :reset_transaction_id
         after_rollback :clear_rolled_back_versions
       end
 
+      def setup_callbacks_from_options(options_on = [])
+        options_on.each do |option|
+          send "paper_trail_on_#{option}"
+        end
+
+        paper_trail_options[:on] = options_on
```
jaredbeck: Isn't this line unnecessary because of the assignment in `setup_model_for_paper_trail`? (This assignment: `self.paper_trail_options = options.dup`)

Also, this line has got me thinking .. what about the case like this:

```
has_paper_trail on: []
paper_trail_on_destroy
```

The result should be that `paper_trail_options[:on] == [:destroy], but I don't think this PR will do that.  Shouldn't the callback methods like `paper_trail_on_destroy` append to `paper_trail_options[:on]`?

Ninigi: > Isn't this line unnecessary because of the assignment in `setup_model_for_paper_trail?` (This assignment: `self.paper_trail_options = options.dup`)

Yes. I think this was from an older version and is no longer needed/redundant now. Removed it.

> The result should be that `paper_trail_options[:on] == [:destroy]`

Agreed and added. I used a guard clause before adding the option-symbol to `paper_trail_options[:on]`, so for example `paper_trail_on_create` would return nil if :on => :create has already been set and the options array if not. Not sure if there is any case anyone would ever want to use the return for anything, but it's a little like `require` behaviour (returning false when already required).

spec/spec_helper.rb
Ninigi: Doesn't matter to me. Removed.

lib/paper_trail/has_paper_trail.rb
Ninigi: Must have been a typo, I didn't even know %(...) existed. Changed to your suggestion.

Is this new by the way? My Rubocop claims "%w-literals should be delimited by '(' and ')'", even though rubocop specs state otherwise and I think the [] makes a lot more sense...

README.md
Ninigi: Added. Since I am not an english native-speaker, every suggestion to my readme addition is much appreciated.

jaredbeck:
Looks good. Thanks Fabian! This will be a big help to people who need to control the order of callbacks.

PR #615, opened 2015-09-11
Add note about location of 'has_paper_trail' in class

radionewzealand:
Paper trail can clash with other callbacks if it is not called first in the class.

Refer to: http://stackoverflow.com/questions/32417619/tests-fail-with-has-paper-trail

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,9 +134,12 @@ end
 When you declare `has_paper_trail` in your model, you get these methods:
 
 ```ruby
+# has_paper-trail should be placed at the top of the class declaration (before any other code)
+# as it needs to register its callbacks first.
 class Widget < ActiveRecord::Base
   has_paper_trail   # you can pass various options here
 end
+
 
 # Returns this widget's versions.  You can customise the name of the association.
 widget.versions


jaredbeck:
```diff
+# has_paper-trail should be placed at the top of the class declaration (before any other code)
+# as it needs to register its callbacks first.
```

This statement is not true of all applications.

A better statement would be "has_paper-trail installs a number of callbacks.  If there are other callbacks in your model, be aware of any potential interactions.  The order of callbacks may matter."

In the future, where appropriate, you can add `[ci skip]` to your commit messages to skip the test run.

jaredbeck:
Closed via 64c4a5b

radionewzealand:
OK, thanks for the feedback and not about CI. You want me to create a new PR?

jaredbeck:
> You want me to create a new PR?

No need, unless you think https://github.com/airblade/paper_trail/commit/64c4a5bbe992ee048900ca7152296fedabe91045 doesn't explain it.

radionewzealand:
Looks good to me. Cheers!

PR #617, opened 2015-09-16
Install debugger: pry

jaredbeck:
I find pry helpful, is it OK if we add it as a development dependency?


batter:
Sounds good to me

PR #623, opened 2015-09-23
Specify required ruby version of 1.9 or greater

batter:
On the Gemspec.. I meant to add this to the branch for #620 but didn't get around to it in time. We should add this in as well. Also I wonder whether we should hold off on merging the removal of 1.8.x into master until we get closer to v5.x?

### DIFF
--- a/paper_trail.gemspec
+++ b/paper_trail.gemspec
@@ -18,6 +18,7 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
 
   s.required_rubygems_version = '>= 1.3.6'
+  s.required_ruby_version = '>= 1.9'
 
   s.add_dependency 'activerecord', ['>= 3.0', '< 6.0']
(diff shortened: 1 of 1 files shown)


jaredbeck:
> We should add this in as well. 

Good catch.

> Also I wonder whether we should hold off on merging the removal of 1.8.x into master until we get closer to v5.x?

I don't think that will make backports any easier, as we already have a mix of 4.1 and 5.0 in master.  Or, we could go straight from 4.0 to 5.0.  I have no problem with that.

jaredbeck:
>> Also I wonder whether we should hold off on merging the removal of 1.8.x into master until we get closer to v5.x?
> 
> I don't think that will make backports any easier, as we already have a mix of 4.1 and 5.0 in master. Or, we could go straight from 4.0 to 5.0. I have no problem with that.

Also, leaving PRs laying around causes conflicts, so I'd rather merge them as soon as they're ready.

PR #627, opened 2015-10-01
Lets the users know that they need to generate a new migration which includes the fields used for meta

heridev:
Remember the users to generate a new migration if they want to metadata.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1057,6 +1057,15 @@ end
 PaperTrail will call your proc with the current article and store the result in
 the `author_id` column of the `versions` table.
 
+### Columns
+Remember to add the columns to the versions table, for instance using migrations:
+```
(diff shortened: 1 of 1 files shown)


jaredbeck:
Thanks for the suggestion.  I feel like this is already explained well by this line:

```
 PaperTrail will call your proc with the current article and store the result in
 the `author_id` column of the `versions` table.
```

But, if you wanted to add a single sentence, like:

```diff
 PaperTrail will call your proc with the current article and store the result in
 the `author_id` column of the `versions` table.
+ Don't forget to add any such columns to your `versions` table.
```

I would be OK with that.

heridev:
@jaredbeck makes sense, so I'll close this PR and I'll create a new one adding only that line

PR #629, opened 2015-10-02
Switch callbacks from `_filter` to `_action`

jashank:
Rails 5, in rails/rails@7644a99d, deprecated controller callbacks suffixed `_filter` in favour of those suffixed `_action`; this PR follows suit.

### DIFF
--- a/lib/paper_trail/frameworks/rails/controller.rb
+++ b/lib/paper_trail/frameworks/rails/controller.rb
@@ -3,9 +3,23 @@ module PaperTrail
     module Controller
 
       def self.included(base)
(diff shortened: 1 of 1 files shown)


jaredbeck:
The way this is written would mean dropping support for rails 3, right? Can this be written to support both rails 3 and 5?

jashank:
@jaredbeck Yes, it can.

jaredbeck:
looks good to me, thanks! We should probably add a rails 5 (or rails master) gemfile to our test suite, similar to the rails 3 gemfile we already have.

batter:
Nice, and I appreciate the backwards compatibility

PR #638, opened 2015-10-20
Fix issue 633 re: reification of through assoc.

### LINKED ISSUE #633: Has many through doesn't reify children
Here's a failing test showing when has_many through's source association is destroyed, it can no longer be accessed via `reify`.

```diff
diff --git a/test/unit/associations_test.rb b/test/unit/associations_test.rb
index ef70ed6..2cc05c6 100644
--- a/test/unit/associations_test.rb
+++ b/test/unit/associations_test.rb
@@ -633,7 +633,20 @@ class AssociationsTest < ActiveSupport::TestCase
             end
           end

-          context "destroy a section" do
+          context "the version before a section is destroyed" do
+            should "have the section and paragraph" do
+              @chapter.update_attributes(:name => 'chapter 6')
+              Timecop.travel 1.second.since
+              @section.destroy
+              assert_equal 4, @chapter.versions.size
+              chapter_v3 = @chapter.versions[3].reify(:has_many => true)
+              assert_equal 1, chapter_v3.sections.size
+              assert_equal 1, chapter_v3.paragraphs.size
+            end
+          end
+
+          context "the version after a section is destroyed" do
             should "not have any sections or paragraphs" do
               @section.destroy
               Timecop.travel 1.second.since
```

jaredbeck:
Hopefully fixes #633 

@tlynam Todd can you give this a try please?

```ruby
gem 'paper_trail', github: 'airblade/paper_trail', branch: 'fix_issue_633'
```

Thanks.

### DIFF
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -219,6 +219,16 @@ module PaperTrail
             through_collection.each do |through_model|
               reify_has_manys(transaction_id, through_model, options)
             end
+
+            # At this point, the "through" part of the association chain has
+            # been reified, but not the final, "target" part. To continue our
+            # example, `model.sections` (including `model.sections.paragraphs`)
+            # has been loaded. However, the final "target" part of the
+            # association, that is, `model.paragraphs`, has not been loaded. So,
+            # we do that now.
+            collection = through_collection.flat_map { |through_model|
+              through_model.public_send(assoc.name.to_sym).to_a
+            }
           else
             collection_keys = through_collection.map { |through_model|
               through_model.send(assoc.association_foreign_key)
@@ -234,8 +244,11 @@ module PaperTrail
             versions = versions_by_id(assoc.klass, version_id_subquery)
             collection = Array.new assoc.klass.where(assoc.klass.primary_key => collection_keys)
             prepare_array_for_has_many(collection, options, versions)
-            model.send(assoc.name).proxy_association.target = collection
           end
+
+          # To continue our example above, assign to `model.paragraphs` the
+          # `collection` (an array of `Paragraph`s).
+          model.send(assoc.name).proxy_association.target = collection
         end
       end
 
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -60,6 +60,14 @@ ActiveRecord::Schema.define(version: 20110208155312) do
     t.boolean "deleted",      default: false
   end
 
+  create_table "chapters", force: :cascade do |t|
+    t.string "name"
+  end
+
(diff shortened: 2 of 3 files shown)


jaredbeck:
Huh, I wonder why this doesn't work with MySQL? (https://travis-ci.org/airblade/paper_trail/builds/86319755) All of the other RDBMS' pass.

lib/paper_trail/reifier.rb
```diff
@@ -219,6 +219,16 @@ def reify_has_many_through(transaction_id, associations, model, options = {})
...
+
+            # At this point, the "through" part of the association chain has
+            # been reified, but not the final, "target" part. To continue our
+            # example, `model.sections` (including `model.sections.paragraphs`)
+            # has been loaded. However, the final "target" part of the
+            # association, that is, `model.paragraphs`, has not been loaded. So,
+            # we do that now.
+            collection = through_collection.map { |through_model|
+              through_model.public_send(assoc.name.to_sym).to_a
+            }.flatten
```
seanlinsley: `flat_map` is a slightly more efficient version that does both `map` and `flatten` in a single operation. It's more efficient because it doesn't have to create the intermediary array that's then discarded.

jaredbeck: Thanks! I switched it to `flat_map`.

jaredbeck:
> Huh, I wonder why this doesn't work with MySQL?

I discovered that MySQL timestamp columns do not support fractional seconds unless defined with "fractional seconds precision".

> .. when MySQL [5.6] stores a value into a column of any temporal data type, it discards any fractional part and does not store it. 
> ...  To define a column that includes a fractional seconds part, use the syntax type_name(fsp), where type_name is TIME, DATETIME, or TIMESTAMP, and fsp is the fractional seconds precision. For example:
> CREATE TABLE t1 (t TIME(3), dt DATETIME(6));
> https://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html

This causes trouble because reification of associations uses timestamp comparison (see `created_at >= ?` in e.g. `reify_has_many_directly`). I worked around the problem here with liberal use of `Timecop.travel 1.second.since`.

We should see if it is possible to add fractional seconds precision to our `versions.created_at` column. In the mean-time, MySQL users should add fractional seconds precision manually.

jaredbeck:
> We should see if it is possible to add fractional seconds precision to our versions.created_at column. In the mean-time, MySQL users should add fractional seconds precision manually.
>

MySQL users should also upgrade to rails 4.2, which is the first version of ActiveRecord with support for fractional seconds in MySQL (https://github.com/rails/rails/pull/14359)

PR #640, opened 2015-11-01
In track_associations, only check db if no config

jaredbeck:
This allows users who need to precompile assets without a database
to configure `track_associations = false`.

See discussion: https://github.com/airblade/paper_trail/issues/636

### DIFF
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -32,7 +32,9 @@ module PaperTrail
     end
 
     def track_associations
-      @track_associations ||= PaperTrail::VersionAssociation.table_exists?
+      @track_associations.nil? ?
+        PaperTrail::VersionAssociation.table_exists? :
+        @track_associations
     end
     alias_method :track_associations?, :track_associations
 


lib/paper_trail/config.rb
```diff
@@ -32,7 +32,9 @@ def serialized_attributes=(value)
     end
 
     def track_associations
-      @track_associations ||= PaperTrail::VersionAssociation.table_exists?
+      @track_associations.nil? ?
+        PaperTrail::VersionAssociation.table_exists? :
+        @track_associations
```
batter: We could actually simplify this to `@track_associations && PaperTrail::VersionAssociation.table_exists?`, right?  Not as explicitly obvious what is going on, but my thought is that this will actually prevent errors in scenarios where users try to turn the feature on without having the underlying required table.  Alternatively we could raise an error or a warning in those scenarios. Thoughts?

jaredbeck: > We could actually simplify this to @track_associations && PaperTrail::VersionAssociation.table_exists?, right?

That conjunction would require a database connection, and thus would not fix https://github.com/airblade/paper_trail/issues/636

batter:
Ahh I see

batter:
I'm thinking it may make sense to merge this upstream into the `4.0-stable` branch and have it be merged into `4.0.1`, since it is a bug fix. Thoughts?

jaredbeck:
> I'm thinking it may make sense to merge this .. into the 4.0-stable branch and have it be merged into 4.0.1, since it is a bug fix.

Sounds good. I'll go ahead and do that merge if you haven't already, and I'll add a line to the changelog about it. What's your timeline for a 4.0.1 release, because I think we've done enough work to justify one, it's just a matter of tracking down what we need to backport.

PR #641, opened 2015-11-02
Use class inst. var. instead of class var.

jaredbeck:
I use the term "class instance variable" and "class variable"
even though `PaperTrail` is a module. This nomenclature is
conventional in ruby.

```ruby
module A; @@x = :x; puts defined?(@@x); end
#=> "class variable"
```

Actually, because `PaperTrail` is a module, the use of a class
variable here is not as problematic as in an actual class, so
this change is solely to conform the the Ruby Style Guide.

> Avoid the usage of class (@@) variables due to their "nasty"
> behavior in inheritance.
> https://github.com/bbatsov/ruby-style-guide#no-class-vars

### DIFF
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -126,9 +126,9 @@ module PaperTrail
 
   # Returns PaperTrail's configuration object.
   def self.config
-    @@config ||= PaperTrail::Config.instance
-    yield @@config if block_given?
-    @@config
+    @config ||= PaperTrail::Config.instance
+    yield @config if block_given?
+    @config
   end
 
   class << self


batter:
My only question about this is regarding thread safety. I sone more or less thread safe?  I know we brought in `request_store` for thread safety concerns regarding the store, does it make sense to consider that here too?

jaredbeck:
> My only question about this is regarding thread safety.

That's an important question. This should have no impact on thread safety.  Both class variables (CV) and class instance variables (CIV) are equally thread-unsafe. That's why we use `RequestStore`.

In a class, there's actually a small difference between CV and CIV (In short, CV are inherited/shared and CIV are not) but in a module there is no difference, because you can't inherit from a module. So, this change should have no effect.

Basically, the ruby style guide says to never use CV (https://github.com/bbatsov/ruby-style-guide#no-class-vars), so this change just brings us in conformance with the style guide.

PR #642, opened 2015-11-02
Extract class: RecordHistory

jaredbeck:
The methods `#index`, `.subsequent`, and `.preceding` contain
duplicated logic re: how to order, chronologically, a set of
versions, specifically a set representing the history of a single
record.

The extracted class only supports `#index` so far, but it should
be possilbe to use in `.subsequent` and `.preceding` also.

### DIFF
--- a/lib/paper_trail/record_history.rb
+++ b/lib/paper_trail/record_history.rb
@@ -1,0 +1,59 @@
+module PaperTrail
+
+  # Represents the history of a single record.
+  # @api private
+  class RecordHistory
+
+    # @param versions - ActiveRecord::Relation - All versions of the record.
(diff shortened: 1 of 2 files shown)


jaredbeck:
@batter Ben what do you think of this design, particularly the constructor? Will this new object be useful elsewhere, do you think? Specifically, in the class methods `.subsequent` and `.preceding`, I hope.

batter:
I like it. We could also use `attr_reader` and/or `attr_accessor` if we think that the attributes being fed in are useful to have accessible from elsewhere.  But I can see a lot of usages for cleaning stuff up here.

jaredbeck:
Cool, I'll go ahead and merge this, thanks!

PR #643, opened 2015-11-02
Ruby style guide

jaredbeck:
I've been happy on other projects using a linter for ruby style, even making the linter part of the build process. For example, a linter can nitpick PRs in a way that would be annoying coming from me. The trick, as usual where code style is concerned, is agreeing on a style :) I like most of [bbatsov/ruby-style-guide][1], and the associated linter [bbatsov/rubocop][2].  What do you like, Ben? @batter 

[1]: https://github.com/bbatsov/ruby-style-guide
[2]: https://github.com/bbatsov/rubocop

### DIFF
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -101,7 +101,8 @@ module PaperTrail
   end
 
   def self.active_record_protected_attributes?
-    @active_record_protected_attributes ||= ::ActiveRecord::VERSION::MAJOR < 4 || !!defined?(ProtectedAttributes)
+    @active_record_protected_attributes ||= ::ActiveRecord::VERSION::MAJOR < 4 ||
+      !!defined?(ProtectedAttributes)
   end
 
   def self.transaction?
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -122,7 +122,7 @@ module PaperTrail
   # Thread-safe hash to hold PaperTrail's data. Initializing with needed
   # default values.
   def self.paper_trail_store
-    RequestStore.store[:paper_trail] ||= { :request_enabled_for_controller => true }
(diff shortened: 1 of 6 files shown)


batter:
Hey, I don't have much experience working with a strict linter other than some basic Javascript linting within a webpack compiled project, but I do realize that the library could be up to slightly better code standards so I don't have an objection with a linter.  I'd like to examine this a little more closely when I have some time later in the week.

lib/paper_trail.rb
```diff
@@ -101,7 +101,8 @@ def self.serializer
   end
 
   def self.active_record_protected_attributes?
-    @active_record_protected_attributes ||= ::ActiveRecord::VERSION::MAJOR < 4 || !!defined?(ProtectedAttributes)
+    @active_record_protected_attributes ||= ::ActiveRecord::VERSION::MAJOR < 4 ||
+      !defined?(ProtectedAttributes).nil?
```
batter: I actually think `!!defined?(ProtectedAttributes)` is clearer than the double negative of `!defined?(ProtectedAttributes).nil?`

jaredbeck: OK, I reverted that change.

jaredbeck:
Other than the double-negative (https://github.com/airblade/paper_trail/pull/643#discussion_r43921922), did you have any concerns? Mind if I merge?

Note that this PR just adds `rubocop` as a dev. dependency, but doesn't introduce any linting to the build process yet. So, this PR just makes a linter available to use, but not required.

batter:
Seems reasonable to me, at least gives us the option. I know that the style and code cleanliness has been lacking within the project over the years so anything we can do to help improve code quality and modularity is good in my book.

PR #648, opened 2015-11-11
make 100% sure that the warning is accurate

robmathews:
This PR addresses this issue by making sure the warning is only displayed when in fact the before_filter for set_paper_trail_whodunnit was not installed.


jaredbeck:
Looks good, thanks!

PR #661, opened 2015-11-28
Code style: break long lines

jaredbeck:
Limits maximum line length to 100 characters.  The ruby style guide recommends 80 characters (https://github.com/bbatsov/ruby-style-guide#80-character-limits), but that would be a big undertaking.

In a few places, extracting a method was the best way to break a long line. I've made separate commits for each of those method extractions. Some of the new methods were hard to name. Also, I'm not thrilled about adding more methods to `PaperTrail::Model`. We already add a lot of clutter to people's AR models. I'm open to other ideas about where these methods could live.

### DIFF
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -40,7 +40,8 @@ module PaperTrail
 
     # Indicates whether PaperTrail is on or off.
     def enabled
-      PaperTrail.paper_trail_store[:paper_trail_enabled].nil? || PaperTrail.paper_trail_store[:paper_trail_enabled]
+      PaperTrail.paper_trail_store[:paper_trail_enabled].nil? ||
+        PaperTrail.paper_trail_store[:paper_trail_enabled]
     end
 
     def enabled= enable
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -38,10 +38,10 @@ module PaperTrail
     end
     alias_method :track_associations?, :track_associations
 
-    # Indicates whether PaperTrail is on or off.
+    # Indicates whether PaperTrail is on or off. Default: true.
     def enabled
-      PaperTrail.paper_trail_store[:paper_trail_enabled].nil? ||
-        PaperTrail.paper_trail_store[:paper_trail_enabled]
+      value = PaperTrail.paper_trail_store.fetch(:paper_trail_enabled, true)
+      value.nil? ? true : value
     end
 
     def enabled= enable
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -212,14 +212,7 @@ module PaperTrail
     # not have an `object_changes` text column.
     def changeset
       return nil unless self.class.column_names.include? 'object_changes'
-      _changes = HashWithIndifferentAccess.new(object_changes_deserialized)
-      @changeset ||= _changes.tap do |changes|
-        if PaperTrail.serialized_attributes?
-          item_type.constantize.unserialize_attribute_changes_for_paper_trail!(changes)
-        end
-      end
-    rescue
-      {}
+      @changeset ||= load_changeset
     end
 
     # Returns who put the item into the state stored in this version.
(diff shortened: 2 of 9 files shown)


paper_trail.gemspec
```diff
@@ -37,7 +37,9 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'pry-nav', '~> 0.2.4'
   s.add_development_dependency 'rubocop', '~> 0.35.1'
 
-  # Allow time travel in testing. timecop is only supported after 1.9.2 but does a better cleanup at 'return'
+  # Allow time travel in testing. timecop is only supported after 1.9.2 but
+  # does a better cleanup at 'return'.
+  # TODO: We can remove delorean, as we no longer support ruby < 1.9.3
   if RUBY_VERSION < "1.9.2"
     s.add_development_dependency 'delorean'
   else
```
seanlinsley: `required_ruby_version` should be updated to `>= 1.9.3` then

lib/paper_trail/version_concern.rb
```diff
@@ -210,9 +212,8 @@ def reify(options = {})
...
     def changeset
       return nil unless self.class.column_names.include? 'object_changes'
-
-      _changes = self.class.object_changes_col_is_json? ? object_changes : PaperTrail.serializer.load(object_changes)
-      @changeset ||= HashWithIndifferentAccess.new(_changes).tap do |changes|
+      _changes = HashWithIndifferentAccess.new(object_changes_deserialized)
+      @changeset ||= _changes.tap do |changes|
         if PaperTrail.serialized_attributes?
           item_type.constantize.unserialize_attribute_changes_for_paper_trail!(changes)
         end
```
seanlinsley: Since this is purely a code style PR, I'm chiming in here to say that IMO `tap` is an anti-pattern because it adds one more thing to understand and read without adding much funcationality. I'd personally prefer this code style:
```ruby
@changeset ||= begin
  changes = HashWithIndifferentAccess.new(object_changes_deserialized)
  if PaperTrail.serialized_attributes?
    item_type.constantize.unserialize_attribute_changes_for_paper_trail!(changes)
  end
  changes
end
```

It also has the performance improvement of only having to run `HashWithIndifferentAccess.new(object_changes_deserialized)` if `changeset` hasn't been called before, whereas currently it will be run every time `changeset` is called.

lib/paper_trail/config.rb
```diff
@@ -40,7 +40,8 @@ def track_associations
 
     # Indicates whether PaperTrail is on or off.
     def enabled
-      PaperTrail.paper_trail_store[:paper_trail_enabled].nil? || PaperTrail.paper_trail_store[:paper_trail_enabled]
+      PaperTrail.paper_trail_store[:paper_trail_enabled].nil? ||
+        PaperTrail.paper_trail_store[:paper_trail_enabled]
```
seanlinsley: The behavior here can more succinctly be written as:
```ruby
PaperTrail.paper_trail_store.fetch(:paper_trail_enabled, true)
```
Where the second argument, true, is used if nothing was set.

paper_trail.gemspec
jaredbeck: Updated in https://github.com/airblade/paper_trail/pull/666, thanks.

lib/paper_trail/version_concern.rb
jaredbeck: Good suggestion.  I like it, though I prefer to extract a new method rather than have an anonymous block that long.

lib/paper_trail/config.rb
jaredbeck: Re: second argument, that is how `Hash#fetch` works but it appears that `RequestStore.fetch` is different.

- Implementation: https://github.com/steveklabnik/request_store/blob/160c635efbc16d9c16063e34f9d15f6673299c18/lib/request_store.rb#L34
- Example: https://github.com/steveklabnik/request_store/blob/160c635efbc16d9c16063e34f9d15f6673299c18/test/request_store_test.rb#L44

So, I guess we'd do something like:

```ruby
PaperTrail.paper_trail_store.fetch(:paper_trail_enabled) { true }
```

Does that sound right?

jaredbeck: Nevermind, I'm looking at `RequestStore.fetch`, but your suggestion uses `Hash#fetch` after all. I'll try it.

jaredbeck:
Thanks for the feedback, I'm going to merge this.  

I'd still appreciate any suggestions regarding my concern over cluttering models.

> In a few places, extracting a method was the best way to break a long line .. I'm not thrilled about adding more .. clutter to people's AR models. I'm open to other ideas about where these methods could live.

To clarify, the methods in question are 

- pt_record_object_changes?
- pt_recordable_object
- pt_recordable_object_changes

If y'all have some ideas about a better place for them to live, let me know. One option could be to extract the `record_*` methods into a new class. Then we'd be free to have lots of little methods without worrying about cluttering people's models.

PR #662, opened 2015-12-01
Better migration for object from text to jsonb

denispeplin:
It probably can be extended with data migration, but still it better than raw SQL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1103,13 +1103,19 @@ column type.
 #### Convert a column from text to json
 
 Postgres' `alter column` command will not automatically convert a `text`
-column to `json`, but it can still be done with plain SQL.
+column to `json`, but it can still be done with a piece of plain SQL.
 
-```sql
-alter table versions
-alter column object type json
-using object::json;
+```ruby
+def up
+  change_column :versions, :object, 'jsonb USING object::jsonb'
+end
+
+def down
+  change_column :versions, :object, 'text USING object::text'
+end
 ```
+
+You'll have to destroy or migrate all `versions` data before running this migration.
 
 ## SerializedAttributes support
 
--- a/README.md
+++ b/README.md
@@ -1102,16 +1102,15 @@ column type.
 
 #### Convert a column from text to json
 
-Postgres' `alter column` command will not automatically convert a `text`
-column to `json`, but it can still be done with a piece of plain SQL.
-
 ```ruby
-def up
-  change_column :versions, :object, 'jsonb USING object::jsonb'
-end
+class ConvertVersionsObjectToJson < ActiveRecord::Migration
+  def up
+    change_column :versions, :object, 'jsonb USING object::jsonb'
+  end
 
-def down
-  change_column :versions, :object, 'text USING object::text'
+  def down
+    change_column :versions, :object, 'text USING object::text'
+  end
 end
 ```
 


README.md
```diff
@@ -1103,14 +1103,20 @@ column type.
...
+def up
+  change_column :versions, :object, 'jsonb USING object::jsonb'
+end
+
+def down
+  change_column :versions, :object, 'text USING object::text'
+end
 ```
 
+You'll have to destroy or migrate all `versions` data before running this migration.
```
jaredbeck: Are you sure? I tested the `alter table` with postgres 9.4 and it seems to convert correctly:

```sql
create table heyojson ( data text not null );
insert into heyojson (data) values ('{ "foo" : "bar" }');
alter table heyojson alter column data type jsonb using data::jsonb;
select * from heyojson;
-- {"foo": "bar"}
```

README.md
```diff
@@ -1103,14 +1103,20 @@ column type.
 #### Convert a column from text to json
 
 Postgres' `alter column` command will not automatically convert a `text`
-column to `json`, but it can still be done with plain SQL.
+column to `json`, but it can still be done with a piece of plain SQL.
```
jaredbeck: If we're going to provide an AR migration here instead of plain SQL, I think we can delete this paragraph.

README.md
```diff
@@ -1103,14 +1103,20 @@ column type.
...
 
-```sql
-alter table versions
-alter column object type json
-using object::json;
+```ruby
```
jaredbeck: Let's include the standard class definition for completeness, like:

```ruby
class ConvertVersionsObjectToJson < ActiveRecord::Migration
```

README.md
denispeplin: You inserted JSON into text column and converted it. But with paper_trail we actually have YAML in this column, and PostgreSQL can't guess what it is and convert it.

README.md
denispeplin: Ok, I'll do it.

README.md
denispeplin: No problem.

README.md
seanlinsley: You can still migrate from YAML to JSON by doing something like this:
```ruby
add_column :versions, :object, :new_object, :jsonb

PaperTrail::Version.find_each do |version|
  version.update_column :new_object, YAML.load(version.object)
end

remove_column :versions, :object
rename_column :versions, :new_object, :object
```

Of course it depends on how much data you have to migrate, and whether downtime of any kind is an option. If it's not, but it's okay for any UI in your app to temporarily no longer have any version object data, you can do this:
```ruby
rename_column :versions, :object, :old_object
add_column :versions, :object, :jsonb
```

That will run instantaneously, and any new version objects will use the new JSONB column. Then you can run a script in the background that moves the YAML data from the old column into the new column. Then, you can remove the old column.

@jaredbeck perhaps :arrow_double_up: should be part of the documentation. I can create a new PR after this one is merged if you'd like.

jaredbeck:
> You'll have to destroy or migrate all `versions` data before running this migration. 

As written, I think this could mislead someone who is already using `PaperTrail::Serializers::JSON`. How about something like "If you are currently using `PaperTrail::Serializers::YAML` (the default), first convert your data to JSON before running this migration."?

@seanlinsley wrote:

> You can still migrate from YAML to JSON by doing something like this:
> 
> add_column :versions, :object, :new_object, :jsonb
> 
> PaperTrail::Version.find_each do |version|
>   version.update_column :new_object, YAML.load(version.object)
> end
> 
> remove_column :versions, :object
> rename_column :versions, :new_object, :object
> 
> Of course it depends on how much data you have to migrate, and whether downtime of any kind is an option. If it's not, but it's okay for any UI in your app to temporarily no longer have any version object data, you can do this:
> 
> rename_column :versions, :object, :old_object
> add_column :versions, :object, :jsonb
> 
> That will run instantaneously, and any new version objects will use the new JSONB column. Then you can run a script in the background that moves the YAML data from the old column into the new column. Then, you can remove the old column.
> 
> @jaredbeck perhaps :arrow_double_up: should be part of the documentation. I can create a new PR after this one is merged if you'd like.

That'd be great, thanks. You don't need to wait for this PR to get merged though; I don't anticipate a conflict.

jaredbeck:
>> You'll have to destroy or migrate all versions data before running this migration.
>
> As written, I think this could mislead someone who is already using PaperTrail::Serializers::JSON. How about something like "If you are currently using PaperTrail::Serializers::YAML (the default), first convert your data to JSON before running this migration."?

Denis, what do you think of this suggestion? Thanks.

denispeplin:
I think that text like this must prepend migration in README to avoid confusion. I also think, that we may mention a possibility to destroy data.

jaredbeck:
Closing via c95086c

jaredbeck:
Thanks Denis!

PR #667, opened 2015-12-09
Use Active Record's type system from 4.2 onwards.

kaspth:
Rails 4.2 deprecates `serialized_attributes` without replacement. However,
it also introduces a type system which lets us treat all attributes the same.

Rails 4.2 has `type_for_attribute` which knows how to serialize and deserialize
itself from a database through `type_cast_for_database` and `type_cast_from_database`.

(In Rails 5 they will be `serialize` and `deserialize` respectively.)

Thus we no longer need the `PaperTrail.config.serialized_attributes` toggle,
and this change makes it do nothing. It's still kept around for backwards compatibility.

I haven't gone whole hog on refactoring the Paper Trail serialization model methods, instead I've defined them conditionally.

I would also like to request a backport of this to 4-0-stable or 4-1-stable.

### DIFF
--- a/lib/paper_trail/attributes_serialization.rb
+++ b/lib/paper_trail/attributes_serialization.rb
@@ -1,0 +1,89 @@
+module PaperTrail
+  module AttributesSerialization
+    class NoOpAttribute
+      def type_cast_for_database(value)
+        value
+      end
+
+      def type_cast_from_database(data)
+        data
+      end
+    end
+    NO_OP_ATTRIBUTE = NoOpAttribute.new
+
+    class SerializedAttribute
+      def initialize(coder)
+        @coder = coder.respond_to?(:dump) ? coder : PaperTrail.serializer
+      end
+
+      def type_cast_for_database(value)
+        @coder.dump(value)
+      end
+
+      def type_cast_from_database(data)
+        @coder.load(data)
+      end
+    end
+
+    SERIALIZE, DESERIALIZE =
+      if ::ActiveRecord::VERSION::MAJOR >= 5
+        [:serialize, :deserialize]
+      else
+        [:type_cast_for_database, :type_cast_from_database]
+      end
+
+    if ::ActiveRecord::VERSION::STRING < '4.2'
+      # Backport Rails 4.2 and later's `type_for_attribute` to build
+      # on a common interface.
+      def type_for_attribute(attr_name)
+        serialized_attribute_types[attr_name.to_s] || NO_OP_ATTRIBUTE
+      end
+
+      def serialized_attribute_types
+        @attribute_types ||= Hash[serialized_attributes.map do |attr_name, coder|
+          [attr_name, SerializedAttribute.new(coder)]
+        end]
+      end
+      private :serialized_attribute_types
+    end
+
+    # Used for `Version#object` attribute.
+    def serialize_attributes_for_paper_trail!(attributes)
+      alter_attributes_for_paper_trail!(SERIALIZE, attributes)
+    end
+
+    def unserialize_attributes_for_paper_trail!(attributes)
+      alter_attributes_for_paper_trail!(DESERIALIZE, attributes)
+    end
+
+    def alter_attributes_for_paper_trail!(serializer, attributes)
+      # Don't serialize before values before inserting into columns of type
+      # `JSON` on `PostgreSQL` databases.
+      return attributes if paper_trail_version_class.object_col_is_json?
+
+      attributes.each do |key, value|
+        attributes[key] = type_for_attribute(key).send(serializer, value)
+      end
+    end
+
+    # Used for Version#object_changes attribute.
+    def serialize_attribute_changes_for_paper_trail!(changes)
+      alter_attribute_changes_for_paper_trail!(SERIALIZE, changes)
+    end
+
+    def unserialize_attribute_changes_for_paper_trail!(changes)
+      alter_attribute_changes_for_paper_trail!(DESERIALIZE, changes)
+    end
+
+    def alter_attribute_changes_for_paper_trail!(serializer, changes)
+      # Don't serialize before values before inserting into columns of type
+      # `JSON` on `PostgreSQL` databases.
+      return changes if paper_trail_version_class.object_changes_col_is_json?
+
+      changes.clone.each do |key, change|
+        type = type_for_attribute(key)
+        changes[key] = Array(change).map { |value| type.send(serializer, value) }
+      end
+    end
+  end
+end
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -5,30 +5,25 @@ module PaperTrail
   class Config
     include Singleton
     attr_accessor :timestamp_field, :serializer, :version_limit
-    attr_reader :serialized_attributes
     attr_writer :track_associations
 
     def initialize
       @timestamp_field = :created_at
       @serializer      = PaperTrail::Serializers::YAML
-
-      # This setting only defaults to false on AR 4.2+, because that's when
-      # it was deprecated. We want it to function with older versions of
-      # ActiveRecord by default.
-      if ::ActiveRecord::VERSION::STRING < '4.2'
-        @serialized_attributes = true
-      end
     end
 
-    def serialized_attributes=(value)
-      if ::ActiveRecord::VERSION::MAJOR >= 5
-        ::ActiveSupport::Deprecation.warn(
-          "ActiveRecord 5.0 deprecated `serialized_attributes` "  +
-          "without replacement, so this PaperTrail config setting does " +
-          "nothing with this version, and is always turned off"
-        )
-      end
-      @serialized_attributes = value
+    def serialized_attributes
+      ActiveSupport::Deprecation.warn(
+        "PaperTrail.config.serialized_attributes is deprecated without " \
+        "replacement and no longer has any effect."
+      )
+    end
+
+    def serialized_attributes=(_)
+      ActiveSupport::Deprecation.warn(
+        "PaperTrail.config.serialized_attributes= is deprecated without " \
+        "replacement and no longer has any effect."
+      )
     end
 
     def track_associations
--- a/test/unit/model_test.rb
+++ b/test/unit/model_test.rb
@@ -228,7 +228,7 @@ class HasPaperTrailModelTest < ActiveSupport::TestCase
           'id'         => [nil, @widget.id]
         }
 
-        assert_equal Time, @widget.versions.last.changeset['updated_at'][1].class
+        assert_kind_of Time, @widget.versions.last.changeset['updated_at'][1]
         assert_equal changes, @widget.versions.last.changeset
       end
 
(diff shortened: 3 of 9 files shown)


test/unit/model_test.rb
```diff
@@ -982,12 +982,9 @@ def without(&block)
   if ::ActiveRecord::VERSION::MAJOR < 5
```
kaspth: Kept this around so you could see I hadn't changed much in the tests. I'll amend it after a review.

kaspth:
Ah, I originally wrote this for 4-0-stable. So I see a test has somehow changed. Looking into it :+1:

test/unit/model_test.rb
kaspth: Opted to just remove it anyway, `?w=1` works fine :grin:

kaspth:
Mostly green!

However, it seems the `created_at` and `updated_at` timestamps have a different format on MySQL with the default Gemfile. Not sure why that is, but the failures seem consistent across Ruby versions.

jaredbeck:
Interesting! This will require some careful review but it does look interesting.  I especially like how many conditionals it would remove, like this one:

```diff
-        if PaperTrail.serialized_attributes?
-          item_type.constantize.unserialize_attribute_changes_for_paper_trail!(changes)
-        end
+        item_type.constantize.unserialize_attribute_changes_for_paper_trail!(changes)
```

> However, it seems the created_at and updated_at timestamps have a different format on MySQL with the default Gemfile.

Minitest's `assert_equal` compares with `==`. Looking at the test output, the actual timestamps seem to be the same, but according to minitest, `==` returned false.  As a debugging step, maybe you could add more specific assertions, like:

```diff
  assert_kind_of Time, @widget.versions.last.changeset['updated_at'][1]
+  assert_equal changes["updated_at"].class, @widget.versions.last.updated_at.class
+  assert_equal changes["updated_at"], @widget.versions.last.updated_at
  assert_equal changes, @widget.versions.last.changeset
end
```

batter:
Thanks for the PR, this would be a great way to accommodate a good spectrum of ActiveRecord versions while maintaining serialization and deserialization functionality across the board.

After looking at this more closely, I actually feel like this really seems like a prime candidate for refactoring this logic out into it's own module(s) that can be mixed in across the board. Happy to take a crack at it at some point soon if that would be helpful.

kaspth:
RuboCop thinks me a sinner, eh? Oh well, repenting now :grin:

kaspth:
Thanks for the suggestions!

I went with @batter's suggestion and extracted the code to an `AttributesSerialization` module that we extend models with in `has_paper_trail`.

I've also made the code forward compatible with Rails 5, made the code shareable across all Rails versions and reduced duplication between attributes and change serialization and deserialization.

Now, MySQL is still bugging me. The tests pass if calling `to_i` on the timestamps, which suggests that microseconds are messing up the tests. @jaredbeck, should we just go around and special case the tests with `to_i`? I'm drawing blanks here. 😁

jaredbeck:
> .. which suggests that milliseconds are messing up the tests ..

I ran into that a few weeks ago.  IIRC MySQL does not use what it calls "fractional seconds precision" unless you tell it to.  I made a note about it in the [migration][1].

> MySQL timestamp columns do not support fractional seconds unless
> defined with "fractional seconds precision". MySQL users should manually
> add fractional seconds precision to this migration, specifically, to
> the `created_at` column.
> (https://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html)
>
> MySQL users should also upgrade to rails 4.2, which is the first
> version of ActiveRecord with support for fractional seconds in MySQL.
> (https://github.com/rails/rails/pull/14359)

Does that help at all?

[1]: https://github.com/airblade/paper_trail/blob/master/lib/generators/paper_trail/templates/create_versions.rb

kaspth:
@jaredbeck it did! If changed the migrations so: https://github.com/kaspth/paper_trail/commit/c2e6c83550a72db201cc043a7a68243ca4e93341

Tests passed locally on my MySQL 5.6, however, it seems Travis isn't as excited: https://travis-ci.org/airblade/paper_trail/builds/97962783. On top of that, it seems the `limit` option bleeds into Postgres and causes the failures there instead (I opted for that instead of the 4.2 specific `precision` option as I figured we would need to special case the migrations per db, seems we're ending up there anyway).

I'm not sure what to do if Travis doesn't have the right MySQL version. We could try removing the precision support from Rails, but that seems hacky. Any ideas? :grin:

jaredbeck:
> Tests passed locally on my MySQL 5.6, however, it seems Travis isn't as excited ..

Yes, Travis still uses MySQL 5.5 and [doesn't seem interested in supporting 5.6][1].

> .. should we just go around and special case the tests with to_i? 

You mean, relax the timestamp comparison, just for MySQL, so that it's comparing whole seconds? That's fine with me.

[1]: https://github.com/travis-ci/travis-ci/issues/1986

kaspth:
@jaredbeck nice, hopefully I'll get around to the special casing tonight. Do we have a quick way to check the adapter we're running on?

jaredbeck:
> Do we have a quick way to check the adapter we're running on?

Programmatically? There's a `using_mysql?` method defined by the `test_helper`.

kaspth:
@jaredbeck yup, that did it. Thanks! Everything is green now :grin:

lib/paper_trail/config.rb
```diff
@@ -5,30 +5,12 @@ module PaperTrail
   class Config
     include Singleton
     attr_accessor :timestamp_field, :serializer, :version_limit
-    attr_reader :serialized_attributes
+    attr_accessor :serialized_attributes # Kept around for backwardscompatibility
```
jaredbeck: Let's add a deprecation warning.

```diff
- attr_reader :serialized_attributes
+ def serialized_attributes
+   ActiveSupport::Deprecation.warn(
+     "PaperTrail.serialized_attributes is deprecated without replacement and no longer has any effect"
+   )
+ end
```

jaredbeck: The mutator method, `#serialized_attributes=`, should also have a deprecation warning.

test/unit/model_test.rb
```diff
@@ -978,100 +978,94 @@ def without(&block)
...
-        # Test for serialization:
+        # Tests for serialization:
+        # Before the serialized attributes fix, the object/object_changes value that was stored was ridiculously long (58723).
+        should 'version.object should not have stored the default, ridiculously long (to_yaml) serialization of the TimeZone object' do
+          assert @person.versions.last.object.length < 105, "object length was #{@person.versions.last.object.length}"
+        end
+        # Need an additional clause to detect what version of ActiveRecord is being used for this test because AR4 injects the `updated_at` column into the changeset for updates to models
         should 'version.object_changes should not have stored the default, ridiculously long (to_yaml) serialization of the TimeZone object' do
-          assert @person.versions.last.object_changes.length < 105, "object_changes length was #{@person.versions.last.object_changes.length}"
+          assert @person.versions.last.object_changes.length < (ActiveRecord::VERSION::MAJOR < 4 ? 105 : 118), "object_changes length was #{@person.versions.last.object_changes.length}"
```
jaredbeck: Please extract a local variable instead of repeating expression (`@person.versions.last.object_changes`) twice.

lib/paper_trail/attributes_serialization.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    SERIALIZE, DESERIALIZE =
+      if ::ActiveRecord::VERSION::MAJOR >= 5
+        [:serialize, :deserialize]
+      else
+        [:type_cast_for_database, :type_cast_from_database]
+      end
+
+    if ::ActiveRecord::VERSION::STRING < '4.2'
```
jaredbeck: Rails >= 4.2 defines `attribute_types` and `type_for_attribute` for us?

jaredbeck:
Except for a few nitpicks, everything looks good to me.  Ben, do you want to review this?

test/unit/model_test.rb
jaredbeck: Or, you might like to use `assert_operator`, e.g.

```ruby
assert_operator(
  @person.versions.last.object_changes.length, 
  :<, 
  (ActiveRecord::VERSION::MAJOR < 4 ? 105 : 118)
)
```

lib/paper_trail/attributes_serialization.rb
```diff
@@ -0,0 +1,86 @@
...
+    class NoOpAttribute
+      def type_cast_for_database(value)
+        value
+      end
+
+      def type_cast_from_database(data)
+        data
+      end
+    end
+    NO_OP_ATTRIBUTE = NoOpAttribute.new
```
sgrif: This can probably just be `ActiveRecord::Type::Value.new`

lib/paper_trail/attributes_serialization.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    class SerializedAttribute
+      def initialize(coder)
+        @coder = coder.respond_to?(:dump) ? coder : PaperTrail.serializer
+      end
+
+      def type_cast_for_database(value)
```
sgrif: FYI these method names changed in 5.0

lib/paper_trail/attributes_serialization.rb
sgrif: Never mind, I see how this is being used

lib/paper_trail/attributes_serialization.rb
sgrif: Never mind I see how this is being used

sgrif:
Unfamiliar with the codebase, so just want to make sure I'm clear -- This ends up applying to all attributes, not just serialized? If so :+1: this looks good

test/unit/model_test.rb
kaspth: You sure, you want me to make this change? It looks the same on master: https://github.com/airblade/paper_trail/blob/94db8436b58b476963110e99926588b621fe72a0/test/unit/model_test.rb#L1041.

All I've done is outdent by two spaces, because I deleted the surrounding Active Record version check. Try appending `?w=1` to the files URL, then GitHub won't show whitespace changes. :grin:

lib/paper_trail/attributes_serialization.rb
kaspth: Only `type_for_attribute`. I can see that `attribute_types` was too general a name, I'll try making it more clear.

lib/paper_trail/config.rb
batter: Agreed

kaspth:
@sgrif Thanks for the lookover! On Rails 4.2 and up, yes. Since Paper Trail handled serialized attributes explicitly on Rails 4.1, I opted to keep that support scoped to those attributes.

lib/paper_trail/attributes_serialization.rb
kaspth: Ya, it's a bit gnarly, but works :grin:

test/unit/model_test.rb
jaredbeck: > You sure, you want me to make this change? It looks the same on master.

Nevermind, I thought it was new.

> Try appending ?w=1 to the files URL, then GitHub won't show whitespace changes.

Wow, that is incredibly useful, thanks!

jaredbeck:
Merging, thanks Kasper! This is a nice simplification for PaperTrail.  Also, thanks for the review Sean!

> I would also like to request a backport of this to 4-0-stable or 4-1-stable.

I'd be happy to backport to 4.1, but I'm not sure it belongs in 4.0 because we are deprecating part of our public API.

lib/paper_trail/attributes_serialization.rb
```diff
@@ -0,0 +1,89 @@
...
+    def serialize_attributes_for_paper_trail!(attributes)
+      alter_attributes_for_paper_trail!(SERIALIZE, attributes)
+    end
+
+    def unserialize_attributes_for_paper_trail!(attributes)
+      alter_attributes_for_paper_trail!(DESERIALIZE, attributes)
+    end
+
+    def alter_attributes_for_paper_trail!(serializer, attributes)
+      # Don't serialize before values before inserting into columns of type
```
batter: Typo: duplicate 'before'

batter:
@kaspth - This looks great, thanks for the PR.

@jaredbeck - I just looked through and the only things I noticed that might be outstanding was a typo in the comments, and also that the [`PaperTrail` base module still has the `serialized_attributes?` accessor](https://github.com/airblade/paper_trail/blob/43e10393326d28d86667a7c5a61c93bd080ce0d5/lib/paper_trail.rb#L30).  I don't know if it makes sense to rip it out entirely, since like the config option, someone may be accessing it in a legacy app. Maybe it makes sense to always have it return `false` AND have a deprecation warning?  Since it's a questionized method it should return a boolean value though IMO.

batter:
Upon closer inspection it should probably be changed in some manner, since the call to the config will currently always return `true`, since it is double-banging the return from `ActiveSupport::Deprecation.warn`, which will always return the string it is passed.

lib/paper_trail/attributes_serialization.rb
kaspth: Ah, didn't spot that. Just stole it from here: https://github.com/airblade/paper_trail/blob/3a72b7f7c747ac22bb29c03f3d02c91bd070a341/lib/paper_trail/has_paper_trail.rb#L199-L200

kaspth:
@jaredbeck it's true that it breaks those config methods, on the other hand if it landed in 4.0.2 people on Rails 4.2 would never have to worry about the missing `serialized_attributes` support breaking their applications. It makes the PaperTrail 4 upgrade a lot smoother. Otherwise, I could see people being forced to go to all the way to 4.1, once that ships.

Is there any estimate on when 4.1 is out? :grin:

Thanks for the merge and great feedback! :grin:

jaredbeck:
> .. the PaperTrail base module still has the serialized_attributes? accessor. I don't know if it makes sense to rip it out entirely, since like the config option, someone may be accessing it in a legacy app. Maybe it makes sense to always have it return false AND have a deprecation warning?

Sounds good to me. I'll make a separate PR.

>>> I would also like to request a backport of this to 4-0-stable or 4-1-stable.
>>
>> I'd be happy to backport to 4.1, but I'm not sure it belongs in 4.0 because we are deprecating part of our public API.
>
> .. on the other hand if it landed in 4.0.2 people on Rails 4.2 would never have to worry about the missing serialized_attributes support breaking their applications. It makes the PaperTrail 4 upgrade a lot smoother. Otherwise, I could see people being forced to go to all the way to 4.1, once that ships.

That's a good point, but I think they'll manage. Backporting from master to a 4.x branch is enough of a hassle that I'd rather not have a 4.0.2 unless we find a big problem with it.  What do you think, Ben?

> Is there any estimate on when 4.1 is out? :grin:

Here's the backport PR: https://github.com/airblade/paper_trail/pull/678 .  As soon as we can get that merged into `4.1-stable`, Ben can cut a release.

kaspth:
We're running a version of this based on `4-0-stable` in production, which wasn't that tough, but since 4.1 sounds like it's imminent no worries 👍

Issue #43, opened 2011-03-21
Object: nil for the 'create' version


tscolari:
I'm getting a nil object on every create event.
Is this the desired behavior?

This gives a problem when the item is deleted, and you want to verify the first version with reify.

stereobooster:
I have following code in migration

```ruby
    create_table :versions do |t|
      t.string   :item_type, null: false
      t.integer  :item_id,   null: false
      t.string   :event,     null: false
      t.string  :whodunnit
      # jsonb requires Postgresql 9.4
      t.jsonb    :object # originally it was text
      t.datetime :created_at
    end

    add_index :versions, [:item_type, :item_id]
    add_index  :versions, :object, using: :gin
    # special type of index for jsonb column
    execute "CREATE INDEX versions_object_company_id ON versions ((object->>'company_id'))"
```

And following model
```ruby
class Version < PaperTrail::Version
  belongs_to :user, foreign_key: :whodunnit

  # using jsonb and foreing_key
  # otherwise I would need to use something like:
  #   PaperTrail::Version.where_object(company_id: @company.id).
  #   select('versions.*, users.first_name, users.last_name').
  #   joins('JOIN users ON users.id = versions.whodunnit')
  scope :by_company_id, -> (id) { where('object @> ?', {company_id: id}.to_json) }
  scope :includes_user, -> { includes(:user) }
end
```

I want to list all changes done for some company. Im doing it that way to fix `N+1` queries problem. But because for create event it stores `nil` object it does not show create events in the log for company. Any suggestions on how to handle it?

jaredbeck:
> I want to  ... Any suggestions on how to handle it?

Per the [contributing guide](https://github.com/airblade/paper_trail/blob/master/CONTRIBUTING.md) please ask usage questions on Stack Overflow.

emerak:
was this feature removed from master ? seems like #172 fixed it but I'm still getting nil on version 6

batter:
@emerak - #172 is related to `object_changes` column getting populated for `create` events, so it's loosely related but not the same thing. A version record is supposed to represent the object "pre-change", that is why when you call `version.object` it returns `nil` where `version.event == 'create'`.

PR #670, opened 2015-12-14
[#43] Save object for create event

stereobooster:
Saving object params for create event, so when you search for event by params, like

```ruby
PaperTrail::Version.where_object(:name => name)
```

it will also return `create` event. While `reify` for `create` event still returns `nil`.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -296,6 +296,7 @@ module PaperTrail
         if paper_trail_switched_on?
           data = {
             :event     => paper_trail_event || 'create',
+            :object    => pt_recordable_object_for_create,
             :whodunnit => PaperTrail.whodunnit
           }
           if respond_to?(:updated_at)
@@ -351,6 +352,21 @@ module PaperTrail
       # @api private
       def pt_recordable_object
         object_attrs = object_attrs_for_paper_trail(attributes_before_change)
+        if self.class.paper_trail_version_class.object_col_is_json?
+          object_attrs
+        else
+          PaperTrail.serializer.dump(object_attrs)
+        end
+      end
+
+      # Returns an object which can be assigned to the `object` attribute of a
+      # nascent version record. If the `object` column is a postgres `json`
+      # column, then a hash can be used in the assignment, otherwise the column
+      # is a `text` column, and we must perform the serialization here, using
+      # `PaperTrail.serializer`.
+      # @api private
+      def pt_recordable_object_for_create
+        object_attrs = object_attrs_for_paper_trail(attributes)
         if self.class.paper_trail_version_class.object_col_is_json?
           object_attrs
         else
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -201,7 +201,7 @@ module PaperTrail
     #   - `:preserve` - Attributes undefined in version record are not modified.
     #
     def reify(options = {})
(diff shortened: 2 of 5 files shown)


jaredbeck:
Please add a description to this PR, and a line item to the changelog.  Also, your rails4/postgres tests are failing, consistently, for all versions of ruby.

stereobooster:
Yep noticed failing tests. Im in process of fixing them.

stereobooster:
It seems my changes didn't brake tests

have the same failure for ruby 2.0.0 + pgsql on master

```
  1) Failure:
InheritanceColumnTest#test_: STI models should work with custom STI inheritance column.  [/Users/slavik/my/paper_trail/test/unit/inheritance_column_test.rb:34]:
Expected: "Dog"
  Actual: "NilClass"
```

still investigating the actual reason

stereobooster:
Tests fixed. Added description for PR. ^@jaredbeck

jaredbeck:
>> Please add a description to this PR
>
> .. when you search for event [using `where_object`] it will also return create event.

Thanks! That makes sense to me, but I don't know why create events were omitted in the first place. Maybe Ben can address that. @batter 

On a technical note, you've changed a number of files which are unrelated to the actual purpose here, specifically introducing vagrant, and editing the YAML database configuration (e.g. `test/dummy/config/database.mysql.yml`).  I'm certain we don't want to introduce vagrant and I'm fairly certain we're not interested in changing our database configuration either. Perhaps you'd like to explain the latter, but I'm certain that we don't want the former.

batter:
Regarding why 'create' events reify to `nil`:

The idea was that since versions are supposed to represent what the version looks like in the state "before" / "prior" to the version being generated, a 'create' event is generated upon version creation, and thus prior to that event occurring the object didn't exist, hence `nil` being returned. I know lots of people ask for this, and I personally when using the gem usually don't save 'create' events because they aren't all that useful IMO.

I actually like the vagrant box config, although not sure it makes sense to put in the Repo itself, perhaps as part of the wiki for windows users that want to be able to tool around on stuff. I also don't think we want to change the Postgres setup to use anything other than `localhost`, and Travis seems to work fine as is, so I don't think we need those changes. The rest of the PR seems sensible to me.

stereobooster:
First of all this "unrelated" changes located in one commit so can be easily detached.

Vagrant is not dependency, but additional testing tool. It installs and isolates all dependencies for testing env similar too travis. You can run all test suites (matrix) with [wwtd](grosser/wwtd). Think of it as one step to run tests the same way as travis do (actually it is not one step, the world is not perfect yet). 

I changed db settings according to travis wiki. And it obvious It didn't break travis. I did it because I want to have it as much canonical as possible, to be able to adjust VM according standards.

stereobooster:
Detached commit with vagrant. [Leaved it in my repo](https://github.com/airblade/paper_trail/compare/master...stereobooster:vagrant-test-env?expand=1)

jaredbeck:
> [I removed the] commit with vagrant

Thanks.

> Please add a .. line item to the changelog.

Please add a line to the changelog, under "Breaking Changes" for 5.0.

jaredbeck:
> Please add a line to the changelog, under "Breaking Changes" for 5.0.

@stereobooster Please add a changelog entry, so that we can merge this, thanks.

jaredbeck:
@stereobooster I haven't heard back from you after repeated requests to add a changelog entry, so I'm going to close this, but I'll be happy to reopen it if you can provide the requested entry.

PR #673, opened 2015-12-18
Fix #672

### LINKED ISSUE #672: Proposed: A convenient way to query versions in chronological order
Found at least one failing for this bug on Mac OS X. Believe there are more hidden

```
  1) Failure:
InheritanceColumnTest#test_: STI models should work with custom STI inheritance column.  [paper_trail/test/unit/inheritance_column_test.rb:35]:
Expected: "Dog"
  Actual: "NilClass"

308 runs, 593 assertions, 1 failures, 0 errors, 0 skips
rake aborted!
Command failed with status (1): [ruby -I"lib:lib:test" -I"/opt/boxen/rbenv/versions/2.0.0-p647/lib/ruby/gems/2.0.0/gems/rake-10.1.1/lib" "/opt/boxen/rbenv/versions/2.0.0-p647/lib/ruby/gems/2.0.0/gems/rake-10.1.1/lib/rake/rake_test_loader.rb" "test/**/*_test.rb" ]

Tasks: TOP => default => test
(See full trace by running task with --trace)
```

From [PostgreSQL documentation](http://www.postgresql.org/docs/9.1/static/sql-select.html) 

> If `ORDER BY` is not given, the rows are returned in whatever order the system finds fastest to produce.

So in my case versions array looks like

```
  >p dog_versions.map { |v| v.reify.class.name }
=>["Dog", "Dog", "Dog", "NilClass"]
```

but actually it should be `["NilClass", "Dog", "Dog", "Dog"]`

stereobooster:


stereobooster:
I will get linux box and return with fixed test

stereobooster:
Im agree with fact that [`default_scope` is evil](http://rails-bestpractices.com/posts/2013/06/15/default_scope-is-evil/). But this means that we generally should not rely on default (unordered queries) for `Version` model like in testcase mentioned in #672.

We can do it with named scope. And mention it documentation. And rewrite tests accordingly.

Other concern is to use ordering consistently. `subsequent` method orders model only using one field, while `timestamp_sort_order` will order using two fields (if two available).

jaredbeck:
In https://github.com/airblade/paper_trail/pull/680 I have fixed the intermittently failing test mentioned in https://github.com/airblade/paper_trail/issues/672.

I don't think we should add a `default_scope`, as proposed in this PR, because it would have difficult to predict side effects, so I'm going to close this, but please let me know if I'm missing something.  Thanks!

PR #675, opened 2015-12-20
Hopefully fix intermittent postgres test

jaredbeck:
Test would fail when FFaker::Color.name returned the same exact
string twice in a row. This is not FFaker's fault.

Incidentally, the FFaker color list currently has 19 entries.
(https://github.com/ffaker/ffaker/blob/master/lib/ffaker/data/color/names_list)
The test suite runs this postgres-specific test six times (3 rubies * 2 major
versions of rails). So, I think the chance of the test suite failing was
(1 / 19)^2 * 6, or 0.016, or about 2%.


batter:
Good catch!

PR #683, opened 2015-12-31
Destroy callback default changed to :before

### LINKED ISSUE #682: Rails 5: belongs_to_required_by_default prevents versioning of destroy events
When using the ActiveRecord configuration belongs_to_required_by_default (set to true for new Rails 5 projects), PaperTrail::Version tries to validate presence of the versioned item. Since versioning is done after destroy (by default), no matching item exists and the version creation silently fails.

See https://gist.github.com/owenr/60b34a2cdc13edf89412 for a test using the bug report template.

Proposed fix: explicitly make the item relationship optional.

owenr:
To allow after_destroy versioning with Rails 5; fixes #682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   that PaperTrail no longer adds the `set_paper_trail_whodunnit` before_filter
   for you. Please add this before_filter to your ApplicationController to
   continue recording whodunnit. See the readme for an example.
+- [#683](https://github.com/airblade/paper_trail/pull/683) /
+  [#682](https://github.com/airblade/paper_trail/issues/682) -
+  Destroy callback default changed to :before to accommodate ActiveRecord 5
+  option `belongs_to_required_by_default` and new Rails 5 default.
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ end
 ```
 
 The `paper_trail_on_destroy` method can be further configured to happen
-`:before` or `:after` the destroy event. By default, it will happen after.
+`:before` or `:after` the destroy event. By default, it will happen before.
 
 ## Choosing When To Save New Versions
 
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -116,9 +116,20 @@ module PaperTrail
       end
 
       # Record version before or after "destroy" event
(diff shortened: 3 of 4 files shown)


jaredbeck:
There may be a better alternative, see discussion in https://github.com/airblade/paper_trail/issues/682.

jaredbeck:
Also, if we were to proceed with this PR, it would need a test, please.

jaredbeck:
> Also, if we were to proceed with this PR, it would need a test, please.

Actually, we don't have tests set up for rails 5 yet, do we? I'll see if I can do that.

owenr:
I've tried setting up a Rails 5 env but it looks like it might be a bit premature (dependencies force us to also support Sinatra 2.0.0 which is alpha, the `rails-controller-testing` gem does not do what it says on the tin, and some deprecation warnings are unavoidable without dropping support for older versions of Ruby/Rails). You can see my progress at [GitHub](https://github.com/owenr/paper_trail/tree/travis-rails5) and [Travis](https://travis-ci.org/owenr/paper_trail/builds/99906367).

jaredbeck:
> I've tried setting up a Rails 5 env but it looks like it might be a bit premature ..

I'm actually working on this right now :) I'm trying out [appraisal][1].  If you want to email me (jared at jaredbeck dot com we can hop on skype or something).

[1]: https://github.com/thoughtbot/appraisal

jaredbeck:
This is looking good. Can you squash your commits, please?

owenr:
Looks like I've made a mess... Do you have reliable documentation on squashing (and/or dropping) commits? Otherwise I'm seriously considering [this](https://xkcd.com/1597/).

lib/paper_trail/version_concern.rb
```diff
@@ -5,7 +5,9 @@ module VersionConcern
     extend ::ActiveSupport::Concern
 
     included do
-      belongs_to :item, :polymorphic => true
+      opts = {:polymorphic => true}
+      opts[:required] = false if ActiveRecord::Base.respond_to? :belongs_to_required_by_default
```
jaredbeck: Do we need both this change (required: false) *and* the change to the default `recording_order`?

owenr: We don't - I was trying to revert that. Let me try again.

jaredbeck:
Thanks Owen!

batter:
:+1:

PR #688, opened 2016-01-11
adding ruby2.3

ferdinandrosario:


jaredbeck:
Thanks, I'd like to make this change, but every ruby we test against adds a lot of time to our test suite.  Currently, our test suite (which includes ruby 2.2) takes about 75 minutes (21 jobs).  If we test against both ruby 2.2 and 2.3, it takes 110 minutes (30 jobs).

What do you think about testing against 2.3 *instead of* 2.2?

daande:
I would like to see 2.3.

ferdinandrosario:
so do u wants to remove 2.2 completed? if you wish I will modify it

jaredbeck:
>> What do you think about testing against 2.3 instead of 2.2?

> do u wants me to remove 2.2?

I'm asking if you think that is a good idea :)

daande:
Yes it's a good idea to goto the newest. rails 5.0 required ruby to be 2.2.2 or greater anyways.

daande:
Also sorry for the late reply I was swamped with work today.

ferdinandrosario:
yeah, moving to 2.3 implies that papertrail only support  2.3 which good and bad for some people.not everyone using the latest ruby, so I suggest keeping all officially supported ruby version with latest patch. 
* 1.9.3
* 2.1.4
* 2.2.4
* 2.3.0
but as you said this will increase the CI timing. waiting for your feedback

seanlinsley:
@jaredbeck I think it's fine to only test against 1.9 and 2.3.

ferdinandrosario:
@seanlinsley +1 but most of the users are using 2.2 so keeping 2.2 will be great, we can remove 2.1..

daande:
As @ferdinandrosario said if most users are using 2.2 then lets remove 2.1 and add 2.3.  I agree with that change.

jaredbeck:
Thanks for the feedback everyone. Going forward, we'll test against 1.9.3 and 2.3.0. See https://github.com/airblade/paper_trail/pull/690

PR #689, opened 2016-01-12
Pass ar5 tests

owenr:
Update test code to address deprecations/warnings and fix one legitimate issue with AR5. Not happy with several aspects of this change:
* Gem `rails-controller-testing` does not work as advertised (possibly an issue with load order?) and I could only get it working when manually included, see `test_helper.rb`
* HTTP request test methods in Rails 5 need parameter hash keys named 'params', so a new wrapper function is introduced twice (once in test_helper, again in spec_helper)
* Addressing the AR5 issue seems to require a way of preventing #save on the has_one object, and there is almost certainly a better way than temporarily redefining #save on the instance

Having said that, the changes pass tests locally on MRI 2.3 (ar5) and 1.9.3-p551 (ar3/ar4). Comments and suggestions welcome.

@jaredbeck

### DIFF
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -154,7 +154,9 @@ module PaperTrail
               else
                 child = version.reify(options.merge(:has_many => false, :has_one => false))
                 model.appear_as_new_record do
-                  model.send "#{assoc.name}=", child
+                  without_persisting(child) do
+                    model.send "#{assoc.name}=", child
+                  end
                 end
               end
             end
@@ -264,6 +266,18 @@ module PaperTrail
           paper_trail_version_class.
           where("id IN (#{version_id_subquery})").
           inject({}) { |acc, v| acc.merge!(v.item_id => v) }
+      end
+
+      # Temporarily suppress #save so we can reassociate with the reified
+      # master of a has_one relationship. Since ActiveRecord 5 the related
+      # object is saved when it is assigned to the association. ActiveRecord
+      # 5 also happens to be the first version that provides #suppress.
+      def without_persisting(record)
+        if record.class.respond_to? :suppress
+          record.class.suppress { yield }
+        else
+          yield
+        end
       end
     end
   end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -102,3 +102,14 @@ RSpec.configure do |config|
   Kernel.srand config.seed
 =end
 end
+
+# Wrap args in a hash to support the ActionController::TestCase and
+# ActionDispatch::Integration HTTP request method switch to keyword args
+# (see https://github.com/rails/rails/blob/master/actionpack/CHANGELOG.md)
+def params_wrapper(args)
+  if defined?(::Rails) && Gem::Version.new(ActiveRecord::VERSION::STRING) >= Gem::Version.new('5.0.0.beta1')
+    { params: args }
+  else
+    args
+  end
+end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -21,6 +21,11 @@ require 'shoulda'
 require 'ffaker'
 require 'database_cleaner'
 
+if Gem::Version.new(ActiveRecord::VERSION::STRING) >= Gem::Version.new('5.0.0.beta1')
+  # See https://github.com/rails/rails-controller-testing/issues/5
+  ActionController::TestCase.send(:include, Rails::Controller::Testing::TestProcess)
+end
+
 Rails.backtrace_cleaner.remove_silencers!
 
 # Run any available migration
@@ -35,7 +40,11 @@ DatabaseCleaner.strategy = :truncation
 # global setup block resetting Thread.current
 class ActiveSupport::TestCase
   if using_mysql?
-    self.use_transactional_fixtures = false
+    if respond_to? :use_transactional_tests=
+      self.use_transactional_tests = false
+    else
+      self.use_transactional_fixtures = false
+    end
     setup { DatabaseCleaner.start }
   end
 
@@ -89,6 +98,22 @@ def change_schema
   reset_version_class_column_info!
 end
 
+# Wrap args in a hash to support the ActionController::TestCase and
+# ActionDispatch::Integration HTTP request method switch to keyword args
+# (see https://github.com/rails/rails/blob/master/actionpack/CHANGELOG.md)
+def params_wrapper(args)
+  if defined?(::Rails) && Gem::Version.new(ActiveRecord::VERSION::STRING) >= Gem::Version.new('5.0.0.beta1')
+    { params: args }
+  else
+    args
+  end
+end
+
+def reset_version_class_column_info!
+  PaperTrail::Version.connection.schema_cache.clear!
+  PaperTrail::Version.reset_column_information
+end
+
 def restore_schema
   ActiveRecord::Migration.verbose = false
   ActiveRecord::Schema.define do
@@ -97,9 +122,4 @@ def restore_schema
   end
   ActiveRecord::Migration.verbose = true
   reset_version_class_column_info!
-end
-
-def reset_version_class_column_info!
-  PaperTrail::Version.connection.schema_cache.clear!
-  PaperTrail::Version.reset_column_information
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,6 @@ gemfile:
 
 matrix:
   fast_finish: true
-  allow_failures:
-    - gemfile: gemfiles/ar5.gemfile
   exclude:
     - gemfile: gemfiles/ar5.gemfile
       rvm: 1.9.3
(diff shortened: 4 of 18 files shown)


owenr:
Having some issues with tests intermittently failing. Any ideas on that are also welcome.

bronson:
Interesting...  Ten tests failed the first time I ran `rake`, but that was it.  Now I can't get any to fail.

Here's the log: https://gist.github.com/bronson/70dd3c4aaf395df4bf93

And, if I `rm -r test/dummy/db/*.sqlite3` and run `rake` again, I get the same ten errors.  Looks to be an issue related to a fresh database?

Is that the problem you're seeing?

bronson:
I'm at 25 iterations, no failures yet.  I'll let it run for a few hundred.
```bash
ctr=; while rake; do echo ITERATION $((ctr += 1)); done
```
But, no, I'm not seeing any intermittent errors on your branch.

I'm using sqlite, are you testing on a different database?

jaredbeck:
Owen, thanks for this! I'll review what I can this week.

>  .. if I rm -r test/dummy/db/*.sqlite3 and run rake again, I get the same ten errors. Looks to be an issue related to a fresh database?

Scott, Yeah, that's a known issue with running the local test suite. If you can fix it, that'd be great.

bronson:
260 iterations, 0 failures.  I don't see any intermittent issues.  (sqlite, mac, ruby 2.2.2)

owenr:
Initially tested using sqlite3 locally, no issues. As you can see, Travis CI failed all my builds for various reasons; it looks like Postgresql is the main issue now. Simulating locally it took me 22 runs before this failure (postgres, mac, 2.3.0):

```
PaperTrailCleanerTest#test_: `clean_versions!` method No options provided should removes the earliest version(s).  [/Users/owen/dev/paper_trail/test/unit/cleaner_test.rb:34]:
--- expected
+++ actual
@@ -1 +1 @@
-["Aglae Schaden", "Carley Doyle", "Karianne McDermott"]
+["Aglae Schaden", "Carley Doyle", "Jaime Witting"]
```

bronson:
Switched to postgres and I finally got a failure after 76 iterations: https://gist.github.com/bronson/01f05504f81efe072b40

Glancing at it, maybe it died because name contained a single quote... A sql injection opportunity?  :)  I'll look closer tomorrow if nobody beats me to it.

Also, there might be a test order dependency.  At least:

`rspec --seed=0`: 0 errors
`rspec --seed=1`: 1 error
`rspec --seed=9`: 10 errors

Do the tests have a required order?

owenr:
Confirmed, unescaped input (only when using postgres). I've written some failing tests and will open an issue. Unfortunately it seems unrelated to the failures from Travis [here](https://travis-ci.org/airblade/paper_trail/jobs/102019148#L355) and [here](https://travis-ci.org/airblade/paper_trail/jobs/101750643#L355).

I'm not familiar with the impact of seeding on this test suite and I couldn't reproduce those spec failures with seeds 1 and 10 (mac, postgres, ruby 2.2.2).

jaredbeck:
When comparing gem versions, I like to use `Gem::Version`, e.g.

```diff
- ActiveRecord::VERSION::STRING >= '5.'
+ Gem::Version.new(ActiveRecord::VERSION::STRING) >= Gem::Version.new("5")
```

PS: If we didn't have to support AR3, we could use the more convenient `ActiveRecord.gem_version`

lib/paper_trail/reifier.rb
```diff
@@ -162,6 +164,15 @@ def reify_has_ones(transaction_id, model, options = {})
         end
       end
 
+      # Suppress #save so we can reassociate with the reified master of a has_one relationship
```
jaredbeck: Can you expand on this comment? Specifically, please explain what changed in AR5.

Also, please alphabetize methods.

jaredbeck:
Owen, I've been cherry-picking the less controversial commits out of here.

- https://github.com/airblade/paper_trail/pull/691
- https://github.com/airblade/paper_trail/pull/692

I thought there was no sense in waiting, but it will mean you'll have to rebase this PR, sorry.

owenr:
No problem at all; like I said, this code is not right and should be improved before merging.

Some comments:
* I've reordered methods alphabetically as requested.
* Switched to Gem::Version for comparisons, but note that comparison with '5' is insufficient ('5' is not >= '5.0.0.beta').
* I can't find the AR5 change to has_one, but from the symptoms it looks like the associated object is persisted before the main object (which seems correct as per the AR4 documentation). I've elaborated on the comment.

jaredbeck:
Owen, It looks like this needs another rebase, sorry.

test/test_helper.rb
```diff
@@ -16,6 +16,17 @@ def using_mysql?
...
+if Gem::Version.new(ActiveRecord::VERSION::STRING) >= Gem::Version.new('5.0.0.beta1')
+  ActiveSupport.on_load(:action_controller) do
+    ActionController::TestCase.send(:include, Rails::Controller::Testing::TestProcess)
+    ActionController::TestCase.send(:include, Rails::Controller::Testing::TemplateAssertions)
+
+    ActionDispatch::IntegrationTest.send(:include, Rails::Controller::Testing::TemplateAssertions)
+    ActionDispatch::IntegrationTest.send(:include, Rails::Controller::Testing::Integration)
+
+    ActionView::TestCase.send(:include, Rails::Controller::Testing::TemplateAssertions)
+  end
```
jaredbeck: This whole `on_load` could use a comment, e.g. why are we including these modules, why are we using `send`, why `on_load`, etc. I don't personally understand any of it, so the comment is mostly for me :p

test/test_helper.rb
```diff
@@ -89,6 +100,19 @@ def change_schema
   reset_version_class_column_info!
 end
 
+def params_wrapper(args)
+  if defined?(::Rails) && Gem::Version.new(ActiveRecord::VERSION::STRING) >= Gem::Version.new('5.0.0.beta1')
+    { params: args }
+  else
+    args
+  end
+end
```
jaredbeck: Please add a comment explaining why this `params_wrapper` method is necessary.

jaredbeck:
> HTTP request test methods in Rails 5 need parameter hash keys named 'params'

Is this change documented in a changelog somewhere, or a guide?

jaredbeck:
>> HTTP request test methods in Rails 5 need parameter hash keys named 'params'
> 
> Is this change documented in a changelog somewhere, or a guide?

Found it. It's in the release notes for ActionPack 5:

> Migrating to keyword arguments syntax in ActionController::TestCase and ActionDispatch::Integration HTTP request methods. (Pull Request)
> http://edgeguides.rubyonrails.org/5_0_release_notes.html#action-pack

spec/spec_helper.rb
```diff
@@ -102,3 +102,11 @@
...
 =end
 end
+
+def params_wrapper(args)
+  if defined?(::Rails) && Gem::Version.new(ActiveRecord::VERSION::STRING) >= Gem::Version.new('5.0.0.beta1')
+    { params: args }
+  else
+    args
+  end
+end
```
jaredbeck: Please add a comment explaining the change to ".. keyword arguments syntax in ActionController::TestCase and ActionDispatch::Integration HTTP request methods." and include a link to the ActionPack changelog (https://github.com/tgxworld/rails/blob/master/actionpack/CHANGELOG.md)

test/test_helper.rb
owenr: Copied wholesale from [rails-controller-testing](https://github.com/rails/rails-controller-testing/blob/master/lib/rails-controller-testing.rb). I don't know why putting it here works but requiring the library doesn't. I've also tried `require: false` in the Gemfile and manually requiring it instead but that doesn't help. Removing this code, I get:
```
NoMethodError: assigns has been extracted to a gem. To continue using it,
        add `gem 'rails-controller-testing'` to your Gemfile.
```
Not knowing if this is caused by something in PaperTrail or an issue with rails-controller-testing itself (or Rails 5), I'm not sure whether I should raise an issue over there.

owenr:
Rebased. I also fixed one more deprecation warning (use_transactional_tests) and added a note about 1-second differences in tests (for MySQL and AR5). Still not happy.

owenr:
I think this is just about done. Please review and let me know; I'll squash the commits.

jaredbeck:
This looks good to me. I will merge this unless Ben has any concerns @batter 

I wish we didn't have to time-travel just to make MySQL happy, but it is something we've run into before (https://github.com/airblade/paper_trail/issues/314, https://github.com/airblade/paper_trail/commit/1d120ab9e4cb02890970b69e8a4e6e994b5fd117)

owenr:
Time travel is required due to a change in ActiveRecord 5 (if no attributes change, nothing is done). We could `sleep` instead of travel if that's preferred.
I've squashed the commits (and backed out an unneeded change to rubocop_todo picked up in review).

jaredbeck:
Merged. Thanks for your work on this, Owen.

bronson:
Wooo!  Great job Owen.

PR #695, opened 2016-01-16
Switch to Railtie, add config.paper_trail.enabled

### LINKED ISSUE #584: Can't remove deprecation warning after updating to 4.0.0
I receive the following error in my app after upgrading to 4.0.0 (from 3.0.8)
```ruby
DEPRECATION WARNING: Currently, Active Record suppresses errors raised within `after_rollback`/`after_commit` callbacks and only print them to the logs. In the next version, these errors will no longer be suppressed. Instead, the errors will propagate normally just like in other Active Record callbacks.

You can opt into the new behavior and remove this warning by setting:

  config.active_record.raise_in_transactional_callbacks = true
```

The line that it's complaining about is `has_paper_trail` in my models. I have the config statement set in my `config.application.rb`, and was only able to remove the warnings by adding `ActiveRecord::Base.raise_in_transactional_callbacks = true` before my `Bundler.require` statement in an earlier part of my `config.application.rb`

According to https://github.com/rails/rails/issues/18084#issuecomment-67503032, it might be an issue somewhere in `paper_trail`. Any thoughts/suggestions?

### LINKED ISSUE #635: PaperTrail always enabled in controllers.
I'm using Puma as my web server, and have in `application.rb`:

```
    config.after_initialize do
      #run `spring stop` before change takes effect.
      PaperTrail.enabled = false
      # do this again just to be sure I've set it
      PaperTrail.paper_trail_store[:paper_trail_enabled] = false
    end
```

I'm able to login to `rails c` and see: `PaperTrail.enabled?  #=> false`

However, when I'm posting a request to one of my endpoint with `binding.pry`, and when I do `PaperTrail.enabled? #=> true` I've created an app from scratch and it's running webrick and still has the same issue. Is this a bug or am I missing something?

Hermanverschooten:
This PR does 2 things.

* It switches to using a railtie to include paper_trail, fixing the deprecation warning with Rails 4.2.5,closes #584
* It allows you to disable of paper_trail from the environment configuration, closes #635


jaredbeck:
> This PR does 2 things.

Would it be awkward to separate this into two PRs? Thanks.

batter:
Also, the `Railtie` class should be housed within [`lib/paper_trail/frameworks/rails`](https://github.com/airblade/paper_trail/tree/master/lib/paper_trail/frameworks/rails)

batter:
And it should be loaded in by the [`PaperTrail::Rails` module](https://github.com/airblade/paper_trail/blob/87751ac882b52a1df0bb4e7bad9aeb15d367dd5b/lib/paper_trail/frameworks/rails.rb) most likely

Hermanverschooten:
I am not familiar with the codebase, I just replaced it in the spot where active record was loaded before.
But I'll have a got at it.

Hermanverschooten:
Ok, I moved the code to their respective frameworks.
About splitting the PR, I don't know where/how to access the config variables outside of the RailTie.
So the second PR would still be dependent on this one.

lib/paper_trail/config.rb
```diff
@@ -34,9 +35,13 @@ def track_associations
     end
     alias_method :track_associations?, :track_associations
 
+    def disable!
+      @disabled = true
+    end
```
jaredbeck: This method seems unnecessary.

lib/paper_trail/frameworks/rails/railtie.rb
```diff
@@ -0,0 +1,14 @@
+module PaperTrail
+  module Rails
+    class Railtie < ::Rails::Railtie
+      config.paper_trail = ActiveSupport::OrderedOptions.new
+      initializer 'paper_trail.initialisation' do |app|
+        ActiveRecord::Base.send :include, PaperTrail::Model
+
+        unless app.config.paper_trail.fetch(:enabled, true)
+          PaperTrail.config.disable!
```
jaredbeck: Why not use the existing `PaperTrail.enabled=`?

test/paper_trail_test.rb
```diff
@@ -16,6 +16,19 @@ class PaperTrailTest < ActiveSupport::TestCase
...
   end
 
+  context 'can be disabled' do
+    setup {PaperTrail.config.disable!}
+
+    should 'be disabled' do
+      Thread.new do
+        assert !PaperTrail.enabled?
+        PaperTrail.enabled = true
+        assert PaperTrail.enabled?
```
jaredbeck: Given that the return value is actually boolean (not merely truthy) please use `assert_equal`, not `assert`. (`assert` tests truthiness)

jaredbeck:
There is another problem with the proposed `@disabled` ivar, that I didn't see before. `PaperTrail::Config` is a singleton, so its instance variables are not thread-safe. That's why we started using `RequestStore`.

As I stated in https://github.com/airblade/paper_trail/issues/635, we have already decided that the flag which enables PT should be thread-safe. It would be very confusing to have two flags controlling whether PT were enabled, and even more confusing if one is thread-safe and the other (`@disabled`) is not.

I'm going to continue exploring this PR, because I want to learn more about railties in hopes that will fix https://github.com/airblade/paper_trail/issues/635. As it stands, we can't merge this, but let me see what I can come up with.

jaredbeck:
Unfortunately, simply removing the non-thread-safe `@disabled` ivar and keeping the rest of the changes re: railties fails to fix https://github.com/airblade/paper_trail/issues/635. Herman, do you have any other ideas?

jaredbeck:
In https://github.com/airblade/paper_trail/issues/635, we are discussing making the main flag "global" again (ie. moving it out of the RequestStore)

jaredbeck:
Closing because this PR is continued by https://github.com/airblade/paper_trail/pull/717, which keeps @Hermanverschooten's idea and authorship (thank you!) but renames certain methods to clarify which are thread-safe and which are not.

PR #697, opened 2016-01-18
Bind parameters in where_object, where_object_changes

### LINKED ISSUE #696: Operations on Postgresql json/jsonb columns inadequately sanitised
If postgresql is configured with json or jsonb columns, searches are not properly parameterised or escaped. See [bug report template gist](https://gist.github.com/owenr/b6490f3ac0e683dd7f65).
Credit to @bronson for finding the issue.

jaredbeck:
Fixes #696 

I don't have a lot of experience using postgres' json and jsonb through activerecord, but it looks like one can use placeholders, e.g.

```ruby
User.where('preferences @> ?', {newsletter: true}.to_json)
```

So, I tried to replace string interpolation with placeholders, and it seems to fix #696

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,6 +17,11 @@ AllCops:
 Style/AlignParameters:
   EnforcedStyle: with_fixed_indentation
 
+# Please use semantic style, e.g. `do` when there's a side-effect, else `{}`.
+# The semantic style is too nuanced to lint, so the cop is disabled.
+Style/BlockDelimiters:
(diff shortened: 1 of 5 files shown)


jaredbeck:
Note to self: I found ["Using PostgreSQL and jsonb with Ruby on Rails"][1] by Nando Vieira helpful while working on this.

[1]: http://nandovieira.com/using-postgresql-and-jsonb-with-ruby-on-rails

Issue #699, opened 2016-01-18
Intermittent failure when running Postgres tests: undefined method 'name' in clean_versions!


bronson:
Got this in two iterations while running `DB=postgres rake` tests.

(re-running, this appears to be a lot more rare than it would originally seem... I'm now at 50 iterations without having hit this again.  First time luck.)

```
PaperTrailCleanerTest#test_: `clean_versions!` method No options provided should removes the earliest version(s). :
NoMethodError: undefined method `name' for nil:NilClass
    /Users/bronson/paper_trail/test/unit/cleaner_test.rb:34:in `block (4 levels) in <class:PaperTrailCleanerTest>'
    /Users/bronson/paper_trail/test/unit/cleaner_test.rb:34:in `map'
    /Users/bronson/paper_trail/test/unit/cleaner_test.rb:34:in `block (3 levels) in <class:PaperTrailCleanerTest>'
    shoulda-context (1.2.1) lib/shoulda/context/context.rb:413:in `instance_exec'
    shoulda-context (1.2.1) lib/shoulda/context/context.rb:413:in `block in create_test_from_should_hash'
    minitest (5.8.3) lib/minitest/test.rb:108:in `block (3 levels) in run'
    minitest (5.8.3) lib/minitest/test.rb:205:in `capture_exceptions'
    minitest (5.8.3) lib/minitest/test.rb:105:in `block (2 levels) in run'
    minitest (5.8.3) lib/minitest/test.rb:256:in `time_it'
    minitest (5.8.3) lib/minitest/test.rb:104:in `block in run'
    minitest (5.8.3) lib/minitest.rb:334:in `on_signal'
    minitest (5.8.3) lib/minitest/test.rb:276:in `with_info_handler'
    minitest (5.8.3) lib/minitest/test.rb:103:in `run'
    minitest (5.8.3) lib/minitest.rb:781:in `run_one_method'
    minitest (5.8.3) lib/minitest.rb:308:in `run_one_method'
    minitest (5.8.3) lib/minitest.rb:296:in `block (2 levels) in run'
    minitest (5.8.3) lib/minitest.rb:295:in `each'
    minitest (5.8.3) lib/minitest.rb:295:in `block in run'
    minitest (5.8.3) lib/minitest.rb:334:in `on_signal'
    minitest (5.8.3) lib/minitest.rb:321:in `with_info_handler'
    minitest (5.8.3) lib/minitest.rb:294:in `run'
[...]

PR #701, opened 2016-01-19
Default order by timestamp to ensure early versions are cleaned

owenr:
Addresses #699: 'cleaner' intermittently fails to keep the latest version because the query assumes but does not enforce chronological order. Postgresql exposes the assumption by failing the tests but not 100% of the time.

I had considered ordering by `id` instead of `timestamp` but was unsure about the reliability of that approach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,10 @@ None
 
 ### Fixed
 
-None
+- [#701](https://github.com/airblade/paper_trail/pull/701) /
+  [#699](https://github.com/airblade/paper_trail/issues/699) -
+  Cleaning old versions explicitly preserves the most recent
+  versions instead of relying on database result ordering.
 
 ## 4.1.0 (Unreleased)
 
--- a/lib/paper_trail/cleaner.rb
+++ b/lib/paper_trail/cleaner.rb
@@ -6,7 +6,8 @@ module PaperTrail
     # Options:
     #
     # - :keeping - An `integer` indicating the number of versions to be kept for
-    #   each item per date. Defaults to `1`.
+    #   each item per date. Defaults to `1`. The most recent matching versions
+    #   are kept.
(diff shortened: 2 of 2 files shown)


bronson:
Very nice.  Personally I find ordering by ID more reliable since timestamps can end up equal.

owenr:
How about `order(PaperTrail.timestamp_field, :id)` ?

bronson:
Sure, that fixes the equality problem.  Does it have a benefit over sorting by id only?

owenr:
If someone changed the assignment of IDs in their database I could see it being a problem, but having checked it looks like the table will always be created with monotonically-increasing IDs. @jaredbeck, any preference?

jaredbeck:
The documentation for the `:keeping` says:

> - :keeping - An `integer` indicating the number of versions to be kept for
> each item per date. Defaults to `1`.

Can you add something about how the one that's kept is the most recent?

For the sort order, I'd suggest `order(PaperTrail.timestamp_field, :id)` because it just makes sense to use timestamps for a chronological ordering. Scott makes a good point about timestamp equality, and adding the id as a secondary sort should give us deterministic results in that case.

Finally, please add a line to the changelog, under 5.0.0 -> Fixed. Thanks!

owenr:
Done.

jaredbeck:
Thanks!

batter:
I believe we need should update this to use [`PaperTrail::Version.timestamp_sort_order`](https://github.com/airblade/paper_trail/blob/cfcd11d7a5ecdd7882a1ab19d00316e18bd9ea92/lib/paper_trail/version_concern.rb#L101). Apologies for the delayed response, just saw this, I'm happy to work that in tomorrow when I get a chance.

batter:
Made the aforementioned update: c2007210050a93223110a2b4fb4cf3a099c2755d

jaredbeck:
Thanks Ben. Good catch.

PR #702, opened 2016-01-20
Fix link to Custom Serializer section

brandonhilkert:
Looks like the heading changed at some point, not allowing one to easily click down to that section


jaredbeck:
Thanks Brandon!

In the future, when your change is documentation-only, you can add `[ci skip]` to your commit message to avoid running the test suite.

PR #704, opened 2016-01-23
Remove duplicate link to RailsCast

stgeneral:
The change removes index [3] as well. Let me know if we need to reorder other indexes and I'll update the PR.


jaredbeck:
Thanks.

Issue #707, opened 2016-02-04
rake aborted! ActiveRecord::StatementInvalid: Mysql2::Error: Table 'my_database_development.version_associations' doesn't exist:


andreydeineko:
I have encountered the problem with papertrail gem. Ruby 2.3.0, rails 4.2.5, papertrail 4.0.2. 

When running 
```ruby
cap staging db:pull
```
I get 
```ruby
rake aborted! 
ActiveRecord::StatementInvalid: Mysql2::Error: Table 'my_database_development.version_associations' doesn't exist: SHOW FULL FIELDS FROM `version_associations`
```
What strange is that I do not have the table:
```ruby
PaperTrail::VersionAssociation.table_exists?
#=> false
```
```ruby
# Gemfile.lock
paper_trail (4.0.2)  # had the same issue with 4.0.0
```
Anything I can do here?

Thanks!

jaredbeck:
Can you reproduce this issue in a brand new rails app without capistrano?

piratebroadcast:
Im getting a similar error in rails admin in a fresh rails 4 app with a handful of models - When I hit the 
/admin path, it blows up because versions_associations does not exist. However, versions_acssociations is listed in the left sidebar and I have a feeling it shouldnt be. See screenshots: 

http://i.imgur.com/OaVFPJp.png

error: http://i.imgur.com/rUlrb3b.png

Full error when loading /admin:

ActiveRecord::StatementInvalid in RailsAdmin::MainController#dashboard
PG::UndefinedTable: ERROR: relation "version_associations" does not exist LINE 5: WHERE a.attrelid = '"version_associations"'::... ^ : SELECT a.attname, format_type(a.atttypid, a.atttypmod), pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod FROM pg_attribute a LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum WHERE a.attrelid = '"version_associations"'::regclass AND a.attnum > 0 AND NOT a.attisdropped ORDER BY a.attnum

hlascelles:
Seeing a similar error, Rails 3.2, paper_trail 4.0.2. Doing a full `db:create db:migrate` from scratch.

```
NoMethodError: undefined method `version' for ActiveModel:Module
config/application.rb:7:in `<top (required)>'
```

Downgrading to paper_trail 4.0.0 solves the issue.

[...]

PR #716, opened 2016-02-19
Added instructions for use with Rails Admin.

piratebroadcast:
Throws an error otherwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,12 @@ has been destroyed.
     bundle exec rails generate paper_trail:install
     bundle exec rake db:migrate
     ```
+    If using RailsAdmin, install with:
+    
+    ```
+    bundle exec rails g paper_trail:install --with-associations
+    bundle exec rake db:migrate
+    ```
 
 1. Add `has_paper_trail` to the models you want to track.
 


seanlinsley:
For context: #707

jaredbeck:
Does `rails_admin` actually require `--with-associations` or is this just a workaround?

piratebroadcast:
Throws errors without --with-associations, papertrail and railsadmin
unusable without the --with-associations command.

On Sun, Feb 21, 2016 at 1:37 PM, Jared Beck <notifications@github.com>
wrote:

> Does rails_admin actually require --with-associations or is this just a
> workaround?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/airblade/paper_trail/pull/716#issuecomment-186882998>.
>



-- 
Jesse Waites
JesseWaites.com

jaredbeck:
Interestingly, it looks like the [rails admin test suite][1] only tests against `paper_trail ~> 3.0`.

PaperTrail is designed to work without tracking associations. In fact, tracking associations is an experimental feature with known issues. So, telling people to turn on associations tracking is not my preferred solution. 

However, I'm inclined to document this workaround anyway, because I think a good chunk of paper trail users are rails admin users (10-20% maybe) and because Jesse (@piratebroadcast) is [not the first to suggest this workaround][2]. 

[1]: https://github.com/sferik/rails_admin/blob/master/gemfiles/rails_4.2.gemfile#L12
[2]: https://github.com/sferik/rails_admin/issues/2405

PR #717, opened 2016-02-22
Add a way to disable in all threads

jaredbeck:
Expands on @Hermanverschooten's suggestion in https://github.com/airblade/paper_trail/pull/695, adds a way to disable PT in all threads, hopefully fixes https://github.com/airblade/paper_trail/issues/635.

Initially, I was concerned about having both a thread-safe flag and thread-unsafe flag.  This PR mitigates that concern by giving these methods very explicit names, like `enabled_in_all_threads=` and `enabled_in_current_thread=`.

- Renames config methods to clarify which are thread-safe and which are not. Deprecates old methods whose names were unclear.
- Adds rails config option: `config.paper_trail.enabled`. See initializer in `paper_trail/frameworks/rails/engine.rb`
- Organizes docs re: the many ways of disabling

Ben, I liked your suggestion in https://github.com/airblade/paper_trail/issues/635 that the all-threads flag be immutable, but I didn't see immediately how to implement said immutability given the way we apply the rails config option in `engine.rb`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,10 +37,13 @@
   [#614](https://github.com/airblade/paper_trail/pull/614)
 - Added `unversioned_attributes` option to `reify`.
   [#579](https://github.com/airblade/paper_trail/pull/579)
+- Added a `config.paper_trail.enabled` option for controlling PaperTrail from the environment config.
+  [#695](https://github.com/airblade/paper_trail/pull/695)
 
 ### Fixed
 
-- None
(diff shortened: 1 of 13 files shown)


jaredbeck:
https://github.com/airblade/paper_trail/pull/720 is a simpler alternative to this, based on conversation with @seanlinsley in https://github.com/airblade/paper_trail/issues/635

jaredbeck:
Closed by #723, a mutually exclusive alternative that fixes #635 but not #584.

PR #720, opened 2016-02-24
PaperTrail.enabled now affects all threads, again

jaredbeck:
This is a simpler alternative to https://github.com/airblade/paper_trail/pull/717 based on conversation with @seanlinsley in https://github.com/airblade/paper_trail/issues/635

As with https://github.com/airblade/paper_trail/pull/717, this expands on @Hermanverschooten's suggestion in https://github.com/airblade/paper_trail/pull/695, particularly re: the `Railtie`.

### DIFF
--- a/lib/paper_trail/frameworks/active_record.rb
+++ b/lib/paper_trail/frameworks/active_record.rb
@@ -2,3 +2,7 @@
 # since otherwise the model(s) will get loaded in via the `Rails::Engine`.
 require "paper_trail/frameworks/active_record/models/paper_trail/version_association"
 require "paper_trail/frameworks/active_record/models/paper_trail/version"
+
+ActiveSupport.on_load(:active_record) do
+  include PaperTrail::Model
+end
--- a/lib/paper_trail/frameworks/rails/engine.rb
+++ b/lib/paper_trail/frameworks/rails/engine.rb
@@ -2,6 +2,11 @@ module PaperTrail
   module Rails
     class Engine < ::Rails::Engine
       paths['app/models'] << 'lib/paper_trail/frameworks/active_record/models'
+      config.paper_trail = ActiveSupport::OrderedOptions.new
+      initializer 'paper_trail.initialisation' do |app|
+        ActiveRecord::Base.send :include, PaperTrail::Model
+        PaperTrail.enabled = app.config.paper_trail.fetch(:enabled, true)
+      end
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,10 +37,13 @@
   [#614](https://github.com/airblade/paper_trail/pull/614)
 - Added `unversioned_attributes` option to `reify`.
   [#579](https://github.com/airblade/paper_trail/pull/579)
+- Added a `config.paper_trail.enabled` option for controlling PaperTrail from the environment config.
+  [#695](https://github.com/airblade/paper_trail/pull/695)
 
 ### Fixed
 
-- None
+- Fixed deprecation warning for Active Record after_callback / after_commit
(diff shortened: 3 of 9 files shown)


seanlinsley:
Pinging @stormsilver since they originally made the change in #541, and may be able to provide more context.

stormsilver:
Okay, I think I'm all caught up on the background here. :)

The context for #541 is when you are using PaperTrail in a multi-threaded environment and you would like to turn it off for some threads but not all.

In our case, we are using Sidekiq. We have some jobs that need to run _with_ PaperTrail, but some jobs (large data imports) that need to run _without_ PaperTrail. So the goal is to be able to disable PaperTrail on a thread-by-thread basis without affecting whether it's enabled for other threads.

I can't remember all of the vagaries of instance variables and threads, but it looks like this patch would remove the ability to selectively disable PaperTrail for some threads. To that end, the modifications in #695 and #717 seem best - you can set an "overall" disabled status which then becomes the default for the per-thread status.

jaredbeck:
Thanks for the quick reply, Eric. I think if we merged this (#720) you could still use `PaperTrail.enabled_for_controller=` for your use case (background jobs) is that not correct? Note that it is poorly named. A better name would be `PaperTrail.enabled_for_current_thread=`.

stormsilver:
Sure - it seems like some people want a way to globally disable PaperTrail and I have a use case for doing it per-thread. If we can come up with mechanisms for accomplishing both of those goals then that seems like a good solution!

jaredbeck:
Eric, Can you please try using `enabled_for_controller=` instead of `enabled=` and confirm that it works for your background jobs? Thanks.

lib/paper_trail/frameworks/active_record.rb
```diff
@@ -2,3 +2,7 @@
 # since otherwise the model(s) will get loaded in via the `Rails::Engine`.
 require "paper_trail/frameworks/active_record/models/paper_trail/version_association"
 require "paper_trail/frameworks/active_record/models/paper_trail/version"
+
+ActiveSupport.on_load(:active_record) do
```
batter: I think this needs to remain within `lib/paper_trail.rb` because this file only conditionally gets loaded when `ActiveRecord` is utilized outside of the context of rails.  Or perhaps there should be an `lib/paper_trail/frameworks/active_model.rb`. Regardless not sure it belongs as part of this PR, seems unrelated. Happy to make a PR for that piece separately if you'd like.

batter: Oh I see now that you also added a similar block inside of the `Rails::Engine`. Did this have issues too related to thread safety?

lib/paper_trail/frameworks/rails/engine.rb
```diff
@@ -2,6 +2,11 @@ module PaperTrail
   module Rails
     class Engine < ::Rails::Engine
       paths['app/models'] << 'lib/paper_trail/frameworks/active_record/models'
+      config.paper_trail = ActiveSupport::OrderedOptions.new
```
batter: Can you explain how this works? I never understood what is supposed to go in the `Rails::Engine` beyond code for autoloading libraries, this looks neat but I don't understand it

batter: cc @Hermanverschooten

jaredbeck:
> .. not sure it belongs as part of this PR, seems unrelated ..

Yeah, this and #717 are based on #695 which really does two things. To make this easier to review I've created https://github.com/airblade/paper_trail/pull/723, a subset of this. Let's start there and then come back to this.

jaredbeck:
Closed by #723, a subset of this PR; it fixes #635 but not #584. I'll fix the conflicts in this and open a new PR with Herman's work.

PR #723, opened 2016-03-03
Make PaperTrail.enabled= affect all threads

jaredbeck:
- Fixes https://github.com/airblade/paper_trail/issues/635
- A subset of https://github.com/airblade/paper_trail/pull/720, this PR does not include the railtie config option


batter:
:+1:

PR #726, opened 2016-03-06
Optimize post-rollback association reset

### LINKED ISSUE #715: Querying for PaperTrail::Version in tests despite PaperTrail being disabled
Hello,
I have a Rails application with PaperTrail disabled in `config/environment/test.rb`:
```ruby
config.after_initialize do
  PaperTrail.enabled = false
  PaperTrail.enabled_for_controller = false
end
```
As expected, this stops *Version*s being created in test environment, but when I look at `test.log` there's still plenty of `PaperTrail::Version Load`s on model creation.

I benchmarked my test suite speed *before* and *after* removing all `has_paper_trail`s from my app, and the improvement is quite significant.

With `has_paper_trail` (median of 3):
> real	1m54.697s
> user	1m33.708s
> sys	0m2.536s

Without (median of 3):
> real	1m11.597s
> user	0m54.632s
> sys	0m1.768s

I assume there is currently no setting to disable PaperTrail entirely in test environment, but could it be implemented?

jaredbeck:
[Fixes #715]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@
 
 ### Fixed
 
+- [#715](https://github.com/airblade/paper_trail/issues/715) -
+  Optimize post-rollback association reset.
 - [#701](https://github.com/airblade/paper_trail/pull/701) /
   [#699](https://github.com/airblade/paper_trail/issues/699) -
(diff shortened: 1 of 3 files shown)


batter:
:+1:

PR #730, opened 2016-03-11
Support for reifying belongs_to relationships

theRealNG:
Support for reifying belongs_to relationships.
https://github.com/airblade/paper_trail/issues/503

So you can call @project.versions.last.reify( belongs_to: true) and get the client model at the time of last change of the project model.

### DIFF
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -13,6 +13,7 @@ module PaperTrail
           mark_for_destruction: false,
           has_one: false,
           has_many: false,
+          belongs_to: false,
           unversioned_attributes: :nil
         )
 
@@ -60,15 +61,6 @@ module PaperTrail
 
       private
 
-      def reify_associations(model, options, version)
-        if options[:has_one]
-          reify_has_ones version.transaction_id, model, options
-        end
-        if options[:has_many]
-          reify_has_manys version.transaction_id, model, options
-        end
-      end
-
       # Set all the attributes in this version on the model.
       def reify_attributes(model, version, attrs)
         enums = model.class.respond_to?(:defined_enums) ? model.class.defined_enums : {}
@@ -114,7 +106,7 @@ module PaperTrail
           elsif version.event == "create"
             options[:mark_for_destruction] ? record.tap(&:mark_for_destruction) : nil
           else
-            version.reify(options.merge(has_many: false, has_one: false))
+            version.reify(options.merge(has_many: false, has_one: false, belongs_to: false))
           end
         end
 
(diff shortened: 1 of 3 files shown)


theRealNG:
Need help. RuboCop tests are failing because of Cyclomatic and Perceived complexity for reify is too high. Don't know how to fix it.

lib/paper_trail/reifier.rb
```diff
@@ -160,6 +165,32 @@ def reify_has_ones(transaction_id, model, options = {})
...
+          collection = if version.nil?
+                         assoc.klass.where(assoc.klass.primary_key => collection_key).first
+                       elsif version.event == "create"
+                         options[:mark_for_destruction] ? collection.mark_for_destruction : nil
+                       else
+                         version.reify(options.merge(has_many: false, has_one: false,
+                                                     belongs_to: false))
+                       end
+
+          model.send("#{assoc.name}=".to_sym, collection)
```
seanlinsley: `to_sym` isn't necessary here -- `send` automatically does that for you

theRealNG: I think the "send" method in ruby v1.9.3 accepts only a symbol.

jaredbeck:
> Need help. RuboCop tests are failing because of Cyclomatic and Perceived complexity for reify is too high. Don't know how to fix it.

I'm sure we can find a way to extract a private method. (http://refactoring.com/catalog/extractMethod.html)

jaredbeck:
> I'm sure we can find a way to extract a private method. (http://refactoring.com/catalog/extractMethod.html)

This should help: https://github.com/airblade/paper_trail/pull/734

Oh!, I see you're also working on that?

jaredbeck:
> Oh!, I see you're also working on that?

Haha, we came up with the same thing :D

theRealNG:
Refactored the "reify_associations" model to return the model, so that in future we can use it in any other place if required.

lib/paper_trail/reifier.rb
```diff
@@ -166,6 +167,32 @@ def reify_has_ones(transaction_id, model, options = {})
...
+          version = assoc.klass.paper_trail_version_class.
+            where("item_type = ?", assoc.class_name).
+            where("item_id = ?", collection_key).
+            where("created_at >= ? OR transaction_id = ?", options[:version_at], transaction_id).
+            order("id").limit(1).first
+
+          collection = if version.nil?
+                         assoc.klass.where(assoc.klass.primary_key => collection_key).first
+                       elsif version.event == "create"
+                         options[:mark_for_destruction] ? collection.mark_for_destruction : nil
```
jaredbeck: Is there a test that covers this line? It seems like the reference to `collection` here would raise a `NameError`. What am I missing?

theRealNG: Nice catch. But I don't think this condition is ever met as the base model is always created before the associated model. Ex:- Widget instance is created before the wotsit instance.

jaredbeck:
Please rebase on master, squash commits, and add a changelog entry. Thanks.

theRealNG:
@jaredbeck  Done

lib/paper_trail/reifier.rb
```diff
@@ -55,20 +56,10 @@ def reify(version, options)
         reify_attributes(model, version, attrs)
         model.send "#{model.class.version_association_name}=", version
         reify_associations(model, options, version)
-        model
```
jaredbeck: I think I'd rather keep the explicit return value here, instead of moving it to `reify_associations`.

jaredbeck:
I've read through your tests and they look very thorough, thanks!

Other than my minor point about the return value of `reify_associations`, I think this is ready to merge.

theRealNG:
@jaredbeck Changes the the return value of `reify_associations` as suggested.

jaredbeck:
Merged. Thanks NG! Please submit a changelog entry when you get a chance.

PR #731, opened 2016-03-11
Maps enums to database values before storing in `object_changes`

cbarton:
Keep consistency between versions with regard to `changes` and
`object_changes` and how enum columns store their values.

Before, `changes` would map the changed attributes enum columns to the database
values (integer values). This allows reifying that version to maintain the
integrity of the enum. It did not do so for `object_changes` and thus, `0`
for non-json columns, and the enum value for json columns would be stored instead.
For the non-json columns, it mapped any non-integer enum value to `0` because
during serialization that column is an `integer`.  Now this is fixed,
so that `object_changes` stores the enum mapped value.

Here is an example:

```ruby
class PostWithStatus < ActiveRecord::Base
  has_paper_trail
  enum status: { draft: 0, published: 1, archived: 2 }
end

post = PostWithStatus.new(status: :draft)
post.published!
version = post.versions.last

# Before
version.changeset #> { 'status' => ['draft', 'draft'] } (stored as [0, 0])

# After
version.changeset #> { 'status' => ['draft', 'published'] } (stored as [0, 1])
```

### DIFF
--- a/lib/paper_trail/attributes_serialization.rb
+++ b/lib/paper_trail/attributes_serialization.rb
@@ -25,12 +25,52 @@ module PaperTrail
       end
     end
 
-    SERIALIZE, DESERIALIZE =
-      if ::ActiveRecord::VERSION::MAJOR >= 5
-        [:serialize, :deserialize]
-      else
-        [:type_cast_for_database, :type_cast_from_database]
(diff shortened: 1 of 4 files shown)


jaredbeck:
Looks good, Chris, thanks!

PR #732, opened 2016-03-13
support for choosing association to track

theRealNG:
It looks like we are saving into version_associations only when the model that is being updated has a belongs_to association.

So basically added a small check in the save_associations method. I'm not sure this is the correct way. If it is fine then I will write specs for the change.

https://github.com/airblade/paper_trail/issues/639


seanlinsley:
Could you add a test for this?

theRealNG:
Need some help here. Not able to figure out where to write specs for this feature.

PR #737, opened 2016-03-21
Allows to use default `.meta` from VersionConcern

ignat-zakrevsky:
In some cases, you want to store the same information in more than one model. It can
be some shared method or kind of interface that implements method through a lot
of models. Such approach gives the ability to reduce the amount of similar code.

Example:

```ruby
class AuthorLog < PaperTrail::Version
  self.table_name = :author_logs

  def self.meta
    { :author_id => :author_id }
  end
end

class Book
  belongs_to :author
  has_paper_trail :class_name => "AuthorLog"
end

class Comment
  belongs_to :author
  has_paper_trail :class_name => "AuthorLog"
end
```

Another variant to use it is nested models

```ruby
class Book
  belongs_to :author
  has_paper_trail :class_name => "AuthorLog"
end

class Mention
  belongs_to :book
  delegate :author_id, :to => :book
  has_paper_trail :class_name => "AuthorLog"
end
```

BTW, which kind of hash syntax should I use? Currently using rocket syntax as used in README and a lot of tests

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -3,7 +3,7 @@
 
 # Offense count: 19
 Metrics/AbcSize:
-  Max: 159
+  Max: 160
 
 # Offense count: 1
 Metrics/BlockNesting:
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -82,7 +82,7 @@ module PaperTrail
           }
         end
 
-        paper_trail_options[:meta] ||= {}
+        paper_trail_options[:meta] ||= paper_trail_version_class.meta
         paper_trail_options[:save_changes] = true if paper_trail_options[:save_changes].nil?
 
         class_attribute :versions_association_name
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -56,6 +56,10 @@ module PaperTrail
         where "event <> ?", "create"
       end
 
+      def meta
+        {}
+      end
+
       # Returns versions after `obj`.
       #
       # @param obj - a `Version` or a timestamp
--- a/spec/modules/version_concern_spec.rb
+++ b/spec/modules/version_concern_spec.rb
@@ -17,6 +17,11 @@ describe PaperTrail::VersionConcern do
     expect(Bar::Document.version_class_name).to eq("Bar::Version")
   end
 
+  it "defines default meta value" do
+    expect(Foo::Version.meta).to eq({})
+    expect(Bar::Version.meta).to eq({})
+  end
+
   describe "persistence", versioning: true do
     before do
       @foo_doc = Foo::Document.create!(name: "foobar")
--- a/test/dummy/app/versions/post_version.rb
+++ b/test/dummy/app/versions/post_version.rb
@@ -1,3 +1,9 @@
 class PostVersion < PaperTrail::Version
   self.table_name = "post_versions"
+
+  attr_accessible :comments_count if ::PaperTrail.active_record_protected_attributes?
+
+  def self.meta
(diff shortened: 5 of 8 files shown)


.rubocop_todo.yml
```diff
@@ -3,7 +3,7 @@
 
 # Offense count: 19
 Metrics/AbcSize:
-  Max: 159
+  Max: 160
```
ignat-zakrevsky: only for `SetUpTestTables.up` method

jaredbeck:
Hi Ignat, thanks for the contribution!

I'd like to compare your suggestion to other ways of achieving the same level of code reuse in plain ruby. For example,

```ruby
module HasAuthorLog
  def paper_trail_metadata_config
    { :author_id => :author_id }
  end
end

class Book
  extend HasAuthorLog
  belongs_to :author
  has_paper_trail :class_name => "AuthorLog", :meta => paper_trail_metadata_config
end

class Comment
  extend HasAuthorLog
  belongs_to :author
  has_paper_trail :class_name => "AuthorLog", :meta => paper_trail_metadata_config
end
```

Given that this can already be done in plain ruby, I'm disinclined to add a feature to PT, but maybe I'm missing something.

ignat-zakrevsky:
Hm, I agree with this but in same time `PaperTrail::Version` 	successors seems a little weird with only table name redefining, such approach moves logic related to PaperTrail in PaperTrail-connected class wich may breathe a new life into them (for example -- use same `ignore` keys). Additionally, modules method overload scope with method related only to PaperTrail. Anyway -- thanks for your approach, I will use it in case if you close PR.

jaredbeck:
> Hm, I agree with this but in same time PaperTrail::Version successors seems a little weird with only table name redefining, such approach moves logic related to PaperTrail in PaperTrail-connected class wich may breathe a new life into them (for example -- use same ignore keys).

The child version model is a logical place to put these defaults, but I have two concerns.

1. It introduces a *second* configuration interface. Currently, there is only one configuration interface, `has_paper_trail`. Should we add a second interface?
2. The configuration of `:ignore` refers to attributes of the app model, not the version model. For example, `title` is an attribute of `Post`, not an attribute of `PostVersion`. So, it seems to me that the configuration of `:ignore` should remain in the `Post` model.

ignat-zakrevsky:
Yep, shared `:ignore` is non-typical using case, but in the same time you can store `version_associations` table name which isn't connected with auditable AR model. I can try to give an example with `DeleteVersion` which watch only delete actions and you want to share `:on` option through all watching models.

The main idea of a second configuration interface is to provide user ability to change default values per `PaperTrail::Version`. I'm agree, that using both interfaces isn't good idea but there is no any problems with using only one of theь. When you see `has_paper_trail :class_name => "AuthorLog"` you understand that there is PT but you don't need to know all that happens underhood, as bonus you have more DRY models with same PT.

Currently for models with shared logic I'm using this: 
```ruby
has_paper_trail class_name: EventLog.to_s, meta: EventLog.meta
```
but this a little weird.

Your approach with modules good enough. But it generates additional module which I need to include in each model when this module will contain only `has_paper_trail` instruction.

jaredbeck:
@batter Do you have an opinion on this? I'm thinking there shouldn't be two ways to configure the same thing, so I'm leaning towards closing this, but I don't feel really strongly about it.

jaredbeck:
Ignat, I'm going to close this for the reasons I've given, but thank you for your contribution and I hope you'll contribute again!

batter:
@jaredbeck - I'm inclined to agree with you. Even if you want to share it through a lot of models as the OP states you can still do that type of thing by making a mixin module that defines the argument you want to pass to the `meta` option. This seems like an unnecessary feature that I doubt would be leveraged frequently.

Issue #465, opened 2015-01-20
When not using new version_associations calling destroy on a version will generate an error


frenkel:
I just upgraded an existing project to paper_trail 4.0.0.beta2 and was presented by this error when running my test suite:

    ActiveRecord::StatementInvalid: Could not find table 'version_associations'

I think this is caused by the `dependent: :destroy` in PaperTrail::VersionConcern in the `has_many :version_assosciations`. What would be the best way to work around this? Is it possible to fix this in the concern, or should I just call delete instead?

batter:
Thanks for reporting this! I need to do a better job of testing this feature out extensively on dummy apps outside of the gem test framework to really try and push all possible outcomes.

frenkel:
No problem, we found it before an official release :)

emmanne08:
Hello @batter and @frenkel:

I happened to encounter this issue on PaperTrail 4.0.0.beta3.

I'm using Rails 3.2.21 and Ruby 1.9.3-p551. Initially, I ran `rails generate paper_trail:install` and I opted to use a separate table for my Post model and named it post_versions instead of the default versions. I generated a migration that's similar to the CreateVersions migration wherein I just configured it to use 'post_versions' instead of 'versions'. After running the migrations, I now have the versions and the post_versions table. I generated a model for the PostVersion that I sub-classed from PaperTrail::Version as instructed in the PaperTrail README.md. I didn't forget to set the table name to 'post_versions' for the PostVersion model, I also set `has_paper_trail class_name: 'PostVersion'` within the Post model.

Now when I'm on within the Rails console, I created a Post model and a version of it was automatically created as well. If I call destroy on that created version, I get an error as described below:

    [3] pry(main)> post.versions.last.destroy
      PostVersion Load (0.7ms)  SELECT `post_versions`.* FROM `post_versions` WHERE `post_versions`.`item_id` = 1 AND `post_versions`.`item_type` = 'Post' ORDER BY `post_versions`.`created_at` DESC, `post_versions`.`id` DESC LIMIT 1
       (0.3ms)  BEGIN
       (0.4ms)  ROLLBACK
    ActiveRecord::StatementInvalid: Mysql2::Error: Table 'dummy_2_development.version_associations' doesn't exist: SHOW FULL FIELDS FROM `version_associations`
[...]

PR #742, opened 2016-03-29
Check connection to database is present for validating table existence.

josisusan:
Check if database connection is present before validation of table existence.

### DIFF
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -34,7 +34,13 @@ module PaperTrail
 
     def track_associations?
       if @track_associations.nil?
-        PaperTrail::VersionAssociation.table_exists?
+        # Validate presence of database before validating
+        # table existence
+        begin
+          PaperTrail::VersionAssociation.table_exists?
+        rescue ActiveRecord::NoDatabaseError
+          false
+        end
       else
         @track_associations
       end
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -38,8 +38,19 @@ module PaperTrail
         # table existence
         begin
           PaperTrail::VersionAssociation.table_exists?
-        rescue ActiveRecord::NoDatabaseError
-          false
+        rescue => error
+          case error.class.to_s
+          when "PG::ConnectionBad"
+            postgres_error_connection_handler(error)
+          when "ActiveRecord::ConnectionNotEstablished" # For Mysql2
+            false
+          when "SQLite3::CantOpenException"
+            sqlite_error_connection_handler(error)
+          when "ActiveRecord::NoDatabaseError"
+            false
+          else
+            raise error
+          end
         end
       else
         @track_associations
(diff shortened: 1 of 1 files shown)


jaredbeck:
I think `NoDatabaseError` was added in rails 4, but we want PT to still support rails 3.

batter:
I need to look into this but it would be splendid if there was a simple way
to just check ActiveRecord::Base.connection_established? Or something to
that effect

jaredbeck:
> I need to look into this but it would be splendid if there was a simple way to just check ActiveRecord::Base.connection_established? Or something to that effect

There is `ActiveRecord::ConnectionHandling.connected?`.

```
ActiveRecord::Base.connected? # => true
PaperTrail::Version.connected? # => true
```

seanlinsley:
It seems possible that there could be no existing database connection, but a completely valid database.

Bivek:
Yes, unless you don't make the real connection, `conntected?` will always be false

josisusan:
@jaredbeck I have tried using PaperTrail::Version.connected? but it gives false like @Bivek have mentioned.

batter:
@josisusan and @jaredbeck - that is the problem with `#connected?`, it will return `false` if you have not made a query against the table within a current session. That seems like unexpected behavior to me but is how it works currently as far as I understand.

jaredbeck:
I thought we couldn't use `NoDatabaseError` because we still support rails 3, but interestingly, having an undefined constant in a `rescue` does not cause a `NameError`.

```
puts "interesting" rescue ConstantThatDoesNotExist # no `NameError`
```

So, I guess we could use `NoDatabaseError` if we had to, even though it won't help rails 3 users. I'd be OK with that.

Susan, can you rebase please?

josisusan:
We (with @Bivek) are figuring out some alternatives to support rails 3 with snippets below:
```ruby
begin
  PaperTrail::VersionAssociation.table_exists?
rescue => error
  case error.class.to_s
  when "::PG::Error", "PG::ConnectionBad"
     error.message.include?("does not exist") ? false : raise(error)
  when "Mysql2::Error"
    ...
  when "ActiveRecord::NoDatabaseError"
    false
 ...
```

jaredbeck:
It's funny how big this method has grown. Originally it was just the `table_exists?` I think :grin:

jaredbeck:
> We (with @Bivek) are figuring out some alternatives to support rails 3 ..

Btw if you want to run the tests against rails 3 locally, you can use `bundle exec appraisal ar3 rake`. See the [Contributing Guide][1] for more tips.

[1]: https://github.com/airblade/paper_trail/blob/master/.github/CONTRIBUTING.md

seanlinsley:
Why do we want this behavior in the first place? I think it would have been fine to keep the behavior we had before #465 (only raising an error once something else tries accessing the versions table).

I go into more detail here: https://github.com/airblade/paper_trail/issues/740#issuecomment-202702068

jaredbeck:
> Why do we want this behavior in the first place? I think it would have been fine to keep the behavior we had before #465 (only raising an error once something else tries accessing the versions table).

For backwards compatibility, we don't want to require people to have the `versions_associations` table. If we required everyone to have the `versions_associations` table, we might not need *any* of this complexity. It's a tempting solution :smiling_imp:

jaredbeck:
> I think it would have been fine to keep the behavior we had before #465 (only raising an error once something else tries accessing the versions table

@seanlinsley can you put together a PR for that so we can compare? Thanks.

jaredbeck:
Another alternative is to require users to configure this. Susan, Sean, Ben, please take a look at https://github.com/airblade/paper_trail/pull/785, thanks.

jaredbeck:
Closed by https://github.com/airblade/paper_trail/pull/785

Thanks Susan, Bardan, Sean, and Ben!

PR #744, opened 2016-03-30
update changelog about support for reifying belongs_to associations

theRealNG:
@jaredbeck sorry forgot to update the changelog for pull request: https://github.com/airblade/paper_trail/pull/730


theRealNG:
I added the change log entry under the "Added" section. Not sure if this is the correct place.
Added it the "Added" section as initially there was no support for reifying a belongs_to association, and now we have added support for it.

jaredbeck:
> I added the change log entry under the "Added" section

Perfect, thanks! See http://keepachangelog.com/ for more info.

PR #749, opened 2016-04-01
Rubocop cleanup

magni-:
This is just some general cleanup to remove a lot of the TODOs from `.rubocop_todos.yml`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,6 +23,9 @@ Metrics/LineLength:
 
 # Migrations often contain long up/down methods, and extracting smaller methods
 # from these is of questionable value.
+Metrics/AbcSize:
+  Exclude:
+    - 'test/dummy/db/migrate/*'
 Metrics/MethodLength:
   Exclude:
     - 'test/dummy/db/migrate/*'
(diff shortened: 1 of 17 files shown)


jaredbeck:
Two problems:

1. Tests are failing
2. Please do not use rubocop comments like `# rubocop: disable`

jaredbeck:
I think the `NameError` (https://travis-ci.org/airblade/paper_trail/jobs/119997293)

```
undefined method 'without' for class #<Class:ARIdMapTest::ActiveRecord::IdentityMap>
```

is caused by 279917857b53d4def0812f3c14d600910a54ef51

magni-:
Hmm, tests were passing locally with each commit, I'll take a look at the failures this weekend.

As for the `disable` comments, what do you suggest? IMO it's better to not disable checks for an entire file for known false positives, but rather to disable the individual bits where those false positives are, in order to prevent true Rubocop offenses from getting added to those files.

The `set_transaction_id` offense is subjective - I _could_ rename it to `transaction_id=` (I explained why I didn't in the commit message). But the `Alias` false positive is an issue with Rubocop not parsing the `instance_eval` blocks properly. Making Rubocop happy for that case would actually end up breaking things.

jaredbeck:
Closed via https://github.com/airblade/paper_trail/pull/750 which is a subset of this PR.

Thanks Paul!

Please submit the other more controversial commits separately, to make discussion easier. Thanks!

PR #750, opened 2016-04-01
Linting by magni

jaredbeck:
Cherry-picking the uncontroversial commits from https://github.com/airblade/paper_trail/pull/749. Isn't git great? :D

@magni-

PR #752, opened 2016-04-02
Rubocop cleanup take 2

magni-:
WIP following #749 and #750 

Repasting my comment from #749 here:

> As for the disable comments, what do you suggest? IMO it's better to not disable checks for an entire file for known false positives, but rather to disable the individual bits where those false positives are, in order to prevent true Rubocop offenses from getting added to those files.
>
> The set_transaction_id offense is subjective - I could rename it to transaction_id= (I explained why I didn't in the commit message). But the Alias false positive is an issue with Rubocop not parsing the instance_eval blocks properly. Making Rubocop happy for that case would actually end up breaking things.

Tests should fail, since I still need to look at the failures from #749

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,12 @@ AllCops:
   Exclude:
     - test/dummy/db/schema.rb
 
+# Migrations often contain long up/down methods, and extracting smaller methods
+# from these is of questionable value.
+Metrics/AbcSize:
+  Exclude:
+    - 'test/dummy/db/migrate/*'
+
 Metrics/ClassLength:
   Exclude:
     - test/**/*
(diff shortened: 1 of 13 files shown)


jaredbeck:
Thanks Paul, but, per my comment in https://github.com/airblade/paper_trail/pull/749, I was hoping we could break this up into smaller chunks, to make discussing some of the more controversial stuff easier? Do you mind closing this and doing a few smaller PRs? In addition to making discussion easier, it should also help with getting the test suite to pass. Thanks.

magni-:
Ah my bad, I misunderstood that comment. As for the failures, they're all from AR 3 builds, from the change made in https://github.com/airblade/paper_trail/pull/752/commits/f6e846b9ab51f7e7e8404a247c508930d715e815. I didn't have time to investigate this weekend but will be looking at this today.

I'll strip out that commit for now while I look into why AR 3 didn't like that change.

PR #753, opened 2016-04-04
Exclude migration file from Metrics/AbcSize check

magni-:
First of many smaller PRs from #752

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -3,8 +3,6 @@
 
 # Offense count: 19
 Metrics/AbcSize:
-  Exclude:
-    - 'test/dummy/db/migrate/20110208155312_set_up_test_tables.rb'
   Max: 46 # Goal: 15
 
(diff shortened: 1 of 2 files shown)


.rubocop_todo.yml
```diff
@@ -3,9 +3,7 @@
 
 # Offense count: 19
 Metrics/AbcSize:
-  Exclude:
-    - 'test/dummy/db/migrate/20110208155312_set_up_test_tables.rb'
-  Max: 46 # Goal: 15
+  Max: 50
```
jaredbeck: I don't understand why the `Max` increased here. `Max` should either stay the same, or decrease.

magni-: Whoops, looks like a bad rebase, I'll fix :+1:

magni-: Fixed

jaredbeck:
Thanks!

PR #754, opened 2016-04-04
Fix Style/MultilineTernaryOperator offenses

magni-:
Yet another small PR from #752

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -221,13 +221,7 @@ module PaperTrail
 
       # Returns the object (not a Version) as it was most recently.
       def previous_version
-        preceding_version =
-          if source_version
-            source_version.previous
-          else
-            send(self.class.versions_association_name).last
-          end
-        preceding_version.reify if preceding_version
+        preceding_version.try(:reify)
       end
 
       # Returns the object (not a Version) as it became next.
@@ -307,6 +301,14 @@ module PaperTrail
 
       def source_version
         send self.class.version_association_name
+      end
+
+      def preceding_version
+        if source_version
+          source_version.previous
+        else
+          send(self.class.versions_association_name).last
+        end
       end
 
       def record_create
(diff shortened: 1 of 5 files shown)


jaredbeck:
These are all fine. I'd prefer to extract private methods, but that can happen later. 

Please rebase.

lib/paper_trail/reifier.rb
```diff
@@ -284,9 +282,8 @@ def reify_has_many_through(transaction_id, associations, model, options = {})
       # table, this method would return the constant `Fruit`.
       def version_reification_class(version, attrs)
         inheritance_column_name = version.item_type.constantize.inheritance_column
-        class_name = attrs[inheritance_column_name].blank? ?
```
magni-: Do people ever set `inheritance_column` to `""` ? That's the only case in which this would break...

Happy to revert this change if needed. Looking through `git`, it's been `blank?` since the code was first written in 2009 (so not in response to a bug report from someone using `inheritance_column = ""`).

jaredbeck: > Do people ever set inheritance_column to "" ?

Seems unlikely, but we should be changing behavior in a linter commit. If we are going to make that change, please make a separate PR. Thanks.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -306,6 +303,14 @@ def source_version
         send self.class.version_association_name
       end
 
+      def preceding_version
+        if source_version
+          source_version.previous
+        else
+          send(self.class.versions_association_name).last
+        end
+      end
```
jaredbeck: I know I said I'd prefer to extract private methods, but we can't do that here because we already have a problem with polluting people's AR models with a ton of methods, and I don't want to make that problem worse. See discussion in https://github.com/airblade/paper_trail/pull/719

jaredbeck:
Closed by https://github.com/airblade/paper_trail/pull/768

PR #755, opened 2016-04-04
Fix Style/MultilineBlockChain offenses

magni-:
From #752


jaredbeck:
This one is fine with me. I don't have a particularly strong opinion on chaining these kind of methods, so I'm happy to go with rubocop's recommendation.

PR #756, opened 2016-04-04
Fix Style/ModuleFunction offenses

magni-:
From #752

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -34,9 +34,3 @@ Style/Alias:
 # Offense count: 33
 Style/Documentation:
   Enabled: false
-
-# Offense count: 2
-Style/ModuleFunction:
-  Exclude:
-    - 'lib/paper_trail/serializers/json.rb'
-    - 'lib/paper_trail/serializers/yaml.rb'
--- a/lib/paper_trail/serializers/json.rb
+++ b/lib/paper_trail/serializers/json.rb
@@ -3,15 +3,15 @@ require "active_support/json"
 module PaperTrail
   module Serializers
     module JSON
-      extend self # makes all instance methods become module methods as well
-
       def load(string)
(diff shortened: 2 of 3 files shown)


jaredbeck:
Do the methods in these serializer modules really need to be instance methods and module methods at the same time? That seems like a poor design (not your fault, obviously). If we can settle on just instance methods, or just module methods, that would be my preference.

magni-:
TBH I haven't really looked into where those methods were being used. I'll take a look today.

magni-:
https://github.com/airblade/paper_trail/pull/196 seems to be where the dual module/instance use comes from.

jaredbeck:
>> Do the methods in these serializer modules really need to be instance methods and module methods at the same time?
>
> #196 seems to be where the dual module/instance use comes from.

I see, there was an actual use case for this. In that case, the design is justified and I'm fine with this PR as is. However @batter might prefer to keep the `extend self`, since he's the one who wrote that line. I'll let him decide what to do with this PR.

jaredbeck:
Hmmm. I looked into this a bit more, and this seems to be a breaking change, because `module_function` makes them private.

```ruby
# `extend self` makes the copies public
module Pub
  extend self
  def x; end
end

class PubClass
  include Pub
end

PubClass.new.x #=> nil

# `module_function` makes the copies public
module Priv
  def x; end
  module_function :x
end

class PrivClass
  include Priv
end

PrivClass.new.x
# NoMethodError: private method `x' called for #<PrivClass:0x007fba1c0e7e90>
```

magni-:
Ooh interesting, I did not think about that behavior. What do you think the best course of action is here?

jaredbeck:
Closed by 3b73566ef78162e71ff2fcd144ace2aa2e8e4522

I asked for clarification about this at https://github.com/bbatsov/ruby-style-guide/issues/556

PR #757, opened 2016-04-04
Fix Style/ClassAndModuleChildren offense

magni-:
From #752. Expecting the AR3 builds to fail for now.


test/unit/ar_id_map_test.rb
```diff
@@ -1,26 +1,17 @@
...
-        class << self
-          alias __without without
-          def without(&block)
-            @unclobbered = true
-            __without(&block)
-          end
-        end
-      end
-
+      ActiveRecord::IdentityMap.expects(:without).once
```
magni-: From what I can understand from https://github.com/airblade/paper_trail/commit/5df9949ff41f22b2588915d87deff59ee4ad3974, all this was previously testing for, was that `ActiveRecord::IdentityMap.without` was called. That seems like the perfect use case for a mock.

I prefer avoiding dependencies, but IMO adding a dependency is the lesser evil here compared to the roundabout way of testing this via monkeypatching AR in the test.

The only test in this file only gets run when `ActiveRecord::IdentityMap` is defined, which is to say only for AR3, so I moved the `defined?` check to the top level of the file, so that we can add the new mocha dependency to the AR3 gemfile only.

jaredbeck:
If I understand correctly, this PR adds a dependency on mocha. I'd rather not add any dependencies, if possible. Can we move this test to rspec instead?

batter:
Agreed

magni-:
All for not adding dependencies :+1: I'll move the test

spec/models/widget_spec.rb
```diff
@@ -126,6 +126,19 @@
     end
   end
 
+  if defined?(ActiveRecord::IdentityMap) && ActiveRecord::IdentityMap.respond_to?(:without)
+    describe "IdentityMap", versioning: true do
+      before do
+        widget.update_attributes name: "Henry", created_at: Time.now - 1.day
```
jaredbeck: Missing a line from the old test:

```
@widget.update_attributes name: "Harry"
```

I'm not sure if it matters ..

spec/models/widget_spec.rb
```diff
@@ -126,6 +126,19 @@
...
+      end
+
+      it "should not clobber the IdentityMap when reifying" do
+        expect(ActiveRecord::IdentityMap).to receive(:without).once
```
jaredbeck: This looks so much nicer as a mock.

jaredbeck:
Closed by https://github.com/airblade/paper_trail/pull/767

PR #758, opened 2016-04-04
Fix Style/Alias offenses

magni-:
From #752.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -255,6 +255,7 @@ module PaperTrail
 
       # Utility method for reifying. Anything executed inside the block will
       # appear like a new record.
+      # rubocop: disable Style/Alias
       def appear_as_new_record
         instance_eval {
           alias :old_new_record? :new_record?
@@ -263,6 +264,7 @@ module PaperTrail
         yield
         instance_eval { alias :new_record? :old_new_record? }
       end
+      # rubocop: enable Style/Alias
 
       # Temporarily overwrites the value of whodunnit and then executes the
       # provided block.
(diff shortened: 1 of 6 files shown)


lib/paper_trail/has_paper_trail.rb
```diff
@@ -252,6 +252,7 @@ def without_versioning(method = nil)
 
       # Utility method for reifying. Anything executed inside the block will
       # appear like a new record.
+      # rubocop: disable Style/Alias
```
magni-: Reposting my original comment here:

> But the Alias false positive is an issue with Rubocop not parsing the instance_eval blocks properly. Making Rubocop happy for that case would actually end up breaking things.

magni-:
Not quite sure what is wrong with the build (the only failure is 2.3.0/AR4/Postgres), since everything passes locally with that configuration.

lib/paper_trail/has_paper_trail.rb
magni-: I submitted a fix to Rubocop here: https://github.com/bbatsov/rubocop/pull/3004

jaredbeck:
Please rebase.

jaredbeck:
Looks good.

All the new `alias`es are in the same lexical scope as the methods they alias, so if I understand the style guide (https://github.com/bbatsov/ruby-style-guide#alias-method-lexically) we're good there.

Dropping `track_associations` (without the question mark) could be considered a breaking change, but I doubt anyone's using it and our next release is a major version (5), so I guess it's OK as long as we add a line to the changelog about it.

I don't like inline linter comments, but these are temporary so I can live with it.

Thanks Paul!

PR #759, opened 2016-04-04
Explicitly disable Style/AccessorMethodName for set_transaction_id

magni-:
From #752.

This method is doing a lot more than just setting `transaction_id` for `self`, so to me it doesn't make sense to rename it to `transaction_id=`.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -20,11 +20,6 @@ Metrics/ModuleLength:
 Metrics/PerceivedComplexity:
   Max: 16
 
-# Offense count: 1
-Style/AccessorMethodName:
-  Exclude:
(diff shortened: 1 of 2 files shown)


jaredbeck:
> This method is doing a lot more than just setting transaction_id for self, so to me it doesn't make sense to rename it to transaction_id=.

If it weren't an AR model, I'd be OK with `transaction_id=`. I don't like overriding attribute mutator methods in activerecord.

How about `update_transaction_id`? It hints at the `save`.

Renaming this method is not a breaking change.

I'd rather not use linter comments, if possible.

jaredbeck:
Closed by https://github.com/airblade/paper_trail/pull/763

PR #760, opened 2016-04-04
Fix Style/RescueModifier offenses

magni-:
From #752.


lib/paper_trail/version_concern.rb
```diff
@@ -218,7 +218,7 @@ def changeset
 
     # Returns who put the item into the state stored in this version.
     def paper_trail_originator
-      @paper_trail_originator ||= previous.whodunnit rescue nil
+      @paper_trail_originator ||= previous && previous.whodunnit
```
jaredbeck: I'd rather use `try` here. PT still depends on activesupport >= 3, so we can't use `try!`.

jaredbeck:
> I'd rather use try here.

This is the only necessary change. The rest looks good!

batter:
:+1:

jaredbeck:
Closed by https://github.com/airblade/paper_trail/pull/762

Thanks Paul!

PR #769, opened 2016-04-07
Take private/protected methods into consideration

john-denisov:
Don't forger about private/public methods when checking if value of `:meta` hash for `has_paper_trail` method is a method
http://ruby-doc.org/core-2.3.0/Object.html#method-i-respond_to-3F


Lyubomyr:
Прикольно, гуд!

batter:
Makes sense to me, cheers for the pull request.

PR #771, opened 2016-04-08
Added has_and_belongs_to_many reification

samboylett:
HABTM associations are saved as attributes with the name
association_ids as an array of the associated model IDs. Changes
made to HABTM associations are stored on the model so on save the
original associations can be stored in the serialized object. On
reification, if the has_and_belongs_to_many option is not passed,
the serialized attributes are removed such that they are not
reified. If it is passed, the association is reified but if an
associated object no longer exists, it is not added to the
CollectionProxy.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -6,6 +6,7 @@ module PaperTrail
   module Model
     def self.included(base)
       base.send :extend, ClassMethods
+      base.send :attr_accessor, :paper_trail_habtm
     end
 
     # :nodoc:
@@ -46,6 +47,12 @@ module PaperTrail
       #   the instance was reified from. Default is `:version`.
       # - :save_changes - Whether or not to save changes to the object_changes
       #   column if it exists. Default is true
+      # - :join_tables - If the model has a has_and_belongs_to_many relation
+      #   with an unpapertrailed model, passing the name of the association to
+      #   the join_tables option will paper trail the join table but not save
+      #   the other model, allowing reification of the association but with the
+      #   other models latest state (if the other model is paper trailed, this
+      #   option does nothing)
       #
       def has_paper_trail(options = {})
         options[:on] ||= [:create, :update, :destroy]
@@ -57,6 +64,42 @@ module PaperTrail
         setup_model_for_paper_trail(options)
 
         setup_callbacks_from_options options[:on]
+
+        setup_callbacks_for_habtm options[:join_tables]
+      end
+
+      def update_for_callback(name, callback, model, assoc)
+        model.paper_trail_habtm ||= {}
+        model.paper_trail_habtm.reverse_merge!(name => { removed: [], added: [] })
+        case callback
+        when :before_add
+          model.paper_trail_habtm[name][:added] |= [assoc.id]
+          model.paper_trail_habtm[name][:removed] -= [assoc.id]
+        when :before_remove
+          model.paper_trail_habtm[name][:removed] |= [assoc.id]
+          model.paper_trail_habtm[name][:added] -= [assoc.id]
+        end
+      end
+
+      attr_reader :paper_trail_save_join_tables
+
+      def setup_callbacks_for_habtm(join_tables)
+        @paper_trail_save_join_tables = Array.wrap(join_tables)
+        # Adds callbacks to record changes to habtm associations such that on
+        # save the previous version of the association (if changed) can be
+        # interpreted
+        reflect_on_all_associations(:has_and_belongs_to_many).
+          reject { |a| paper_trail_options[:skip].include?(a.name.to_s) }.
+          each do |a|
+            added_callback = lambda do |*args|
+              update_for_callback(a.name, :before_add, args[-2], args.last)
+            end
+            removed_callback = lambda do |*args|
+              update_for_callback(a.name, :before_remove, args[-2], args.last)
+            end
+            send(:"before_add_for_#{a.name}").send(:<<, added_callback)
+            send(:"before_remove_for_#{a.name}").send(:<<, removed_callback)
+          end
       end
 
       def setup_model_for_paper_trail(options = {})
@@ -442,6 +485,11 @@ module PaperTrail
       # Saves associations if the join table for `VersionAssociation` exists.
       def save_associations(version)
         return unless PaperTrail.config.track_associations?
+        save_associations_belongs_to(version)
+        save_associations_has_and_belongs_to_many(version)
+      end
+
+      def save_associations_belongs_to(version)
         self.class.reflect_on_all_associations(:belongs_to).each do |assoc|
           assoc_version_args = {
             version_id: version.id,
@@ -459,6 +507,27 @@ module PaperTrail
 
           if assoc_version_args.key?(:foreign_key_id)
             PaperTrail::VersionAssociation.create(assoc_version_args)
+          end
+        end
+      end
+
+      def save_associations_has_and_belongs_to_many(version)
+        # Use the :added and :removed keys to extrapolate the HABTM associations
+        # to before any changes were made
+        self.class.reflect_on_all_associations(:has_and_belongs_to_many).each do |a|
+          next unless
+            self.class.paper_trail_save_join_tables.include?(a.name) ||
+                a.klass.paper_trail_enabled_for_model?
+          assoc_version_args = {
+            version_id: version.id,
+            foreign_key_name: a.name
+          }
+          assoc_ids =
+            send(a.name).to_a.map(&:id) +
+            (@paper_trail_habtm.try(:[], a.name).try(:[], :removed) || []) -
+            (@paper_trail_habtm.try(:[], a.name).try(:[], :added) || [])
+          assoc_ids.each do |id|
+            PaperTrail::VersionAssociation.create(assoc_version_args.merge(foreign_key_id: id))
           end
         end
       end
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -14,6 +14,7 @@ module PaperTrail
           has_one: false,
           has_many: false,
           belongs_to: false,
+          has_and_belongs_to_many: false,
           unversioned_attributes: :nil
         )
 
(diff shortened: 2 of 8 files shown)


lib/paper_trail/reifier.rb
```diff
@@ -60,10 +61,14 @@ def reify(version, options)
...
 
       # Set all the attributes in this version on the model.
-      def reify_attributes(model, version, attrs)
+      def reify_attributes(model, version, attrs, options)
         enums = model.class.respond_to?(:defined_enums) ? model.class.defined_enums : {}
         model.class.unserialize_attributes_for_paper_trail! attrs
 
+        # Reifies habtm if the option is passed, otherwise sets the habtm
+        # attributes to nil so they are not reified
+        reify_has_and_belongs_to_many(model, attrs, options)
```
jaredbeck: All of the other association types (has_one, has_many, and belongs_to) happen in `reify_associations`. Why should HABTM happen here, in `reify_attributes`? So that it can "[set] .. the attributes to nil so they are not reified"? Can you explain that further, please?

lib/paper_trail/reifier.rb
```diff
@@ -83,6 +88,23 @@ def reify_attributes(model, version, attrs)
...
+        model.class.reflect_on_all_associations(:has_and_belongs_to_many).
+          each { |a|
+            # Don't let habtm associations be reified as attributes
+            ids = attrs.delete("#{a.name.to_s.singularize}_ids")
+            next unless options[:has_and_belongs_to_many]
+            # Update by using `id` IN (...) so if an object has been destroyed
+            # no errors are thrown, although the models association can not be
+            # fully reified (follows standard for has_many & has_many through)
+            model.send(a.name).proxy_association.target =
+              ids.present? ? a.name.to_s.singularize.classify.constantize.where(id: ids).to_a : []
```
jaredbeck: Won't this load the live (current) records? Shouldn't reification query the `versions` table? What am I missing?

jaredbeck:
Thanks for taking a stab at this @samboylett (Sam?). Rails has 6 association types:

```
    belongs_to
    has_one
    has_many
    has_many :through
    has_one :through
    has_and_belongs_to_many
```

Reification currently supports four or five of these, right? So adding HABTM support would be great!

@theRealNG (Ng?) recently added belongs_to support, in https://github.com/airblade/paper_trail/pull/730 so let's see if we can get him to look at this.

has_many reification was added by @bli, @lyfeyaj, and @NullVoxPopuli in https://github.com/airblade/paper_trail/pull/439 so maybe they can help too.

mikecmpbll:
in our case we wanted to reify the associations and not the associated objects. before this patch there wasn't any way to achieve this. it was implemented as attributes because with HABTM there's no join model to version.

jaredbeck:
>> Won't this load the live (current) records?
>
> in our case we wanted to reify the associations and not the associated objects.

All of the existing assoc. reification methods restore the associated objects. I think it will be misleading if HABTM works differently from all the rest.

mikecmpbll:
fair enough, maybe it needs renaming or something then. the lack of this feature is really limiting though. we found the same with the current way the HMT association works, we didn't find it possible to restore the has_many record, and not the has_many through target.

lib/paper_trail/reifier.rb
samboylett: As the join model can't be versioned, the opposing habtm models ID is stored instead as an attribute in the serialised object. So `has_and_belongs_to_many :bars` would be saved as `bar_ids = [1, 2, 3, etc.]` in the serialised object column of the versions table. This means that upon reification of attributes, `reify_attributes` would attempt to set `model.bar_ids = [1, 2, 3]` even if the option to reify habtm is set to false. So these keys must be deleted from the attrs hash (they are deleted, not set to nil, my mistake)

So if the habtm option is not passed, the model retains the habtm associations from the latest version.

samboylett:
I have added in a query to the versions table and some appropriate tests so HABTM should work as other association reifications (also added `has_and_belongs_to_many: false` to the `options.merge` for other reifiers)

jaredbeck:
**tl;dr** Tests look good, but I still have concerns about implementation, specifically persistence.

> .. HABTM should [now] work as other association reifications (also added has_and_belongs_to_many: false to the options.merge for other reifiers)

Great! I've read through the tests as of c0f15f2ccd7253389b18301571047dd513045fe7. They demonstrate the behavior I'd expect and seem thorough. Given these tests, I'm pretty sure we'll be able to add this feature, it's just a question of making sure we have the right implementation.

It's important that we find the right implementation here, especially regarding persistence, because it'll be very difficult to change the format once people start recording data in production.

If I understand correctly, this implementation stores a list of ids in the `versions.object` column. However, that column was only designed to persist attributes, not associations. Using it to store associations would also be inconsistent with other association types, which use the `version_associations` table.

```ruby
create_table :version_associations do |t|
  t.integer  :version_id
  t.string   :foreign_key_name, null: false
  t.integer  :foreign_key_id
end
```
 
In a Has Many Through (HMT) association, I think we insert one record into `version_associations` for each record in the "through table", and, in turn, for each record in the "target table". (to use the terminology from `Reifier#reify_has_many_through`)

Couldn't we still use this table for a HABTM association?

```ruby
bar1 = BarHabtm.create
bar2 = BarHabtm.create
FooHabtm.create(bar_habtms: [bar1])
FooHabtm.update_attributes(bar_habtms: [bar2]) # line 4
```

On line 4, when the update event is recorded in the `versions` table, we could insert a record into the `versions_associations` table, with a `foreign_key_name` of `bar_habtms` and a `foreign_key_id` equal to `bar1.id`.

It seems that in this way we could use the `version_associations` for all association types. This consistency is desirable.

samboylett:
Hi, I have changed the implementation as per your suggestion, also added the option to save only the join table, such that the associated objects are shown in their current state if they are not paper trailed.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -57,6 +57,41 @@ def has_paper_trail(options = {})
         setup_model_for_paper_trail(options)
 
         setup_callbacks_from_options options[:on]
+
+        setup_callbacks_for_habtm options[:join_tables]
```
jaredbeck: Please add this new option (join_tables) to the method documentation above.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -57,6 +57,41 @@ def has_paper_trail(options = {})
...
+      end
+
+      def update_for_callback(name, callback, model, assoc)
+        paper_trail_habtm = model.instance_variable_get(:@paper_trail_habtm) || {}
```
jaredbeck: Is there any way we can avoid using `instance_variable_get`? Could we, for example, add an instance accessor method?

jaredbeck:
Please add a changelog entry.

mikecmpbll:
Commit updated 👍🏼

jaredbeck:
Looks good to me. @batter any concerns?

CC: The people who have worked on associations: @theRealNG, @bli, @lyfeyaj, and @NullVoxPopuli your feedback still very much welcome.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -57,6 +64,42 @@ def has_paper_trail(options = {})
         setup_model_for_paper_trail(options)
 
         setup_callbacks_from_options options[:on]
+
+        setup_callbacks_for_habtm options[:join_tables]
+      end
+
+      def update_for_callback(name, callback, model, assoc)
+        model.paper_trail_habtm ||= {}
+        model.paper_trail_habtm.reverse_merge!(name => { removed: [], added: [] })
```
tlynam: Why not do this? It seems easier to understand

```ruby
model.paper_trail_habtm[name] ||= {removed: [], added: []}

File: lib/paper_trail/has_paper_trail.rb [hunk 6]
```diff
@@ -463,6 +511,27 @@ def save_associations(version)
...
+        self.class.reflect_on_all_associations(:has_and_belongs_to_many).each do |a|
+          next unless
+            self.class.paper_trail_save_join_tables.include?(a.name) ||
+                a.klass.paper_trail_enabled_for_model?
+          assoc_version_args = {
+            version_id: version.id,
+            foreign_key_name: a.name
+          }
+          assoc_ids =
+            send(a.name).to_a.map(&:id) +
```
tlynam: `to_a` is unnecessary as map is implemented on ActiveRecord Relation

lib/paper_trail/has_paper_trail.rb
```diff
@@ -463,6 +511,27 @@ def save_associations(version)
...
+            (@paper_trail_habtm.try(:[], a.name).try(:[], :removed) || []) -
+            (@paper_trail_habtm.try(:[], a.name).try(:[], :added) || [])
```
tlynam: What do you think of this?
```ruby
history = @paper_trail_habtm || {}
history = history[a.name] || Hash.new([])

assoc_ids = send(a.name).map(&:id) + history[:removed] - history[:added]
```

lib/paper_trail/has_paper_trail.rb
```diff
@@ -46,6 +47,12 @@ module ClassMethods
       #   the instance was reified from. Default is `:version`.
       # - :save_changes - Whether or not to save changes to the object_changes
       #   column if it exists. Default is true
+      # - :join_tables - If the model has a has_and_belongs_to_many relation
+      #   with an unpapertrailed model, passing the name of the association to
+      #   the join_tables option will paper trail the join table but not save
+      #   the other model, allowing reification of the association but with the
+      #   other models latest state (if the other model is paper trailed, this
+      #   option does nothing)
```
tlynam: > (if the other model is paper trailed, this option does nothing)

Are you sure that's the case? I just read through the code and don't see where that would happen. It seems like if one side of the association has `join_tables` set, and the other side of the association has paper_trail enabled without `join_tables`, this code will correctly load the versions for both sides of the association.

Perhaps you meant to say that there's no point enabling `join_tables` on both models?

tlynam: Oh I must have misunderstood -- HABTM models are tracked automatically, and this option exists to disable that.

I'm curious why someone would want to do that, though.

samboylett: Hi, if both models are paper trailed then the join table is also paper trailed, allowing a complete history of the models and associations. If only one model is paper trailed, and the other is not, then by default nothing is saved in relation to both models. However, by passing the `join_tables` option, the association changes are saved. You would only ever need `join_tables` when you want to paper trail an association but not the associated model. If you take a look at `save_associations_has_and_belongs_to_many` you can see the logic behind the saving.

jaredbeck:
It sounds like we're all generally happy with this, so I'm going to merge it.

@tlynam and @theRealNG Thanks for the review. Todd, good suggestions, feel like making a PR?

@samboylett and @mikecmpbll thanks for the contribution!

PR #773, opened 2016-04-09
Explicit requires

jaredbeck:
Explicit requires have three advantages:

1. Allow us to add files without them being automatically required. This may be useful for e.g. cleaner.rb
2. Enable static analysis in IDEs
3. Shave a few microseconds off of load time by eliminating a few calls to the filesystem to list directories.


jaredbeck:
@batter any concerns with this?

Issue #555, opened 2015-06-25
Switch the default serializer to JSON?


jaredbeck:
Ben, this is just a placeholder for a future discussion, starting where we left off in https://github.com/airblade/paper_trail/pull/551

> I will say that I have been wondering for a while whether it makes sense to switch the default serializer to `PaperTrail::Serializers::JSON` (instead of `PaperTrail::Serializers::YAML`), because JSON syntax is more consistent in it's parsing (not as much of an issue anymore, but when it was unclear whether you'd get `syck` or `psych` depending on which Ruby version you were using for a while, it did cause some compatibility issues), and also dumps into shorter and more concise syntax than YAML (it's also faster as far as I know).  This would probably prevent some of these column limit size issues that people like @brodock were running up against. That being said, it would also be a breaking change for anyone who didn't read the documentation and realize they needed to manually swap in the old YAML serializer if they have legacy data (or update all their old data to JSON... perhaps we could provide a script / mechanism).
> 
> There is some appropriate / notable discussion about the pros / cons of JSON vs YAML in #194 & a bit of commentary on #504. There was also a user submitted script for updating legacy YAML data to JSON that we might be able to leverage in #388..  Since it's a non backwards compatible change, if we do decide to go this route, it would probably need to be part of a major version release.

Let's pick up this discussion again in a few months, after we get 4.0 out the door.

seanlinsley:
This is how I'd imagine you could provide backwards compatibility.
```ruby

def serializer
  @serializer ||= determine_serializer
end

def determine_serializer
  record = PaperTrail::Version.where.not(object: [nil, '', '{}']).first
  return :json unless record
  object = record.attributes_before_type_cast['object']

  begin
    JSON.parse object
    return :json
  rescue JSON::ParserError
  end

  begin
    YAML.load object
    return :yaml
  rescue YAML::SyntaxError
  end

  fail <<-ERROR.strip_heredoc
    Could not determine whether your existing version history was encoded in JSON or YAML.
    Please check the data in this record with:
    `PaperTrail::Version.find(#{record.id}).attributes_before_type_cast['object']`
  ERROR
end
```

If later on you wanted to force people to specify YAML in their config file, you could change this part:
```ruby
  begin
    YAML.load object
    fail "paper_trail 5.0 requires that you specify YAML as your serializer (JSON is now the default)"
  rescue YAML::SyntaxError
  end
```

jaredbeck:
[...]

PR #774, opened 2016-04-11
Require user to specify serializer

jaredbeck:
Per the discussion in https://github.com/airblade/paper_trail/issues/555, if we *were* to change the default serializer (which has not been decided yet) I'd prefer to add a deprecation warning like this in one major version (currently, the next is 5) and then change the default in the version after that (e.g. 6).

@seanlinsley @batter

### DIFF
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -1,4 +1,5 @@
 require "singleton"
+require "active_support/core_ext"
 require "paper_trail/serializers/yaml"
 
 module PaperTrail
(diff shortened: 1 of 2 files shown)


jaredbeck:
To clarify, please do not merge this, it's just an aid to the discussion in #555. :no_good: :no_entry_sign:

seanlinsley:
I think it'd be preferable to only print this warning if YAML is actually in use. Are there downsides to detecting the current format like I described on #555?

jaredbeck:
Closing. This was just for the sake of discussion. (Which I guess should continue in #555)  Sean, I will get back to you about the pros/cons of format detection. I have some thoughts on it, I just have to organize them :)

PR #778, opened 2016-04-18
fixes #777. Also modified test (changed authorship.person to authorsh…

### LINKED ISSUE #777: reify_has_many_through doesn't consider :source option
If a has_many association uses both the :through and the :source option, reify restores completely arbitrary associations.

I isolated this problem here: https://gist.github.com/Mask/29e2e3d286183277d2f391aafc997025#file-bug_report_template-rb-L71

Notice that Alice starts with one friend (Bob), then gains another friend (so we now have Bob and Charlie in the `friends` collection). After reify, the `friends` collection isn't reverted back to only Bob, but suddenly contains Alice.
After tracing through the code, I think I have found the culprit:

https://github.com/airblade/paper_trail/blob/7733894b0d13c2a2a172ff0a8a6b1509eb02011c/lib/paper_trail/reifier.rb#L258

It calls `association_foreign_key` to find out how to follow the association. If we look at the implementation of `association_foreign_key`, we find:

https://github.com/rails/rails/blob/ac027338e4a165273607dccee49a3d38bc836794/activerecord/lib/active_record/reflection.rb#L388-L390

```
def association_foreign_key
  @association_foreign_key ||= options[:association_foreign_key] || class_name.foreign_key
end
```
Well, `options[:association_foreign_key]` will always we nil, since it can only be specified for HABTM associations, not has_many :through. So this method always falls back to `class_name.foreign_key`. The result is that paper_trail's reify_has_many_through only follows the correct association if its name matches the class it's pointing to. If my example, it returns `user_id`(since to points to the `User` model) instead of `to_id` (as specified in the `:source` option). 

I'll research this a bit more and see if I can come up with a PR to fix this.

Mask:
…ip.author so that source is not the same as the class name) to verify this case.

### DIFF
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -255,7 +255,7 @@ module PaperTrail
             }
           else
             collection_keys = through_collection.map { |through_model|
-              through_model.send(assoc.association_foreign_key)
+              through_model.send(assoc.source_reflection.foreign_key)
             }
 
             version_id_subquery = assoc.klass.paper_trail_version_class.
--- a/test/dummy/app/models/authorship.rb
+++ b/test/dummy/app/models/authorship.rb
@@ -1,5 +1,5 @@
 class Authorship < ActiveRecord::Base
   belongs_to :book
-  belongs_to :person
+  belongs_to :author, class_name: "Person"
   has_paper_trail
 end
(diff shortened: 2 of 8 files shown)


jaredbeck:
Looks good! Two small things:

1. Please update `test/dummy/db/schema.rb` (e.g. `authorships.person_id`)
2. Please squash your commits

Thanks!

Mask:
no problem - how's this?

jaredbeck:
Closed by https://github.com/airblade/paper_trail/pull/779, preserving your authorship. Thanks Martin!

PR #784, opened 2016-04-26
Note about PaperTrail::Version as an AR instance

tnm:
I've noticed a bit of confusion about `PaperTrail::Version` — e.g, folks think it's more magical than it really is. Noting that it's really just a good old subclass of `AR:Base` might help.


jaredbeck:
Makes sense, thanks Ted.

PR #794, opened 2016-05-04
Fix bug in AR integration

### LINKED ISSUE #789: How to solve has_paper_trail method missing in rails 3.2
**This issues is how to solve a problem, please create a pull request to fix it or update the documentation.**

Using rails 3.2 and paper_trail 5.0.0 I receive the following error:
``` bash
.../activerecord-3.2.19/lib/active_record/dynamic_matchers.rb:55:in `method_missing': undefined local variable or method `has_paper_trail' for #<Class:0x007f339a7baaf0> (NameError)
```

To fix it, create an initiailizer and load the active record framework:

``` bash
$ echo "require 'paper_trail/frameworks/active_record'" > config/initializers/paper_trail.rb
``` 

The problem was introducted by https://github.com/airblade/paper_trail/commit/2029ea37ac8766ed3d6141b4b48476fd7ab286b7 in https://github.com/airblade/paper_trail/pull/725

jaredbeck:
[Fixes #789]

### DIFF
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -176,6 +176,10 @@ unless PaperTrail.active_record_protected_attributes?
   end
 end
 
+ActiveSupport.on_load(:active_record) do
+  include PaperTrail::Model
+end
+
 # Require frameworks
 require "paper_trail/frameworks/sinatra"
 if defined?(::Rails) && ActiveRecord::VERSION::STRING >= "3.2"
--- a/lib/paper_trail/frameworks/active_record.rb
+++ b/lib/paper_trail/frameworks/active_record.rb
@@ -2,7 +2,3 @@
 # since otherwise the model(s) will get loaded in via the `Rails::Engine`.
 require "paper_trail/frameworks/active_record/models/paper_trail/version_association"
 require "paper_trail/frameworks/active_record/models/paper_trail/version"
-
-ActiveSupport.on_load(:active_record) do
-  include PaperTrail::Model
-end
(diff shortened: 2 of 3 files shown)


jaredbeck:
Reverts the part of https://github.com/airblade/paper_trail/commit/2029ea37ac8766ed3d6141b4b48476fd7ab286b7 which appears to cause issue https://github.com/airblade/paper_trail/issues/789

I would prefer if users were required to manually `include PaperTrail::Model` in each of their models, but that's a whole 'nother discussion..

jaredbeck:
`on_load` is explained in `activesupport-4.2.6/lib/active_support/lazy_load_hooks.rb` as follows:

```
# lazy_load_hooks allows Rails to lazily load a lot of components and thus
# making the app boot faster. Because of this feature now there is no need to
# require <tt>ActiveRecord::Base</tt> at boot time purely to apply
# configuration. Instead a hook is registered that applies configuration once
# <tt>ActiveRecord::Base</tt> is loaded. Here <tt>ActiveRecord::Base</tt> is
# used as example but this feature can be applied elsewhere too.
#
# Here is an example where +on_load+ method is called to register a hook.
#
#   initializer 'active_record.initialize_timezone' do
#     ActiveSupport.on_load(:active_record) do
#       self.time_zone_aware_attributes = true
#       self.default_timezone = :utc
#     end
#   end
#
# When the entirety of +activerecord/lib/active_record/base.rb+ has been
# evaluated then +run_load_hooks+ is invoked. The very last line of
# +activerecord/lib/active_record/base.rb+ is:
#
#   ActiveSupport.run_load_hooks(:active_record, ActiveRecord::Base)
```

batter:
> I would prefer if users were required to manually include PaperTrail::Model in each of their models, but that's a whole 'nother discussion..

Agreed that would be ideal, as it does seem a bit excessive including it on every `ActiveRecord` instance in a project where the gem is being used. Maybe we can shoot for that for PaperTrail 6?

PR #799, opened 2016-05-09
Update documentation to note object type returned by `versions`

devonestes:
It was unclear in the documentation that overriding the `versions` method would
result in changed behavior not just when retrieving versions but also in
creating them. This commit adds some documentation so the user is aware of this
behavior.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -8,6 +8,6 @@ Please use our template
 when reporting bugs.
 
 For other questions, please see our contributing guide.
-(https://github.com/airblade/paper_trail/blob/master/.github/CONTRIBUTING.MD)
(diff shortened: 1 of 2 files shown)


jaredbeck:
```diff
  # Returns this widget's versions.  You can customise the name of the
-# association.
+# association. This method returns an ActiveRecord::Relation object,
+# so it is important to note that overriding this method will result in
+# changes both to how versions are retrieved AND how they are created.
```

I wonder if people might get confused between "customise" and "override". What do you think about something simpler, like:

```diff
  # Returns this widget's versions.  You can customise the name of the
-# association.
+# association, but overriding it is not supported.
```

This is close to the top of the docs, where we're still covering basics, and I don't want to get into the details of why overriding it is a bad idea. If you want to cover that in detail, let's do it lower in the file, under the "ActiveRecord" section, near "Single Table Inheritance".

devonestes:
Good point - I'll re-word it.

PR #800, opened 2016-05-09
Update documentation to note object type returned by `versions`

devonestes:
It was unclear in the documentation that overriding the `versions` method would
result in changed behavior not just when retrieving versions but also in
creating them. This commit adds some documentation so the user is aware of this
behavior.


jaredbeck:
Seems reasonable, thanks Devon.

Issue #807, opened 2016-05-17
Errors are not logged when versions can't be created


md5:
I ran into two issues in the last day that resulted in versions not being persisted.

The first was caused by an incompatibility between PaperTrail and `uuid` primary keys in Postgres (cf. https://github.com/airblade/paper_trail/issues/363), which caused no versions to be persisted. The second was caused by the fact that `belongs_to` defaults to `required: true` on new Rails 5 projects, which caused `destroy` versions to not be persisted (this is fixed in PaperTrail 5.0, cf. https://github.com/airblade/paper_trail/issues/682, https://github.com/airblade/paper_trail/issues/683, https://github.com/airblade/paper_trail/issues/684).

It seems like it would be useful to log an error when a `Version` object cannot be persisted. I did some initial POC work on logging errors in my fork and it was very useful in helping me diagnose the second issue (after I banged my head against the first issue for a couple hours yesterday): https://github.com/md5/paper_trail/commit/01f522af9c335028c25f32b18b4e483a289308c1

md5:
Also, since the `record_destroy` issue is only fixed in PaperTrail 5.0, it seems like the 4.1 series should not be marked as being compatible with AR < 6 unless that fix is going to be backported.

jaredbeck:
Great suggestions, Mike!

> .. since the record_destroy issue is only fixed in PaperTrail 5.0, it seems like the 4.1 series should not be marked as being compatible with AR < 6 unless that fix is going to be backported.

Please submit a backport PR to the `4.1-stable` branch, thanks!

> It seems like it would be useful to log an error when a Version object cannot be persisted. I did some initial POC work on logging errors in my fork and it was very useful in helping me diagnose the second issue (after I banged my head against the first issue for a couple hours yesterday): md5@01f522a

01f522af9c3 looks good to me, at first glance. Please submit a PR to the `master` branch, thanks!

md5:
@jaredbeck Backport PR is at https://github.com/airblade/paper_trail/pull/808

As for the logging fix, are you guys OK with logging `self.inspect`? My only concern is that this could leak sensitive information into the logs. If you have some examples of how model identities should be logged, I'd be happy to change this aspect before creating a PR.

jaredbeck:
> As for the logging fix, are you guys OK with logging self.inspect? My only concern is that this could leak sensitive information into the logs. 

Good thinking.

> If you have some examples of how model identities should be logged, I'd be happy to change this aspect before creating a PR.

[...]

PR #809, opened 2016-05-18
Log errors when versions can't be created

md5:
As discussed in #807, this PR updates `record_update` and `record_destroy` to log a warning listing all validation errors when a `Version` instance cannot be saved. It does not update `record_create` since that calls `Version.create!` instead of `Version.create` and should raise an error for any validation failures.

Due to Rubocop complaining, I have also slightly refactored `has_paper_trail.rb` to fix some of the complaints. There is still a complaint about the `InstanceMethods` module being too many lines, but I didn't see any easy way to reduce the line count of these changes without going too far off topic and into the weeds.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -375,9 +375,7 @@ module PaperTrail
           if pt_record_object_changes? && changed_notably?
             data[:object_changes] = pt_recordable_object_changes
           end
-          if self.class.paper_trail_version_class.column_names.include?("transaction_id")
-            data[:transaction_id] = PaperTrail.transaction_id
-          end
+          add_transaction_id_to(data)
           version = send(self.class.versions_association_name).create! merge_metadata(data)
           update_transaction_id(version)
           save_associations(version)
@@ -397,9 +395,7 @@ module PaperTrail
           if pt_record_object_changes?
             data[:object_changes] = pt_recordable_object_changes
           end
-          if self.class.paper_trail_version_class.column_names.include?("transaction_id")
-            data[:transaction_id] = PaperTrail.transaction_id
-          end
+          add_transaction_id_to(data)
           version = send(self.class.versions_association_name).create merge_metadata(data)
           update_transaction_id(version)
           save_associations(version)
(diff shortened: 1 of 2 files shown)


jaredbeck:
> There is still a complaint about the InstanceMethods module being too many lines ..

Just bump up the number in rubocop_todo.yml .. it's a fight for another day :)

lib/paper_trail/has_paper_trail.rb
```diff
@@ -397,12 +395,12 @@ def record_update(force = nil)
...
-            data[:transaction_id] = PaperTrail.transaction_id
-          end
+          add_transaction_id(data)
           version = send(self.class.versions_association_name).create merge_metadata(data)
-          update_transaction_id(version)
-          save_associations(version)
+          unless log_version_errors(version, :update)
+            update_transaction_id(version)
+            save_associations(version)
+          end
```
jaredbeck: I find this conditional hard to read. I'd prefer:

```ruby
if version.errors.any?
  log_version_errors(version, :update)
else
  update_transaction_id(version)
  # etc.
```

If rubocop complains about method complexity, just bump up the numbers; it's a fight for another day.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -638,6 +636,11 @@ def save_version?
         (if_condition.blank? || if_condition.call(self)) && !unless_condition.try(:call, self)
       end
 
+      def add_transaction_id(data)
+        return unless self.class.paper_trail_version_class.column_names.include?("transaction_id")
+        data[:transaction_id] = PaperTrail.transaction_id
+      end
```
jaredbeck: Nice method extraction. How about `add_transaction_id_to` as a method name? (note the "to"). Also, if it's easy to do this method extraction in a separate commit, that would be great. If not, don't worry about it.

lib/paper_trail/has_paper_trail.rb
md5: That's what I had originally until RuboCop complained about `AbcSize`. I've changed it back, though subsequent edits seem to have removed the need to bump `AbcSize`.

lib/paper_trail/has_paper_trail.rb
md5: Done in 71b93758bf191560b2fab0ea96b50a231c26007d

md5:
I've refactored and bumped the `Metrics/ModuleLength` limit in `.rubycop_todo.yml` 👍

jaredbeck:
Thanks Mike!

md5:
:metal:

PR #810, opened 2016-05-18
Break AttributesSerialization into small modules

jaredbeck:
It already was organized *in memory* but now it's organized in files.

Also, there are six fewer methods being mixed into the model class
(they became ObjectAttribute and ObjectChangesAttribute).

I like how the legacy (rails < 4.2) support is now all in one file,
so it'll be really easy to drop when the time comes.

Finally, a minor thing, I renamed CastedAttributeSerializer to
CastAttributeSerializer, because I don't think "casted" is the
correct conjugation.


jaredbeck:
@batter Ben, what do you think? I'm really happy with it, but I just wrote it so I may be biased :)

@kaspth Kasper, any objections? You wrote CastedAttributeSerializer, IIRC.

Note that this is basically the same implementation as before, just organized a bit into separate files.

kaspth:
No objections from me, if you like this organization better then 👍

PR #812, opened 2016-05-18
Fix for HABTM associations when updated within 1 transaction

samboylett:
When updating 2 models with a HABTM association, and using
accepts_nested_attributes_for, the association was using the
version ID instead of the transaction ID to reference the change.
This was braking the association logic and preventing proper
reification.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -526,7 +526,7 @@ module PaperTrail
             self.class.paper_trail_save_join_tables.include?(a.name) ||
                 a.klass.paper_trail_enabled_for_model?
           assoc_version_args = {
-            version_id: version.id,
+            version_id: version.transaction_id,
             foreign_key_name: a.name
           }
           assoc_ids =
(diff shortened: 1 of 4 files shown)


lib/paper_trail/has_paper_trail.rb
```diff
@@ -526,7 +526,7 @@ def save_associations_has_and_belongs_to_many(version)
             self.class.paper_trail_save_join_tables.include?(a.name) ||
                 a.klass.paper_trail_enabled_for_model?
           assoc_version_args = {
-            version_id: version.id,
+            version_id: version.transaction_id,
```
jaredbeck: Remind me again why we store `transaction_id` in the `version_id` column? Is it because there is no "through" table in a HABTM?

jaredbeck:
This looks good, thanks Sam. Please add a changelog entry.

jaredbeck:
Thanks, Sam! Merged offline as 4ad6254b742c535a133687d03ec74abbf44b6a1f with tiny fix to changelog.

PR #813, opened 2016-05-18
Fix deprecation warning for paper_trail_on_destroy(:after)

md5:
* Warn for either a `"after"` or `:after`
* Correctly warn on beta versions of ActiveRecord 5

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -174,8 +174,8 @@ module PaperTrail
           raise ArgumentError, 'recording order can only be "after" or "before"'
         end
 
-        if recording_order == "after" &&
-            Gem::Version.new(ActiveRecord::VERSION::STRING) >= Gem::Version.new("5")
+        if recording_order.to_s == "after" &&
+            Gem::Version.new(ActiveRecord::VERSION::STRING).release >= Gem::Version.new("5")
           if ::ActiveRecord::Base.belongs_to_required_by_default
             ::ActiveSupport::Deprecation.warn(
               "paper_trail_on_destroy(:after) is incompatible with ActiveRecord " +


md5:
I added a spec to ensure that the deprecation warning is happening and it's not passing for some reason. In the process, I discovered that `Gem::Version#<=>` does not consider beta versions of AR 5 to be `>= Gem::Version.new("5")` (which makes sense).

spec/models/callback_modifier_spec.rb
```diff
@@ -23,6 +23,19 @@
...
             expect(modifier.versions.last.reify).to be_flagged_deleted
           end
+
+          if active_record_gem_version.release >= Gem::Version.new("5")
+            context "when ActiveRecord::Base.belongs_to_required_by_default is enabled" do
+              it "should display a deprecation warning" do
+                allow(::ActiveRecord::Base).to receive(:belongs_to_required_by_default).
+                  and_return(true)
+                expect(::ActiveSupport::Deprecation).to receive(:warn).
+                  with(/paper_trail_on_destroy(:after) is incompatible with ActiveRecord/)
```
jaredbeck: Mike, I think your test doesn't pass because your regex doesn't match the actual output.

```ruby
pattern = /paper_trail_on_destroy(:after) is incompatible with ActiveRecord/
output = "paper_trail_on_destroy(:after) is incompatible with ActiveRecord"
pattern.match(output) # => nil
```

jaredbeck: You can run the AR5 tests locally with `bundle exec appraisal ar5 rake`.

md5: Good catch on the parenthesis thing! I should have noticed that previous and I've fixed in an amended commit: https://github.com/airblade/paper_trail/pull/813/commits/0bf252eb4c22af36f111b0f8812d96b973c0ab2d

However, that doesn't seem to be the actual issue. If that were the problem, I would have seen the method called 1 time with a mismatched argument, but instead I'm seeing it called 0 times...

md5: I think the real issue is that `paper_trail_on_destroy` is a class method that is called as soon as the `AfterDestroyModifier` class is loaded, so the warning is happening before the example is run in this spec.

jaredbeck:
> I think the real issue is that paper_trail_on_destroy is a class method that is called as soon as the AfterDestroyModifier class is loaded, so the warning is happening before the example is run in this spec.

Makes sense. We can live without the spec, thanks.

md5:
@jaredbeck Dropped the useless spec 👍

jaredbeck:
Thanks Mike!

md5:
👍

PR #826, opened 2016-06-15
Adding guard clause to prevent calling user_for_paper_trail

### LINKED ISSUE #817: Don't call current_user when paper_trail is not enabled for the controller
Hi there!

I've been adding paper trail to an app, and got surprised that after adding papertrail, some tests started failing even after adding the rspec helper

Turns out I had some tests stubbing `current_user` and I was surprised to see paper trail was attempting to call it even when disabled, since the method can trigger queries and other things.

Looking at this https://github.com/airblade/paper_trail/blob/master/lib/paper_trail/frameworks/rails/controller.rb#L94 it's pretty easy to see the cause. I'd be happy to provide a PR with a simple guard on the `enabled_for_controller` call, if you guys think it's something worth changing.

IMO paper trail before actions should be a complete no-op when disabled.

Draiken:
When PaperTrail is disabled, it should not call `user_for_paper_trail` on the controller

Closes #817

### DIFF
--- a/lib/paper_trail/frameworks/rails/controller.rb
+++ b/lib/paper_trail/frameworks/rails/controller.rb
@@ -90,10 +90,11 @@ module PaperTrail
       end
 
       def warn_about_not_setting_whodunnit
-        enabled = ::PaperTrail.enabled_for_controller?
+        return unless ::PaperTrail.enabled_for_controller?
+
         user_present = user_for_paper_trail.present?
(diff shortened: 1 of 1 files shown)


Draiken:
I wasn't sure where to test this, if you tell me where to put it, I can add some tests.
Only saw this tested on functional tests, and this is more of a unit test case in my view.

jaredbeck:
Thanks!

PR #829, opened 2016-06-22
Corrected API-Calls to Versioning

sreuterle:
Versioning methods are called directly on widget instead through widget.paper_trail

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,23 +170,23 @@ widget.version
 
 # Returns true if this widget is the current, live one; or false if it is from
 # a previous version.
-widget.paper_trail.live?
+widget.live?
 
 # Returns who put the widget into its current state.
-widget.paper_trail.originator
+widget.paper_trail_originator
 
 # Returns the widget (not a version) as it looked at the given timestamp.
-widget.paper_trail.version_at(timestamp)
+widget.version_at(timestamp)
 
 # Returns the widget (not a version) as it was most recently.
-widget.paper_trail.previous_version
+widget.previous_version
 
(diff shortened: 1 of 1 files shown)


jaredbeck:
Thanks Sven, but these changes would be incorrect. For example, `widget.live?` is actually deprecated in favor of `widget.paper_trail.live?`. See https://github.com/airblade/paper_trail/pull/719

sreuterle:
Oh, sorry for that. I've version 5.0.1 installed and there seems to be no `widget.paper_trail` (for corresponding object for `widget`). I try to update to latest version. Thanks you :)

jaredbeck:
> Oh, sorry for that. I've version 5.0.1 installed and there seems to be no widget.paper_trail (for corresponding object for widget). I try to update to latest version. Thanks you :)

`widget.paper_trail` has not been released yet. You are reading the documentation for the `master` branch. I'll see if I can clarify that a bit better at the top of the readme.

sreuterle:
Maybe it would be better to make a release branch until release so that master corresponds to released version on every time. That's how we handled branches in our company.
Then, of course, the readme would correspond to the released version, too.

jaredbeck:
> Maybe it would be better to make a release branch until release so that master corresponds to released version on every time.

Thanks for the suggestion, but there is a strong convention in the ruby OSS community that the `master` branch represents the very latest work, and tags are used for releases. Following the convention has value.

sreuterle:
Didn't knew that. Thanks for clarifying!

PR #846, opened 2016-08-14
Fix enum serialization in rails 4.2

### LINKED ISSUE #798: Bug with enum / enums
Hi,
I have a bug with `enums` and `paper_trail`. I will explain the bug with a dummy example.

```ruby
# We have this model `models/post.rb`
class Post < ActiveRecord::Base
  has_paper_trail
  enum kind: { document: 0, newspaper: 1 }
end
```

```ruby
# We create a post
@post = Post.create(title: "Hello world", kind: :document)

# We update the post for set the enum `kind` to `newspaper`
@post.update_attribute(kind: :newspaper)

# We print the versions we have (only for more informatión)
@post.versions
# return: 
[

    [0] <PaperTrail::Version:0x00558ca746a800> {
                    :id => 3,
             :item_type => "Post",
               :item_id => 916,
                 :event => "create",
             :whodunnit => "684",
                :object => "---\nid: 916\nuser_id: 739\ntitle: 'Hello world'\nkind: \n...",
            :created_at => Mon, 09 May 2016 16:15:55 UTC +00:00,
        :transaction_id => 3
    },
    [1] <PaperTrail::Version:0x00558ca746a6c0> {
                    :id => 4,
             :item_type => "Post",
               :item_id => 916,
                 :event => "update",
             :whodunnit => "684",
                :object => "---\nid: 916\nuser_id: 739\ntitle: 'Hello world'\nkind: \n...",
            :created_at => Mon, 09 May 2016 16:16:15 UTC +00:00,
        :transaction_id => 4
    }
]

# We check previous version
@old_version = @post.previous_version
# return:
<Post:0x00558ca743fad8> {
                    :id => 916,
               :user_id => 739
                 :title => 'Hello world',
                  :kind => nil, # this needs to be "newspaper"
            :created_at => Mon, 09 May 2016 16:15:55 UTC +00:00,
            :updated_at => Mon, 09 May 2016 16:16:05 UTC +00:00
}
```

**Some extra info**

- Ruby 2.2.5
- Rails 4.2.6
- PostgreSQL 9.5.2
- gem PaperTrail 5.0.1 with **Associations** (`--with-associations`)

jaredbeck:
Fixes #798

Our test suite has one model that tests enums, PostWithStatus. Its
spec did not cover `touch_with_version`. Somehow, probably during
our upgrade to rails 5, we broke `touch_with_version` for enums.
Unlike other methods, when an enum arrives at
`CastAttributeSerializer#serialize` during `touch_with_version`, it
is already a number. We assumed it was always a string.

This is sort of a hack, just handling both strings and numbers in
`#serialize`. I'd rather figure out how a number got to `#serialize`
in the first place. However, a hack is acceptable, as it's only
for rails 4.2, and the hack will eventually be deleted when we
drop support for 4.2.

### DIFF
--- a/lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
+++ b/lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
@@ -34,8 +34,13 @@ module PaperTrail
         end
 
         def serialize(attr, val)
-          val = defined_enums[attr][val] if defined_enums[attr]
-          @klass.type_for_attribute(attr).type_cast_for_database(val)
+          castable_val = val
+          if defined_enums[attr]
+            # `attr` is an enum. Find the number that corresponds to `val`. If `val` is
+            # a number already, there won't be a corresponding entry, just use `val`.
+            castable_val = defined_enums[attr][val] || val
+          end
+          @klass.type_for_attribute(attr).type_cast_for_database(castable_val)
         end
 
         def deserialize(attr, val)
@@ -49,6 +54,8 @@ module PaperTrail
 
         private
 
+        # ActiveRecord::Enum was added in AR 4.1
+        # http://edgeguides.rubyonrails.org/4_1_release_notes.html#active-record-enums
         def defined_enums
           @defined_enums ||= (@klass.respond_to?(:defined_enums) ? @klass.defined_enums : {})
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@
 
 ### Fixed
 
-- None
+- [#798](https://github.com/airblade/paper_trail/issues/798) -
+  Fix a rare bug with serialization of enums in rails 4.2 only when
+  using `touch_with_version`.
 
 ## 5.2.0 (2016-06-27)
 
--- a/spec/models/post_with_status_spec.rb
+++ b/spec/models/post_with_status_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
-# This model is in the test suite soley for the purpose of testing ActiveRecord::Enum,
-# which is available in ActiveRecord4+ only
+# This model tests ActiveRecord::Enum, which was added in AR 4.1
+# http://edgeguides.rubyonrails.org/4_1_release_notes.html#active-record-enums
 describe PostWithStatus, type: :model do
   if defined?(ActiveRecord::Enum)
     with_versioning do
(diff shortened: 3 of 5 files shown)


lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
```diff
@@ -34,8 +34,13 @@ def initialize(klass)
         end
 
         def serialize(attr, val)
-          val = defined_enums[attr][val] if defined_enums[attr]
-          @klass.type_for_attribute(attr).type_cast_for_database(val)
+          castable_val = val
+          if defined_enums[attr]
+            # `attr` is an enum. Find the number that corresponds to `val`. If `val` is
+            # a number already, there won't be a corresponding entry, just use `val`.
+            castable_val = defined_enums[attr][val] || val
```
batter: Don't think you need the || clause here if your comment is correct since you already assigned val on line 37

jaredbeck: I think we do need the disjunction, otherwise `castable_val` could be incorrectly set to `nil`.

jaredbeck:
Looks like build 1162 (https://travis-ci.org/airblade/paper_trail/builds/152138082) failed because the table under test (`post_with_statuses`) lacked timestamp columns. I keep forgetting that `touch_without_version` requires timestamps. We should probably raise an error instead of failing silently.

Hmmm, still failing on AR 5 / MySQL .. not sure why. Any ideas?

jaredbeck:
> Hmmm, still failing on AR 5 / MySQL .. not sure why. Any ideas? MySQL does have issues with fractional seconds precision .. but I'm not sure ..

Yeah, it was probably a MySQL fractional seconds issue. Adding a `Timecop.travel 1.second.since` seems to have fixed it.

Look good to you, Ben?

lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
batter: Sorry for the delayed response, but if desirable you could condense the code like so:

```ruby
castable_val = defined_enums[attr] && defined_enums[attr][val]
castable_val ||= val
```

The way you have it may be preferable since it allows for the more explicit comment.

batter:
👍

lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
jaredbeck: That looks fine (the `||=` style), Ben. As long as we can keep the comment, I have no preference.

PR #849, opened 2016-08-23
Update the Undeleting instruction

Eric-Guo:


jaredbeck:
Thanks Eric, good improvement, but there is a small problem with using `find_by`. Consider the (presumably rare) case where a model is destroyed once, reified, and destroyed again. In this case, `find_by` would incorrectly find the first version record where `event = 'destroy'`. It's a minor distinction, but hey programmers can't help but think of edge cases, right?

I'd prefer to close this in favor of something like https://github.com/airblade/paper_trail/pull/850

PR #851, opened 2016-08-23
PaperTrail is not compatible with sinatra 2.0 yet

jaredbeck:
In the meantime, this commit fixes our CI.

See https://github.com/sinatra/rack-protection/issues/117

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -11,6 +11,7 @@ appraise "ar3" do
   gem "activerecord", "~> 3.2.22"
   gem "i18n", "~> 0.6.11"
   gem "request_store", "~> 1.1.0"
+  gem "sinatra", "~> 1.4.6"
 
   group :development, :test do
     gem 'railties', '~> 3.2.22'
@@ -23,6 +24,7 @@ end
 
 appraise "ar4" do
   gem "activerecord", "~> 4.2"
+  gem "sinatra", "~> 1.4.6"
 end
 
 appraise "ar5" do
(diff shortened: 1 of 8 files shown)


Appraisals
```diff
@@ -23,16 +24,15 @@ end
...
   gem "rspec-rails", "~> 3.5.1"
   gem 'rails-controller-testing'
 
-  # Sinatra stable conflicts with AR5's rack dependency. Sinatra master requires
-  # rack-protection master. Specify exact `ref` so it doesn't break in the future.
-  # Hopefully there'll be a sinatra 2.0 release soon.
-  gem 'sinatra', github: 'sinatra/sinatra', ref: "a7483f48b0a18ba792e642a"
-  gem "rack-protection", github: "sinatra/rack-protection", ref: "7e723a74763bb83989d12"
+  # The AR5 version of PaperTrail is not compatible with sinatra 2 yet.
+  # Contributions welcome.
```
zzak: Are the compatibility issues documented somewhere?

I'm wondering if it's something I broke.

jaredbeck: Thanks for taking a look. I tried testing against sinatra 2.0.0.beta2 but got something like "url_helpers undefined". If you want to tinker with it, I can make you a branch that reproduces the error.

zzak: @jaredbeck sure!

jaredbeck: Thanks! See https://github.com/airblade/paper_trail/issues/856

PR #785, opened 2016-04-27
Do not track associations by default

jaredbeck:
[Fixes #740]

Alternative to https://github.com/airblade/paper_trail/pull/742

PR #854, opened 2016-08-25
Generate an initializer with `track_associations`,  improve docs

### LINKED ISSUE #830: Generate an initializer with track_associations=false
As discussed in #567, the `paper_trail:install` generator should generate an initializer with `::PaperTrail.config.track_associations = false`, given the deprecation warning that was added in https://github.com/airblade/paper_trail/pull/785

naw:
Closes #830

This PR updates the install generator to create an initializer:

```ruby
PaperTrail.config.track_associations = false
```

If `--with-associations` is used, the generator sets the value to `true` instead of `false`.

Also included in this PR is documentation [suggested here](https://github.com/airblade/paper_trail/issues/830#issuecomment-230179211) as well as other documentation improvements focusing on the `track_associations` configuration, configuration in general, and basic upgrade information regarding tracking associations.

Some of this might be overkill, but updating from v3 to v5, I was a little surprised to see a deprecation warning for a feature I wasn't even using, and further surprised when I didn't see mention of an initializer in the docs (or an aggregated configuration section at all). 

Please feel free to remove anything you feel is overkill, edit, or suggest improvements. Thanks!

### DIFF
--- a/lib/generators/paper_trail/install_generator.rb
+++ b/lib/generators/paper_trail/install_generator.rb
@@ -20,7 +20,8 @@ module PaperTrail
       desc: "Store transactional IDs to support association restoration"
     )
 
-    desc "Generates (but does not run) a migration to add a versions table."
+    desc "Generates (but does not run) a migration to add a versions table." \
+         "  Also generates an initializer file for configuring PaperTrail"
 
     def create_migration_file
       add_paper_trail_migration("create_versions")
@@ -29,6 +30,13 @@ module PaperTrail
         add_paper_trail_migration("create_version_associations")
         add_paper_trail_migration("add_transaction_id_column_to_versions")
       end
+    end
+
+    def create_initializer
+      create_file(
+        "config/initializers/paper_trail.rb",
+        "PaperTrail.config.track_associations = #{!!options.with_associations?}"
+      )
     end
 
     def self.next_migration_number(dirname)
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ has been destroyed.
   - [6.b. Custom Serializer](#6b-custom-serializer)
 - [7. Testing](#7-testing)
 - [8. Sinatra](#8-sinatra)
+- [9. Configuration](#9-configuration)
+- [10. Upgrading](#10-upgrading)
 
 ## 1. Introduction
 
@@ -68,7 +70,7 @@ has been destroyed.
 
     `gem 'paper_trail'`
 
-1. Add a `versions` table to your database.
+1. Add a `versions` table to your database and an initializer file for configuration:
 
     ```
     bundle exec rails generate paper_trail:install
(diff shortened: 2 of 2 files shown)


jaredbeck:
Thanks Nathan. The changes to the generator look fine. :+1: 

I have some suggestions about the proposed documentation.

Adding a "configuration" section is not in keeping with the rest of the document. We already have a few sections that deal with configuration, topically, like section 2 "Limiting What is Versioned, and When", and section 4 "Saving More Information About Versions". So, discussion of the `track_associations` option should go under section 4.b. "Associations".

Regarding the proposed "Upgrading" section. At first glance, I agree that it seems overkill. All of that information is already in the changelog. However, to someone less familiar with the project, it may be helpful. I like how the [rails upgrading guide][1] is organized, where each section focuses on a particular upgrade, e.g. "Upgrading from Rails 4.2 to Rails 5.0". Then, within that section there is a list of concrete steps to be taken. Would you mind removing the "upgrading" section from this PR and we can discuss it in a separate PR? Thanks.

[1]: http://guides.rubyonrails.org/upgrading_ruby_on_rails.html

naw:
@jaredbeck I have removed the upgrading section.  

Regarding the configuration section:

> Adding a "configuration" section is not in keeping with the rest of the document. 

The configuration section is meant to be complementary, not a replacement and not redundant.  It serves two purposes:

1. A central place to explain _how_ PaperTrail is configured at the global level   Something to clearly spell out that configuration setters should (usually) go in an initializer.  The individual sections just have a single line like `PaperTrail.config.some_option = some_value`, and the novice reader is left wondering "Where should I put that?"  Does it go in a special file, does it go in the model, etc.

2. A table-of-contents to the _globally_ configurable options.  Many things are configurable at the model level, but only a few are configured globally. Having a central place with pointers back to the relevant sections is a high-level way to communicate what is configured globally without having to piece it together from all the disparate sections.  As written, each listed option just links back to the sections you mentioned.

Of course it's your call, so all I can do is offer my perspective as someone who needed information from the README and found it difficult to find.

All of that said, I can remove the configuration section altogether if that is your preference.

jaredbeck:
In the future, when a configuration option is changed or removed, I only want to update one place in the documentation. Said another way, I don't want to describe the same configuration option in two different places.

However, I agree that a rails novice may not know where to put global configuration options like `version_limit`. How about a new section 1.e. "Configuration" with a small paragraph to that effect?

naw:
@jaredbeck I have condensed the configuration section so that it no longer mentions any specific options or settings, and I have moved it to 1.e instead of 9.

jaredbeck:
Merged. Thanks, Nathan!

lib/generators/paper_trail/install_generator.rb
```diff
@@ -31,6 +32,13 @@ def create_migration_file
       end
     end
 
+    def create_initializer
+      create_file(
```
batter: Does it make sense to only generate this file if `options.with_associations? == true` since the default is `false`? Not saying that is my preference, just asking the question.

batter:
👍 Looks good on the whole

lib/generators/paper_trail/install_generator.rb
jaredbeck: > Does it make sense to only generate this file if options.with_associations? == true since the default is false? 

Yeah, that would make sense. One fewer initializer file for most people (people not using the experimental associations features)

naw: The system outputs a deprecation warning is you don't explicitly specify false.  Therefore, all new users will see a deprecation warning if this initializer isn't generated for `false` by default, which I believe was the original goal of #830 that this PR was meant to solve.   Handling the `true` case was extra.

naw: Either that, or remove the deprecation warning, but I'm not sure what ramifications that has for users upgrading.

naw: Looks like the deprecation warning is important for people who previously were tracking associations solely by the presence of the `version_associations` table.  See #785 for reference

jaredbeck: > Looks like the deprecation warning is important for people who previously were tracking associations solely by the presence of the version_associations table.

Correct.

batter: Ok then that settles it. Explicit is the way to go here. Disregard my comment.

PR #858, opened 2016-09-03
Use `before_action` instead of `before_filter`

znz:
Warning message use `before_action` instead of `before_filter` since https://github.com/airblade/paper_trail/commit/3ce2b34b98345cad736f64a3a5e1374f40adc652 .
So I think documentations should use `before_action` too.


jaredbeck:
Thanks!

PR #859, opened 2016-09-04
Remove explicit dependency on activesupport

jaredbeck:
activerecord depends on activesupport, so there's no need for us
to explicitly depend on it.

### DIFF
--- a/paper_trail.gemspec
+++ b/paper_trail.gemspec
@@ -21,7 +21,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 1.9.3"
 
   s.add_dependency "activerecord", [">= 3.0", "< 6.0"]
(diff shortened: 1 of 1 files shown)


batter:
True, `ActiveRecord` does depend on `ActiveSupport`, so perhaps the explicit dependency is unecessary, however, [there are parts of the code where we utilize ActiveSupport directly](https://github.com/airblade/paper_trail/search?utf8=%E2%9C%93&q=ActiveSupport), and if at some point in the future ActiveSupport was removed as a dependency for ActiveRecord (doubtful this would happen anytime soon if at all), then it would break the code.

I always think it's better to be explicit about these things. Sorry for not chiming in sooner.

jaredbeck:
> .. if at some point in the future ActiveSupport was removed as a dependency for ActiveRecord ..

Seems unlikely, but if you want to revert this it's fine with me.

PR #862, opened 2016-09-07
Need a helper method to help locate all dependent versions.

vanboom:
My "dependent_versions" function allows you to collect versions of associated objects from the parent object.  

Usage:
foo = Foo.first
reified_children = foo.dependent_versions(Bar, :foo_id)

The return value can be used to view all children of the parent even if the associated child has been destroyed.

If you would like to merge this, I would be happy to update the README.md file with details.  Unfortunately, I could use help writing a spec/test for this function in your framework.

### DIFF
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -129,13 +129,12 @@ module PaperTrail
 
       def dependent_versions(class_name, foreign_key)
         children = []
-        PaperTrail::Version.where(:item_type=>class_name).where("created_at >= ?", self.created_at).each do |m|
+        PaperTrail::Version.where(item_type: class_name).where("created_at >= ?", created_at).each do |m|
           child = m.reify
-          if child.present?
-            # check the child's belongs_to to make sure it is a child
-            if child.send(foreign_key) == self.id
-              children << child
-            end
+          next unless child.present?
+          # check the child's belongs_to to make sure it is a child
+          if child.send(foreign_key) == id
+            children << child
           end
         end
         children
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -127,18 +127,6 @@ module PaperTrail
         ::PaperTrail::RecordTrail.new(self)
       end
 
-      def dependent_versions(class_name, foreign_key)
-        children = []
-        versions = PaperTrail::Version.where(item_type: class_name)
-        versions.where("created_at >= ?", created_at).each do |m|
-          child = m.reify
-          next unless child.present?
-          # check the child's belongs_to to make sure it is a child
-          children << child if child.try(foreign_key) == id
-        end
-        children
-      end
-
       # @deprecated
       def live?
         self.class.paper_trail_deprecate "live?"
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -129,13 +129,12 @@ module PaperTrail
 
       def dependent_versions(class_name, foreign_key)
         children = []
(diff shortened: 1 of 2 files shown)


lib/paper_trail/has_paper_trail.rb
```diff
@@ -129,13 +129,12 @@ def paper_trail
 
       def dependent_versions(class_name, foreign_key)
         children = []
-        PaperTrail::Version.where(item_type: class_name).where("created_at >= ?", created_at).each do |m|
```
batter: If anything, why not chain it like

```ruby
PaperTrail::Version.where(item_type: class_name).
  where("created_at >= ?", created_at).each do |m|
    ..
  end
```

lib/paper_trail/has_paper_trail.rb
```diff
@@ -127,6 +127,18 @@ def paper_trail
         ::PaperTrail::RecordTrail.new(self)
       end
 
+      def dependent_versions(class_name, foreign_key)
+        children = []
+        versions = PaperTrail::Version.where(item_type: class_name)
```
batter: I feel like this code could be cleaner like so:

```ruby
def dependent_versions(class_name, foreign_key)
  PaperTrail::Version.where(item_type: class_name).
    subsequent(created_at).map(&:reify).select do |obj|
      obj.try(foreign_key) == id
    end
  end
end
```

Also it should probably have some logic to take polymorphous associations into account which is missing at the moment, correct?

jaredbeck:
Thanks for the suggestion, Don. I think I understand *what* this does, but I don't understand *why*. Can you give a real-world example? We need to determine if this will be useful to other people as well.

> .. I could use help writing a spec/test for this function in your framework. 

Tests are required for new features. Please see our [contributing guide][1], and let me know what, specifically, you have trouble with.

Finally, this method belongs in `RecordTrail`. We aren't adding any new methods to `InstanceMethods`.

[1]: https://github.com/airblade/paper_trail/blob/master/CONTRIBUTING.md

vanboom:
Example:  Album has_many Photos


I would like to present the user a full view of all changes to Album and it's associated Photos.  In order to do that, I need to locate all of the PaperTrail Versions for the associated photos... even if they have been destroyed.


This would allow the user to resurrect the destroyed Photo objects.


Thank you for your consideration, suggestions on the code and specs.  I will try to improve it.

________________________________

lib/paper_trail/has_paper_trail.rb
vanboom: I am not sure what you mean by polymorphous associations because the foreign_key argument can be any method or attribute of the class.

jaredbeck:
> .. I need to locate all of the PaperTrail Versions for the associated photos... even if they have been destroyed. 

To clarify, you are trying to find the same version records that the following query would find?

```
PaperTrail::Version.
  joins(:version_associations).
  where(
    item_type: "Photo",
    version_associations: { foreign_key_name: "album_id", foreign_key_id: album_id }
  )
```

vanboom:
Not sure - the README recommends not using Associations in production, so I do not have a version_associations table.

________________________________

jaredbeck:
> Not sure - the README recommends not using Associations in production, so I do not have a version_associations table.

That's right, association tracking is still experimental. There are a lot of known issues, so we don't recommend its use in production. However, it sounds like this experimental feature might be helpful to you. For example, the query I've given above may meet your needs efficiently, in a single query.

I think we should be careful not to add any features that overlap, or duplicate the functionality of, association tracking. PT already has a lot of code to maintain, so we have to be careful when adding new features. @batter what do you think?

jaredbeck:
Given my concerns above, specifically that this feature overlaps with existing features, and pending a response from Ben, I'm going to close this for now.

Thanks for your contribution, Don. Please try some of our suggestions and let us know how it goes. Good luck!

PR #864, opened 2016-09-08
Remove deprecated model methods

jaredbeck:
Removes the model methods deprecated in 5.2.0 (by https://github.com/airblade/paper_trail/pull/719) Use paper_trail.x instead of x.


batter:
@jaredbeck - This should warrant a major version bump as per semantic versioning, since it makes an non backwards compatible API change, correct?

jaredbeck:
> @jaredbeck - This should warrant a major version bump as per semantic versioning, since it makes an non backwards compatible API change, correct?

Absolutely, positively, yes sir. :)

PR #868, opened 2016-09-27
issue with primary key other than `id`

notEthan:
attempting to reify records with a primary key other than `id` can run into an error `NoMethodError: undefined method 'find_by_id'` when the Version#item association does not exist, due to either not being in the default scope, or the record no longer existing.

this fixes use of `find_by_id` to instead use whatever the primary key is. 

I tried to heed rubocop warnings and follow existing testing conventions, but that may need some changes. I had some trouble testing the full matrix, will see what travis does with this PR.

### DIFF
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -31,7 +31,8 @@ module PaperTrail
           klass = version_reification_class(version, attrs)
           # The `dup` option always returns a new object, otherwise we should
           # attempt to look for the item outside of default scope(s).
-          if options[:dup] || (item_found = klass.unscoped.find_by_id(version.item_id)).nil?
+          find_cond = { klass.primary_key => version.item_id }
+          if options[:dup] || (item_found = klass.unscoped.where(find_cond).first).nil?
             model = klass.new
           elsif options[:unversioned_attributes] == :nil
             model = item_found
--- a/test/dummy/app/models/custom_primary_key_record.rb
+++ b/test/dummy/app/models/custom_primary_key_record.rb
@@ -1,0 +1,13 @@
+require "securerandom"
+class CustomPrimaryKeyRecord < ActiveRecord::Base
+  self.primary_key = :uuid
+
+  has_paper_trail class_name: "CustomPrimaryKeyRecordVersion"
+  # this unusual default_scope is to test the case of the Version#item association
+  # not returning the item due to unmatched default_scope on the model.
+  default_scope -> { where(name: "custom_primary_key_record") }
+
+  before_create do
+    self.uuid ||= SecureRandom.uuid
+  end
+end
--- a/test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
+++ b/test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
@@ -271,6 +271,23 @@ class SetUpTestTables < ActiveRecord::Migration
     end
     add_index :bar_habtms_foo_habtms, [:foo_habtm_id]
     add_index :bar_habtms_foo_habtms, [:bar_habtm_id]
+
+    # custom_primary_key_records use a uuid column (string)
+    create_table :custom_primary_key_records, primary_key: "uuid", id: :string, force: true do |t|
+      t.string :name
+      t.timestamps null: true
+    end
+
+    # and custom_primary_key_record_versions stores the uuid in item_id, a string
+    create_table :custom_primary_key_record_versions, force: true do |t|
+      t.string   :item_type, null: false
+      t.string   :item_id,   null: false
+      t.string   :event,     null: false
+      t.string   :whodunnit
+      t.text     :object
+      t.datetime :created_at
+    end
+    add_index :custom_primary_key_record_versions, [:item_type, :item_id]
   end
 
   def down
--- a/test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
+++ b/test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
@@ -287,7 +287,7 @@ class SetUpTestTables < ActiveRecord::Migration
       t.text     :object
       t.datetime :created_at
     end
-    add_index :custom_primary_key_record_versions, [:item_type, :item_id]
+    add_index :custom_primary_key_record_versions, [:item_type, :item_id], name: "idx_cust_pk_item"
   end
 
   def down
--- a/test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
+++ b/test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
@@ -273,7 +273,8 @@ class SetUpTestTables < ActiveRecord::Migration
     add_index :bar_habtms_foo_habtms, [:bar_habtm_id]
 
     # custom_primary_key_records use a uuid column (string)
-    create_table :custom_primary_key_records, primary_key: "uuid", id: :string, force: true do |t|
+    create_table :custom_primary_key_records, id: false, force: true do |t|
+      t.column :uuid, :string, primary_key: true
       t.string :name
       t.timestamps null: true
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
+- [#868](https://github.com/airblade/paper_trail/pull/868)
+  Fix usage of find_by_id when primary key is not id, affecting reifying certain records.
+
 ## 5.2.2 (2016-09-08)
 
 ### Breaking Changes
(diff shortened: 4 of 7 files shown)


test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
```diff
@@ -114,6 +114,23 @@ def up
...
+    # and orange_versions stores the orange uuid in item_id, a string
+    create_table :orange_versions, force: true do |t|
+      t.string   :item_type, null: false
+      t.string   :item_id,   null: false
+      t.string   :event,     null: false
+      t.string   :whodunnit
+      t.text     :object
+      t.datetime :created_at
+    end
+    add_index :orange_versions, [:item_type, :item_id]
```
notEthan: this is largely modeled after banana, without the Kitchen namespace, and changed to use a string item_id

spec/models/orange_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  describe "#versions" do
+    it "returns instances of OrangeVersion", versioning: true do
+      orange = described_class.create!
+      orange.update_attributes!(name: 'bob')
+      version = orange.versions.last
+      expect(version).to be_a(OrangeVersion)
+      version_from_db = OrangeVersion.last
+      expect(version_from_db.reify).to be_a(Orange)
+      orange.destroy
+      expect(OrangeVersion.last.reify).to be_a(Orange)
```
notEthan: this may be doing more than one test should

test/dummy/app/models/orange.rb
```diff
@@ -0,0 +1,9 @@
+class Orange < ActiveRecord::Base
+  has_paper_trail class_name: "OrangeVersion"
+  default_scope -> { where(:name => 'orange') }
```
notEthan: this default scope is weird but needed something that would result in Version#item not returning the record

notEthan:
ugh. I don't know why it is complaining about this 

```
lib/paper_trail/reifier.rb:34:23: C: Space inside { missing.
          find_cond = {klass.primary_key => version.item_id}
                      ^
lib/paper_trail/reifier.rb:34:60: C: Space inside } missing.
          find_cond = {klass.primary_key => version.item_id}
                                                           ^
```

that is a hash, it should not have spaces

lib/paper_trail/reifier.rb
```diff
@@ -31,7 +31,8 @@ def reify(version, options)
           klass = version_reification_class(version, attrs)
           # The `dup` option always returns a new object, otherwise we should
           # attempt to look for the item outside of default scope(s).
-          if options[:dup] || (item_found = klass.unscoped.find_by_id(version.item_id)).nil?
+          find_cond = { klass.primary_key => version.item_id }
+          if options[:dup] || (item_found = klass.unscoped.where(find_cond).first).nil?
```
jaredbeck: This part looks good to me.

test/dummy/app/models/orange.rb
```diff
@@ -0,0 +1,9 @@
+class Orange < ActiveRecord::Base
+  has_paper_trail class_name: "OrangeVersion"
+  default_scope -> { where(name: "orange") }
+
+  self.primary_key = :uuid
+  before_create do
+    self.uuid ||= SecureRandom.uuid
+  end
+end
```
jaredbeck: Instead of calling this model `Orange`, what about `CustomPrimaryKeyRecord`?

We have so many models in our test suite it's getting hard to remember how they differ.

jaredbeck: Please `require "secure_random"` at the top of this file.

test/dummy/app/versions/orange_version.rb
```diff
@@ -0,0 +1,3 @@
+class OrangeVersion < PaperTrail::Version
+  self.table_name = "orange_versions"
+end
```
jaredbeck: The `orange_versions` table is necessary because you need a table whose `item_id` column is a string, right?

test/dummy/db/migrate/20110208155312_set_up_test_tables.rb
```diff
@@ -114,6 +114,23 @@ def up
...
+
```
jaredbeck: When we commit changes to `20110208155312_set_up_test_tables.rb`, we also must commit the corresponding changes to `test/dummy/db/schema.rb`.

jaredbeck:
> that is a hash, it should not have spaces

This project follows the Ruby Style Guide, see https://github.com/bbatsov/ruby-style-guide#collections

test/dummy/app/models/orange.rb
```diff
@@ -0,0 +1,9 @@
+class Orange < ActiveRecord::Base
+  has_paper_trail class_name: "OrangeVersion"
+  default_scope -> { where(name: "orange") }
+
+  self.primary_key = :uuid
```
batter: Can we make this the first declaration within the model?

notEthan:
>> that is a hash, it should not have spaces
>
> This project follows the Ruby Style Guide, see https://github.com/bbatsov/ruby-style-guide#collections

well, quoting 

```ruby
# good - space after { and before }
{ one: 1, two: 2 }

# good - no space after { and before }
{one: 1, two: 2}
```

but I have changed it.

notEthan:
I'm not sure how to get the schema.rb updated, I have had trouble getting the tests to run locally in many cases. I think I have addressed the rest of the feedback, can you help me with the schema.rb bit?

test/dummy/app/versions/orange_version.rb
notEthan: correct

notEthan:
hm. item_id as string doesn't seem to work so well with postgres. investigating further ...

jaredbeck:
> can you help me with the schema.rb bit?

I think I ususally `cd` into `test/dummy` and run `db:migrate`.

notEthan:
okay, so this error:

```
PG::UndefinedFunction: ERROR:  operator does not exist: character varying = integer
LINE 1: ...RE "custom_primary_key_record_versions"."item_id" = 651 AND ...
                                                             ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
: SELECT  "custom_primary_key_record_versions".* FROM "custom_primary_key_record_versions"  WHERE "custom_primary_key_record_versions"."item_id" = 651 AND "custom_primary_key_record_versions"."item_type" = 'CustomPrimaryKeyRecord' ORDER BY "custom_primary_key_record_versions"."created_at" DESC, "custom_primary_key_record_versions"."id" DESC LIMIT 1
```

was just an issue with how I was specifying the uuid primary key, in a way that wasn't supported by older activerecord, and `uuid` was being incorrectly created as an int instead of string.

that does surface an issue that if you have a versions table with a string versions.item_id, it can only be used with string primary key fields in postgres. if you have an application you perhaps started with traditional numeric id fields but are transitioning to a uuid (or any string) primary key, you won't be able to use the same versions table to track both sorts of models. 

I don't have a solution for that (perhaps smart type casting in a query somewhere - not sure what capability activerecord even offers in that regard), just noting it as it occurs to me.

notEthan:
seems to be largely passing, I think feedback is addressed. `Gem::InstallError: pg requires Ruby version >= 2.0.0.` on 1.9.3 appears unrelated to this pull request

jaredbeck:
> .. Gem::InstallError: pg requires Ruby version >= 2.0.0. on 1.9.3 appears unrelated to this pull request

Correct, and fixed by https://github.com/airblade/paper_trail/pull/870. Please rebase. It would be great if you could get the tests green before we review again. Let us know if you have trouble, thanks.

notEthan:
💥 rebased, green

test/dummy/app/models/custom_primary_key_record.rb
```diff
@@ -0,0 +1,11 @@
+require "securerandom"
+class CustomPrimaryKeyRecord < ActiveRecord::Base
+  self.primary_key = :uuid
+
+  has_paper_trail class_name: "CustomPrimaryKeyRecordVersion"
+  default_scope -> { where(name: "custom_primary_key_record") }
```
jaredbeck: What's the purpose of this default_scope?

jaredbeck:
This all looks good, my only question is what is the purpose of the `default_scope` in `custom_primary_key_record.rb `?

test/dummy/app/models/custom_primary_key_record.rb
notEthan: it is to prevent the record from being returned on `version.item`. this bug surfaces in the `else` path of https://github.com/airblade/paper_trail/blob/v5.2.2/lib/paper_trail/reifier.rb#L25-L40

`version.item` returns nil when the item is destroyed, and when a default scope on the model prevents the association from returning it. I added this default scope to test both of those cases.

jaredbeck: ok that makes sense. Please add your explanation as a comment above the `default_scope`. 

Also, please add a line in `CHANGELOG.md` and I think we'll be done here.

notEthan:
added explanatory comment and CHANGELOG entry

jaredbeck:
Merged, thanks!

notEthan:
cool, thanks

PR #870, opened 2016-09-27
Constrain dev dep: pg to fix ruby 1.9.3 tests

jaredbeck:
PS: We may want to discuss dropping support for ruby 1.9.3 soon, if other important gems like pg are doing so.


jaredbeck:
IIRC rubocop has also recently dropped support for ruby 1.9.3

PR #871, opened 2016-09-29
Drop support for ruby <= 2.1

jaredbeck:
According to the [2016 Rails Hosting Survey](http://rails-hosting.com/2016/index.html), usage of ruby 1.9.3 has dropped to around 10%, with much of this drop occuring in 2016.

> Usage of 1.9.3 dropped by 16% [since the 2015 survey] and adoption of Ruby 2.2 and 2.3 jumped up to 27% each.

The 2016 survey had 1,417 respondents.

IMO it is acceptable for development of a new major version of to continue, while the 10% of the community using 1.9.3 continues to use the previous version.

Other major gems which have dropped 1.9.3 support, in no particular order:

| gem | ruby |
| --- | --- |
| devise | >= 2.1.0 |
| rubocop | >= 2.0.0 |
| rails | >= 2.2.2 |
| capistrano | > 2.0 |

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@ language: ruby
 cache: bundler
 rvm:
   - 2.3.1
-  - 1.9.3
-  - jruby-19mode
+  - 2.1.10
 env:
   global:
     - TRAVIS=true
@@ -33,9 +32,7 @@ matrix:
   fast_finish: true
   exclude:
     - gemfile: gemfiles/ar_5.0.gemfile
-      rvm: 1.9.3
-    - gemfile: gemfiles/ar_5.0.gemfile
-      rvm: jruby-19mode
+      rvm: 2.1.10
   allow_failures:
     - gemfile: gemfiles/ar_master.gemfile
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Breaking Changes
 
-- None
+- Drop support for ruby 1.9.3, whose EOL was 2015-02-23
+- Drop support for ruby 2.0.0, whose EOL was 2016-02-24
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -64,7 +64,8 @@ has been destroyed.
 
 | paper_trail    | branch     | tags   | ruby     | activerecord  |
 | -------------- | ---------- | ------ | -------- | ------------- |
-| 6              | master     | v6.x   | >= 1.9.3 | >= 4.0, < 6   |
+| unreleased     | master     |        | >= 2.1.0 | >= 4.0, < 6   |
+| 6              | 6-stable   | v6.x   | >= 1.9.3 | >= 4.0, < 6   |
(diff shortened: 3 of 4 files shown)


batter:
A few questions:

1) Is there anything that Ruby19 is missing that makes it incompatible with the current codebase? Is this mostly for the purpose of maintaining compatibility with Rubocop?
2) What version of PaperTrail are you targeting for this?

Not against it but it feels like just yesterday we dropped support for Ruby18 😄

jaredbeck:
Thanks Ben, good questions.

> 1) Is there anything that Ruby19 is missing that makes it incompatible with the current codebase? 

No.

> Is this mostly for the purpose of maintaining compatibility with Rubocop?

The primary motivation is to test against the latest version of `pg`, which we are not currently doing (see https://github.com/airblade/paper_trail/pull/870). Do you think we could add some conditional logic to our gemspec so that we could test against different versions of `pg` depending on which ruby version we are testing?

> 2) What version of PaperTrail are you targeting for this?

Version 6.0.0.

> Not against it but it feels like just yesterday we dropped support for Ruby18 😄

Yeah, it was only four months ago, in PT 5.0.0, on 2016-05-02.

I wish rubygems.org gave us statistics about what versions of ruby our users are using. They could easily collect those data.

jaredbeck:
> The primary motivation is to test against the latest version of pg, which we are not currently doing (see #870). Do you think we could add some conditional logic to our gemspec so that we could test against different versions of pg depending on which ruby version we are testing?

It looks like we can do just that. 

Closed by https://github.com/airblade/paper_trail/pull/872

Looks like we can postpone dropping 1.9.3 for now.

jaredbeck:
CI has failed again (in e.g. https://github.com/airblade/paper_trail/pull/917) because yet another dependency has dropped support for ruby 1.9.3. This time it's nokogiri, one of the most important gems in the ecosystem. I think it's time for us to follow suit on `master`. People who still need ruby 1.9.3 can use older versions of PT.

jaredbeck:
This PR also drops support for ruby 2.0.0, which reached EOL about 10 months ago. Again, people who cannot upgrade their ruby may use old versions of PT.

jaredbeck:
@batter any concerns about this? I know compatibility is important to you, but it is becoming hard to maintain, as the rest of the ruby ecosystem drops EOL rubies.

batter:
@jaredbeck - No, this seems reasonable to me since the current release of Rails requires Ruby 2.2+.  Also, since there is little to no work necessary to upgrade from 2.0 to 2.x, I think it reduces concern of complexity required for those who desire to upgrade.

Historically I remember being stuck working on apps that were stuck on Ruby 1.8 and it was a big pain to upgrade to 1.9 and successively 2.x but that doesn't seem to be the case with modern versions of ruby, so this seems reasonable to me.

jaredbeck:
> Historically I remember being stuck working on apps that were stuck on Ruby 1.8 and it was a big pain to upgrade to 1.9 and successively 2.x but that doesn't seem to be the case with modern versions of ruby, so this seems reasonable to me.

Yeah, funny how the 1.8 - 1.9 upgrade was much worse than 1.9 - 2.0. And, I agree, the 2s (2.0, 2.1) have been very easy upgrades indeed. A sign of the maturity of the interpreter, I guess.

PR #878, opened 2016-10-17
Fix typo in paper_trails callbacks docs

MistaTwista:
There was a typo in callback names not paper_trail.on_destroy, but paper_trail_on_destroy, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,13 +317,13 @@ class Article < ActiveRecord::Base
   has_paper_trail :on => []
 
   # Add callbacks in the order you need.
-  paper_trail.on_destroy    # add destroy callback
-  paper_trail.on_update     # etc.
-  paper_trail.on_create
+  paper_trail_on_destroy    # add destroy callback
+  paper_trail_on_update     # etc.
+  paper_trail_on_create
 end
 ```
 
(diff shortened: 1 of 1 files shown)


batter:
This is actually the desired syntax for the PaperTrail 5.2+ if I'm not mistaken, right @jaredbeck?

Using either syntax should work, but in the future I believe we want to deprecate the `paper_trail_method_name` syntax in favor of the `paper_trail.method_name` syntax.

@MistaTwista - See #719 for more context.

jaredbeck:
Thanks, Maxim, but `paper_trail.on_create` (with dot) is correct. See https://github.com/airblade/paper_trail/pull/719

jaredbeck:
> Using either syntax should work, but in the future I believe we want to deprecate the paper_trail_method_name syntax in favor of the paper_trail.method_name syntax.

Those methods have already been removed, by https://github.com/airblade/paper_trail/pull/864

batter:
@jaredbeck - I meant in released versions of PaperTrail (aka v5.x)

jaredbeck:
> @jaredbeck - I meant in released versions of PaperTrail (aka v5.x)

Ah, quite right. They've been removed, but only in `master`.

PR #881, opened 2016-10-18
Add `have_a_version_with_changes` matcher

aried3r:


jaredbeck:
Looks good to me. Personally, I'm a little leery of `where_object_changes` because it uses a `LIKE` query, but that doesn't seem to bother other people, so I see no reason not to have a RSpec matcher for this.

jaredbeck:
On an unrelated note, I'd like to see the rspec stuff move into a separate gem (called `rspec-paper_trail`) eventually. Doesn't have to happen now.

jaredbeck:
Merged. Thanks Anton!

aried3r:
> because it uses a LIKE query

Shouldn't performance be a little be better when people are using the `JSONB` column type? Of course, without benchmarking, setting proper indices etc I'm merely speculating.

Anyway, here's the source for anyone who's seeing this on how it is implemented, depending on the column type:
https://github.com/airblade/paper_trail/blob/5-stable/lib/paper_trail/version_concern.rb#L144-L169

batter:
> Shouldn't performance be a little be better when people are using the JSONB column type? Of course, without benchmarking, setting proper indices etc I'm merely speculating.

I would think so, but also don't have benchmarks to prove that assertion.

Should we add a bit to [the RSpec portion ofREADME](https://github.com/airblade/paper_trail#7b-rspec) regarding this feature?

aried3r:
> I would think so, but also don't have benchmarks to prove that assertion.

I'd like to do some benchmarks, but how do you suggest such a benchmark should look like? I'm asking because since you guys write and maintain paper_trail, you guys probably know best. :)

I found this article on [Rails and JSONB](http://nandovieira.com/using-postgresql-and-jsonb-with-ruby-on-rails) which also includes benchmarks ([see code](https://gist.github.com/fnando/f672c9243186933b3c8e)) which I'd like to base it on.

jaredbeck:
> I'd like to do some benchmarks, but how do you suggest such a benchmark should look like?

You probably want to benchmark the SQL queries in isolation, without ruby. So, I would take the SQL generated by PT+AR, put it in a few `.sql` files, run them using `psql` and `time`.

aried3r:
> You probably want to benchmark the SQL queries in isolation, without ruby.

Hmm, while this would measure the performance quite accurately for the PostgreSQL, wouldn't it be more useful for the users (most being Rails users, I guess) to measure "real world" performance, as in, using Ruby for the benchmark?

PR #882, opened 2016-10-24
Don't load version_association and version files for AR

ka8725:
We don't need to load the `version_association` and `version` files for the `ActiveRecord` framework because both these files are already loaded in the main file (`paper_trial.rb`) at the moment.

I need this fix because it leads to the following exception:

```ruby
Cannot define multiple 'included' blocks for a Concern (ActiveSupport::Concern::MultipleIncludedBlocks)
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/concern.rb:126:in `included'
/usr/local/Cellar/rbenv/1.0.0/versions/2.3.1/lib/ruby/gems/2.3.0/gems/paper_trail-5.2.2/lib/paper_trail/version_association_concern.rb:9:in `<module:VersionAssociationConcern>'
/usr/local/Cellar/rbenv/1.0.0/versions/2.3.1/lib/ruby/gems/2.3.0/gems/paper_trail-5.2.2/lib/paper_trail/version_association_concern.rb:6:in `<module:PaperTrail>'
/usr/local/Cellar/rbenv/1.0.0/versions/2.3.1/lib/ruby/gems/2.3.0/gems/paper_trail-5.2.2/lib/paper_trail/version_association_concern.rb:3:in `<top (required)>'
/usr/local/Cellar/rbenv/1.0.0/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bootscale-0.5.2/lib/bootscale/core_ext.rb:4:in `require'
/usr/local/Cellar/rbenv/1.0.0/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bootscale-0.5.2/lib/bootscale/core_ext.rb:4:in `require'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:274:in `block in require'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:240:in `load_dependency'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:274:in `require'
/usr/local/Cellar/rbenv/1.0.0/versions/2.3.1/lib/ruby/gems/2.3.0/gems/paper_trail-5.2.2/lib/paper_trail/frameworks/active_record/models/paper_trail/version_association.rb:1:in `<top (required)>'
/usr/local/Cellar/rbenv/1.0.0/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bootscale-0.5.2/lib/bootscale/core_ext.rb:4:in `require'
/usr/local/Cellar/rbenv/1.0.0/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bootscale-0.5.2/lib/bootscale/core_ext.rb:4:in `require'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:274:in `block in require'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:240:in `load_dependency'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:274:in `require'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:360:in `require_or_load'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:317:in `depend_on'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:233:in `require_dependency'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/engine.rb:472:in `block (2 levels) in eager_load!'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/engine.rb:471:in `each'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/engine.rb:471:in `block in eager_load!'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/engine.rb:469:in `each'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/engine.rb:469:in `eager_load!'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/engine.rb:346:in `eager_load!'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/application/finisher.rb:56:in `each'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/application/finisher.rb:56:in `block in <module:Finisher>'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/initializable.rb:30:in `instance_exec'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/initializable.rb:30:in `run'
/usr/local/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.7.1/lib/rails/initializable.rb:55:in `block in run_initializers'
```

It happens in conjunction with the `bootscale` gem.

### DIFF
--- a/lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
+++ b/lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
@@ -1,5 +1,3 @@
-require "paper_trail/version_concern"
-
 module PaperTrail
   # This is the default ActiveRecord model provided by PaperTrail. Most simple
   # applications will only use this and its partner, `VersionAssociation`, but
--- a/lib/paper_trail/frameworks/active_record/models/paper_trail/version_association.rb
+++ b/lib/paper_trail/frameworks/active_record/models/paper_trail/version_association.rb
@@ -1,5 +1,3 @@
-require "paper_trail/version_association_concern"
-
 module PaperTrail
   # This is the default ActiveRecord model provided by PaperTrail. Most simple
   # applications will only use this and its partner, `Version`, but it is


ka8725:
In other words these lines are duplications of [these ones](https://github.com/airblade/paper_trail/blob/master/lib/paper_trail.rb#L7-L8).

batter:
I don't think I follow why multiple requires of the same file is causing an error on your application since the [Ruby `require` method](https://ruby-doc.org/core-2.3.1/Kernel.html#method-i-require) by default, if a file is already loaded, will:

> Loads the given name, returning true if successful and false if the feature is already loaded.

So, it should simply return `false` in these cases and then move on. Does Bootscale always raise errors like this when a file is required twice? Seems like this may be an issue on Bootscale's end.

Personally, I think it's important to explicitly require the dependencies for these classes within the classes. These model classes may be loaded by a Rails Engine in some cases, and thus could be loaded prior to the PaperTrail base class itself. If that were to occur then you would result in a uninitialized constant error. So if anything I would lean more towards removing the requires from `paper_trail.rb` and leaving them in their respective models, but ideally we wouldn't need this at all.

batter:
Possibly related to rails/rails#15767.

@jaredbeck Looks like the Concerns could potentially be autoloaded via the `PaperTrail::Rails::Engine` class as well if we relocated them into `frameworks/active_record/models/concerns/paper_trail`. Maybe we could play around with that.

@ka8725 - can you provide a failing test using the [PaperTrail Issue Template](https://github.com/airblade/paper_trail/blob/master/.github/ISSUE_TEMPLATE.md)? That would make it a lot easier for us to work with this.

ka8725:
`bootscale` gem redefines the `require` method - https://github.com/byroot/bootscale/blob/921dcd00c68e9c5d912888ddcf66b42669b99f1e/lib/bootscale/core_ext.rb#L3

Ok, I will provide you a template soon.

jaredbeck:
Thanks for opening the issue, Andrey, and for providing a PR. That said ..

>> [when called twice, `require`] should simply return false .. Does Bootscale always raise errors like this when a file is required twice? Seems like this may be an issue on Bootscale's end.

> bootscale gem redefines the require method 

I agree Ben, calling `require` twice is not a mistake. There are often good reasons to have the same `require` repeated throughout a codebase, as you have given above.

> Possibly related to rails/rails#15767.

Good research, seems related. Did rails/rails#15767 help you, Andrey?

My intuition is that this is not a PaperTrail issue, and should be closed. If the error message can be reproduced using our bug report template *without bootscale*, we would of course be happy to reopen the issue.

PR #886, opened 2016-11-06
Adding user active active model instance.

SaimonL:
Now you can directly get the user object so you can run methods from it directly.

Not sure about where to put the code so feel free to move it from file:
` lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb `

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ v.created_at                # When the `event` occurred
 v.whodunnit                 # If the update was via a controller and the
                             # controller has a current_user method, returns the
                             # id of the current user as a string.
+v.user                      # Gets the user model instance if any (works with device).                   
 widget = v.reify            # The widget as it was before the update
                             # (nil for a create event)
 ```
@@ -227,6 +228,13 @@ version.paper_trail_originator
 version.terminator
 version.whodunnit
 version.version_author
+
+# Returns user active user model instance
+version.user
+# This method takes two arguments. First is the model name and the second is
+# primary key. By default the model is `User` and the primary key is `id`.
+# If you are using a different model (example: Customer) to store users then
+# make a call like this: version.user(:Customer)
 
 # Returns the next version.
 version.next
--- a/lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
+++ b/lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
@@ -11,7 +11,7 @@ module PaperTrail
     # You can now get user object (supports Device defaults)
     def user(user_model = :User, primary_key = :id)
       unless whodunnit.nil?
-      	user_model.to_s.constantize.find_by(primary_key.to_s, whodunnit)
+      	user_model.to_s.constantize.find_by(primary_key.to_s.concat(' = ?'), whodunnit)
       end
     end
   end
--- a/lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
+++ b/lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
@@ -7,5 +7,12 @@ module PaperTrail
   # See the readme for details.
   class Version < ::ActiveRecord::Base
     include PaperTrail::VersionConcern
+
+    # You can now get user object (supports Device defaults)
+    def user(user_model = :User, primary_key = :id)
+      unless whodunnit.nil?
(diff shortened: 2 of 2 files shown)


jaredbeck:
Hi Saimon, I like this idea, but I have some questions about the implementation. I am busy today, but maybe early this week.

SaimonL:
Next week is fine. As for implementation I have no idea where to or how to properly implement this.

SaimonL:
Woops forgot to have rails target by id.

lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
```diff
@@ -7,5 +7,12 @@ module PaperTrail
   # See the readme for details.
   class Version < ::ActiveRecord::Base
     include PaperTrail::VersionConcern
+
+    # You can now get user object (supports Device defaults)
+    def user(user_model = :User, primary_key = :id)
+      unless whodunnit.nil?
+      	user_model.to_s.constantize.find_by(primary_key.to_s.concat(' = ?'), whodunnit)
+      end
+    end
```
jaredbeck: I like the idea of taking the class name as an argument, and I like the default, `:User`.

The second argument, `primary_key`, is unnecessary. We should use `ActiveResource::Base#primary_key`.

The code should live in `version_concern.rb`. We don't put any code in `version.rb`.

Finally, I believe the hash form of `find_by` is preferable.

Putting that all together:

```ruby
def user(user_model = :User)
  unless whodunnit.nil?
    model_class = user_model.to_s.constantize
    model_class.find_by(model_class.primary_key => whodunnit)
  end
end
```

README.md
```diff
@@ -124,6 +124,7 @@ v.created_at                # When the `event` occurred
 v.whodunnit                 # If the update was via a controller and the
                             # controller has a current_user method, returns the
                             # id of the current user as a string.
+v.user                      # Gets the user model instance if any (works with device).
```
jaredbeck: Let's name this method `whodunnit_record`. Not everyone's app has "users". Also, I'd like it to be obvious that the id comes from the `whodunnit` column.

Also, what does "(works with device)" mean?

README.md
```diff
@@ -228,6 +229,13 @@ version.terminator
 version.whodunnit
 version.version_author
 
+# Returns user active user model instance
+version.user
+# This method takes two arguments. First is the model name and the second is
+# primary key. By default the model is `User` and the primary key is `id`.
+# If you are using a different model (example: Customer) to store users then
+# make a call like this: version.user(:Customer)
```
jaredbeck: Please move these detailed comments above the method definition, in `version_concern.rb`.

lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
batter: I feel like the argument name would be more accurate to be named `klass_name` or something to that effect than `user_model`

lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
```diff
@@ -7,5 +7,12 @@ module PaperTrail
   # See the readme for details.
   class Version < ::ActiveRecord::Base
     include PaperTrail::VersionConcern
+
+    # You can now get user object (supports Device defaults)
+    def user(user_model = :User, primary_key = :id)
+      unless whodunnit.nil?
+      	user_model.to_s.constantize.find_by(primary_key.to_s.concat(' = ?'), whodunnit)
```
batter: Why not use `find_by(primary_key => whodunnit)`? Skips the extra call to `to_s` and reduces complexity / improes readabiity at the same time

lib/paper_trail/frameworks/active_record/models/paper_trail/version.rb
batter: 👍 ... Only improvements I would throw out are renaming the argument name as I mentioned above, and also the method name, since as you mentioned, `user` may not be appropriate for all scenarios.

jaredbeck:
Saimon, are you still working on this? Would you like to keep this PR open?

jaredbeck:
Closing due to lack of activity. If in the future, you find time to address the problems we've raised in our code review, please feel free to open a new PR.

PR #887, opened 2016-11-11
Various 2016-11-11

jaredbeck:
Various minor refactoring in `record_trail.rb`. Working towards improving our `AbcSize` metric from 30 to 29. Baby steps. :)

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -309,19 +309,21 @@ module PaperTrail
       end
     end
 
+    # When a record is created, updated, or destroyed, we determine what the
+    # HABTM associations looked like before any changes were made, by using
+    # the `paper_trail_habtm` data structure. Then, we create
+    # `VersionAssociation` records for each of the associated records.
     def save_associations_habtm(version)
-      # Use the :added and :removed keys to extrapolate the HABTM associations
-      # to before any changes were made
       @record.class.reflect_on_all_associations(:has_and_belongs_to_many).each do |a|
         next unless
           @record.class.paper_trail_save_join_tables.include?(a.name) ||
               a.klass.paper_trail.enabled?
-        assoc_version_args = {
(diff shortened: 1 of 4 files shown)


lib/paper_trail/record_trail.rb
```diff
@@ -414,13 +410,29 @@ def add_transaction_id_to(data)
       data[:transaction_id] = PaperTrail.transaction_id
     end
 
+    # Given a HABTM association, returns an array of ids.
+    # @api private
+    def habtm_assoc_ids(habtm_assoc)
+      current = @record.send(habtm_assoc.name).to_a.map(&:id) # TODO: `pluck` would use less memory
```
batter: I believe in `ActiveRecord` you're able to call something like `@record.send("#{habtm_assoc.name}_ids") to get the current list of ids, and it should give you the most up to date values for it too...

If the concern is that it might get out of sync you could always do `@record.send("#{habtm_assoc.name}").reload` first.

batter: Upon closer inspection looks like you'll still need to invoke the association for removed and added records but the first line could be condensed using that helper

jaredbeck: Thanks Ben, I'll give it a try.

jaredbeck: > I believe in ActiveRecord you're able to call something like @record.send("#{habtm_assoc.name}_ids") to get the current list of ids ..

According to the [associations guide](http://guides.rubyonrails.org/association_basics.html#has-and-belongs-to-many-association-reference) we'd have to `singularize` the association name.

```ruby
@record.send("#{habtm_assoc.name.singularize}_ids")
```

Simply using `singularize` introduces the potential for a `NoMethodError` with irregular nouns unless properly configured in the `Inflector`. I looked for a method like `name_singular`, that would use the model config instead of relying on the inflector config, but I couldn't find one.

PR #889, opened 2016-11-17
BUGFIX: Log version validation errors with ID from record

owenr:
I came across a case where the logging of errors in PaperTrail is causing an exception (and hiding the underlying issue). I have a [convoluted example](https://gist.github.com/owenr/b0a3bf7cbe90b96d031e403bbee49a21) in the usual bug report format and a fix attached. No tests provided since I am unsure how to reproduce this without doing something complex or unusual.

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -421,7 +421,8 @@ module PaperTrail
 
     def log_version_errors(version, action)
       version.logger.warn(
-        "Unable to create version for #{action} of #{@record.class.name}##{id}: " +
+        "Unable to create version for #{action} of #{@record.class.name}#" +
+          "#{@record.id}: " +
           version.errors.full_messages.join(", ")
       )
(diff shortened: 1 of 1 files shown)


batter:
I ran your bug report and got this error:

```ruby
/Users/batkins/.rvm/gems/ruby-2.3.1@paper_trail/gems/sqlite3-1.3.10/lib/sqlite3/database.rb:91:in `initialize': SQLite3::SQLException: table versions has no column named transaction_id: CREATE  INDEX "index_versions_on_transaction_id" ON "versions" ("transaction_id") (ActiveRecord::StatementInvalid)
```

After commenting out the offending line, I re-ran it, and got the error you're reporting. Good catch here and the fix looks like it's spot on. Thanks for the PR!

jaredbeck:
Thanks Owen. This one's probably my fault from https://github.com/airblade/paper_trail/pull/719

PR #892, opened 2016-11-22
Constrain AR < 5.1

jaredbeck:
Rails 5.1 will deprecate the current behavior of AR::Dirty (https://github.com/rails/rails/pull/25337) which is pretty important to PT. I haven't found good instructions yet on how to upgrade, but Sean Griffin described it as easy, on his podcast (http://bikeshed.fm/87).

### DIFF
--- a/paper_trail.gemspec
+++ b/paper_trail.gemspec
@@ -20,7 +20,11 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
   s.required_ruby_version = ">= 1.9.3"
 
-  s.add_dependency "activerecord", [">= 3.0", "< 6.0"]
+  # Rails 5.1 will deprecate the current behavior of AR::Dirty
+  # (https://github.com/rails/rails/pull/25337) which is pretty important
+  # to PT. I haven't found good instructions yet on how to upgrade, but
+  # Sean Griffin described it as easy, on his podcast (http://bikeshed.fm/87).
+  s.add_dependency "activerecord", [">= 3.0", "< 5.1"]
   s.add_dependency "request_store", "~> 1.1"
 
   s.add_development_dependency "appraisal", "~> 2.1"


jaredbeck:
@sgrif did offer to patch PT (https://twitter.com/sgrif/status/738901870031212544) but I think he was being too generous for his own good. If it's as easy as he describes, I'm sure we can handle it :)

sgrif:
Oh thanks for the reminder. I can send a PR in a few. ;)

sgrif:
Literally all of the deprecation warnings are backwards... I need to fix this upstream.

jaredbeck:
If you do take this on (much appreciated), the instructions for running the test suite locally are in https://github.com/airblade/paper_trail/blob/master/.github/CONTRIBUTING.md or you can just open a PR and use our CI. Thanks!

sgrif:
Yup, got it running. Thanks!

sgrif:
Got everything updated. Your needs were a bit trickier, since you also call these methods from an `after_destroy` callback, which didn't persist any changes, so the "before last call to save" set of methods can't be used 100% of the time. The good news is that all of these methods already needed to be pulled out into a conditional to make sure it wasn't using them with AR < 5.1. The bad news is that conditional will continue to stick around after support for 5.0 is dropped.

I've got the rspec test suite passing. The tests run by `rake test` (your integration tests?) are failing. They were failing before, but I'm seeing one additional failure in those compared to before my changes but still pointing at Rails master. The minitest test runner makes it a pain to see *which* failure is new (especially if it's surrounded by infinite deprecation warnings), so I'm just going to go ahead and fix all of them.

Would you be ok with me also setting you up to run against Rails master as part of your CI (as an allowed failure)?

batter:
@sgrif - Thank you so much for taking a closer look at this, we very much appreciate it.

> Would you be ok with me also setting you up to run against Rails master as part of your CI (as an allowed failure)?

Seems reasonable to me if the objective of this work is to set us up for compatibility with newer versions of rails going forward. @jaredbeck is more knowledgable about how to best accomplish this but I believe the easiest approach is to slap a new group into the [`Appraisals` file](https://github.com/airblade/paper_trail/blob/master/Appraisals).

And just for clarification, the `ActiveModel::Dirty` callbacks are being deprecated in Rails 5.1 but not removed until 6.x or later, correct? rails/rails#25337 seems to indicate that they may not even be removed in 6.x?

sgrif:
The methods themselves aren't deprecated, and they will not be removed. The deprecation is for their current behavior, which will be changing in the next version of Rails (either 5.2 or 6.0) to be the same as if you had run that code after the call to `.save` (as the name `after_save` would imply). For example:

```ruby
class Foo < ActiveRecord::Base
  after_save do
    raise if changed?
  end
end

Foo.create!
```

would raise in 5.1 and earlier, but will pass in 5.2/6.0 and later. For every method in dirty, there is a new method introduced for the "old" behavior, and another new method introduced for the "new" behavior. In 5.2/6.0 the current methods in dirty will change to behave the same as the "new" behavior.

None of the methods in any of those groups are planned to be removed. Outside of `after_(save|create|update)` callbacks, the behavior of the existing methods and the behavior of the "new" methods are identical. Outside of callbacks, the "new" behavior methods have the behavior that you would get if you implemented them using `previous_changes` instead of `changes` (there are no equivalent methods in 5.0 and earlier).

PR #893, opened 2016-11-29
Suggestion to move record update data to it's own method

jnstq:
This is to open a discussion.

We have a patch in our initializers, to strip away object on updates as our tables have a lot of columns, and our users can at any point extra all their data import it again. So we need to keep it at minimum.

Instead of doing a freedom patch we could just override this method.

Is this is something that you would consider?

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -165,6 +165,15 @@ module PaperTrail
 
     def record_create
       return unless enabled?
+      versions_assoc = @record.send(@record.class.versions_association_name)
+      version = versions_assoc.create! data_for_create
+      update_transaction_id(version)
+      save_associations(version)
+    end
+
+    # Returns data for record create
+    # @api private
+    def data_for_create
       data = {
         event: @record.paper_trail_event || "create",
         whodunnit: PaperTrail.whodunnit
@@ -176,23 +185,12 @@ module PaperTrail
         data[:object_changes] = recordable_object_changes
       end
       add_transaction_id_to(data)
-      versions_assoc = @record.send(@record.class.versions_association_name)
-      version = versions_assoc.create! merge_metadata(data)
-      update_transaction_id(version)
-      save_associations(version)
+      merge_metadata(data)
     end
 
     def record_destroy
       if enabled? && !@record.new_record?
-        data = {
-          item_id: @record.id,
-          item_type: @record.class.base_class.name,
-          event: @record.paper_trail_event || "destroy",
-          object: recordable_object,
-          whodunnit: PaperTrail.whodunnit
-        }
-        add_transaction_id_to(data)
-        version = @record.class.paper_trail.version_class.create(merge_metadata(data))
+        version = @record.class.paper_trail.version_class.create(data_for_destroy)
         if version.errors.any?
           log_version_errors(version, :destroy)
         else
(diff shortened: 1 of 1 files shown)


jaredbeck:
Hi Jon. Thanks for the suggestion.

> We have a patch in our initializers, to strip away object on updates as our tables have a lot of columns, and our users can at any point extra all their data import it again. So 
>

The `object` column is a crucial part of PT, and a lot of PT features will not work without it. We consider the contents of the `object` column to be an internal implementation detail of PT. I'd be concerned about people overriding it and then being surprised when PT doesn't work.

> .. we need to keep it at minimum.

What do you want to minimize? Disk usage? Number of records in the `versions` table? You can limit the number of versions for a given record (https://github.com/airblade/paper_trail#2e-limiting-the-number-of-versions-created) but of course that's not appropriate for every app.

jaredbeck:
I do like the term "freedom patch" by the way. 😆  The freedom to patch anything you want is one of the great strengths of ruby, in my opinion.

jnstq:
> The object column is a crucial part of PT, and a lot of PT features will not work without it. We consider the contents of the object column to be an internal implementation detail of PT. I'd be concerned about people overriding it and then being surprised when PT doesn't work.

This is very true, we don't relay on so many of PT features, we mostly use it as it very convenient way to persist record changes, exclude attributes, save controller info included request uuid. So this would just make it possible to override, nothing recommended or for the README.

It's doable now to override `recordable_object` but returning nil, will result in json null and not real database null as of Rails 5.

> What do you want to minimize? Disk usage? Number of records in the versions table? You can limit the number of versions for a given record (https://github.com/airblade/paper_trail#2e-limiting-the-number-of-versions-created) but of course that's not appropriate for every app.

We have maybe 150-200 columns on some of our tables, and if only one value is changed, will result in all attributes is stored in object, and we don't use them, as our undo only uses object_changes

jaredbeck:
> So this would just make it possible to override, nothing recommended or for the README.

Yes, I see now why you added the `# @api private` comment, to discourage people from using it.

As written though, I don't like how the new method returns only a subset of the data to be recorded. I'd rather have a method that returns the entire set of data.

```ruby
# lib/paper_trail/record_trail.rb
def data_for_update(probably, some, args)
  # returns a hash, the entire set of data, including timestamps,
  # transaction id, user-supplied metadata (http://bit.ly/2gFhjmL), etc.
end
```

You would then override it using `Module#prepend`.

```ruby
module MyRecordTrailHacks
  def data_for_update
    data = super
    data[:object] = {}
    data
  end
end
::PaperTrail::RecordTrail.prepend(MyRecordTrailHacks)
```

I like this method extraction on it's own, regardless of whether it solves your problem. It separates the *preparation* of data from its *persistence*.

We should also extract `data_for_create` and `data_for_destroy`.

How does that sound?

jnstq:
Thanks for guiding me through this, `merge_metadata` do return a new object, so it's last in the data methods

I have changed as little as possible, for a short moment I got feeling, but rubocop stepped in 😄 

Great, then this is win-win. This will make it easier to maintain my freedom (patch)

jaredbeck:
Thanks Jon, looks good.

jaredbeck:
> Great, then this is win-win. This will make it easier to maintain my freedom (patch)

Freeeeeeedom!

giphy

PR #894, opened 2016-11-29
Avoid allocating new hash when merging metadata

jnstq:
What I could see, `merge_metadata` only used internal so it feels safe to rename it unless we jeopardise someones patch.

This feels more consistant with `add_transaction_id_to`

What do you think?

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -97,7 +97,7 @@ module PaperTrail
     end
 
     # @api private
-    def merge_metadata(data)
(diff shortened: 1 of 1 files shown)


jaredbeck:
I prefer "pure" functions that do not modify their arguments, so I prefer `merge` over `merge!`.

batter:
Agree with @jaredbeck. This adds complexity and also I don't love the idea of mutating the hash if there isn't a great reason to do so.

jnstq:
Point taken, thanks for quick response!

PR #896, opened 2016-11-30
Run CI against rails/master

sgrif:
This adds Rails master to CI as an allowed failure. While builds will
continue to pass regardless of this run, it makes it easier to keep an
eye on if changes need to be made for the upcoming version of Rails, as
well as automatically test shims to add compatibility


sgrif:
Ironically, this works even though https://github.com/airblade/paper_trail/blob/6411ab1ce6f661075092d7597d32e4c1f9eae7ff/paper_trail.gemspec#L27 is present because `5.1.0.alpha` and any other prerelease version is still considered to be `< 5.1`

jaredbeck:
I think that https://github.com/airblade/paper_trail/pull/898 might conflict with this. Let's merge this first so we don't have to ask Sean to rebase again *and* fix a conflict. :)

batter:
👍

PR #898, opened 2016-11-30
Drop support for rails 3

jaredbeck:
In https://github.com/airblade/paper_trail/pull/895, Sean asked:

> .. since it looks like the next version is planned to be 6.0.0, how would
> you feel about dropping support for Rails 3 in that version? It has
> reached end of life, meaning it no longer receives security patches. We
> don't recommend that gem authors continue to support it.

To which, I responded:

> We could continue maintaining PT 5 if we had to, so I think it's
> fine to drop support for rails 3 in PT 6. If Ben agrees, I'd be
> happy to make that change. I'll just update the gemspec and drop
> testing support, for starters. We can clean up the conditionals in
> the code over time.


jaredbeck:
What do you think, Ben? @batter

jaredbeck:
Please merge https://github.com/airblade/paper_trail/pull/896 before this, in case of conflict.

jaredbeck:
Ben, can you please weigh in on this? Thanks. @batter

batter:
Initially I was reluctant to endorse this since it didn't seem like there was a great reason to drop support for ActiveRecord 3, however, after looking more closely at the code in #899, it appears that the functionality and supporting methods for retrieving diffs between objects changed in memory vs the database via `ActiveModel::Dirty` have improved significantly after ActiveRecord 3, and the way we were accomplishing it prior (invoking `send`) always felt a little hackish.

As both you and @sgrif noted, since we're looking to merge these changes to support AR 5.1 more effectively in before releasing PaperTrail 6, this seems like the appropriate time to do it, as major releases are the appropriate time to make incompatible API changes and drop backwards compatibility as per semantic versioning standards.

In addition, it seems like that it could cause significant complexity to attempt to continue supporting both AR 3 and future versions. I also agree that we could continue backporting some bugfixes (or even functionality) to PT 5 if significant and worthy alterations were found. I'm always in support of changes that reduce complexity.

*TLDR* - I was reluctant initially but I think it's appropriate since it reduces complexity going forward and makes it easier to support future versions of ActiveRecord.

PR #899, opened 2016-11-30
Add support for Rails 5.1

sgrif:
This updates the code to work with Rails 5.1.0.alpha, as of commit
https://github.com/rails/rails/commit/1bdc395d956f789b6612796ac6f130cde90d3066.

The biggest change was to call the new methods provided by
`ActiveRecord::Dirty` when needed. In the next version of Rails after
5.1, dirty will be "cleared" before after_save callbacks are run. This
means that code in an after_save callback will behave the same as if it
was run after calling `.save`.

Since the concept of "changed" is fairly nebulous, two new method sets
were introduced with names that specify whether they're operating on
changes from what we think is in the database to what's in memory, or
the changes that were just persisted. Paper trail will now use the
latter set of methods if available and called from an after_ callback.

There were a few other unrelated changes required to get this working on
master. The first was the change from `appear_as_new_record` to
`appear_as_unpersisted`. This was due to a single test that broke as a
result of
https://github.com/rails/rails/commit/c546a2b045600b6d700140adf2a4df666d6e08ce
where reifying a version with a nil has_one association was persisting the
change to the foreign key. I am actually unsure how that commit caused
the problem (or more specifically, I'm unsure how it was passing
before). However, as best as I can tell, the purpose of
`appear_as_new_record` was to attempt to prevent the callbacks in
`AutosaveAssociation` (which is the module responsible for persisting
foreign key changes earlier than most people want most of the time
because backwards compatibility or the maintainer hates himself or
something) from running. By also stubbing out `persisted?` we can
actually prevent those. A more stable option might be to use `suppress`
instead, similar to the other branch in `reify_has_one`.

The remaining breakage was due to the `Song` model relying on internals
that have changed from underneath it. From Rails 5 onwards there is a
public API available to do what it wants, so we can just use that
instead.

This commit is not expected to pass on Rails 3, as #898 makes it sound
like support might be dropped. If this needs to be ammended to support
Rails 3, I will do so, but I will also grumble very loudly about it.

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -23,13 +23,13 @@ module PaperTrail
     end
 
     def attributes_before_change
-      @record.attributes.map do |k, v|
+      Hash[@record.attributes.map do |k, v|
         if @record.class.column_names.include?(k)
           [k, attribute_in_previous_version(k)]
         else
           [k, v]
         end
-      end.to_h
+      end]
     end
 
     def changed_and_not_ignored
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -3,24 +3,33 @@ module PaperTrail
   class RecordTrail
     def initialize(record)
       @record = record
+      @in_after_callback = false
     end
 
     # Utility method for reifying. Anything executed inside the block will
     # appear like a new record.
-    def appear_as_new_record
+    def appear_as_unpersisted
       @record.instance_eval {
         alias :old_new_record? :new_record?
         alias :new_record? :present?
+        alias :old_persisted? :persisted?
+        alias :persisted? :nil?
       }
       yield
-      @record.instance_eval { alias :new_record? :old_new_record? }
+      @record.instance_eval {
+        alias :new_record? :old_new_record?
+        alias :persisted? :old_persisted?
+      }
     end
 
     def attributes_before_change
-      changed = @record.changed_attributes.select { |k, _v|
-        @record.class.column_names.include?(k)
-      }
-      @record.attributes.merge(changed)
+      @record.attributes.map do |k, v|
+        if @record.class.column_names.include?(k)
+          [k, attribute_in_previous_version(k)]
+        else
+          [k, v]
+        end
+      end.to_h
     end
 
     def changed_and_not_ignored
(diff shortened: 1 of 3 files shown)


sgrif:
This PR resulted in https://github.com/rails/rails/commit/ef993d36bc952046f312a3695617eaa6d5fe352e and https://github.com/rails/rails/commit/1bdc395d956f789b6612796ac6f130cde90d3066 upstream so thanks for that!

lib/paper_trail/record_trail.rb
```diff
@@ -474,5 +489,41 @@ def version
...
+    def attribute_changed_in_latest_version?(attr_name)
+      if @in_after_callback && rails_51?
+        @record.saved_change_to_attribute?(attr_name.to_s)
+      else
+        @record.attribute_changed?(attr_name.to_s)
+      end
+    end
+
+    def rails_51?
+      ActiveRecord::VERSION::MAJOR >= 5 && ActiveRecord::VERSION::MINOR >= 1
```
sgrif: This can also be written as `ActiveRecord.version >= Gem::Version.new("5.1.0.alpha")` if we don't need to support Rails 3. Unsure which you'd prefer, so I opted for the non-allocating version since this seems like it's called from a semi-hotpath

jaredbeck:
Please also update the `gemspec` and `Appraisals`. See #898 for new naming convention in `Appraisals` (eg. `appraise "ar-5.0"`). For example, you should add `appraise "ar-5.1"`, please.

sgrif:
I didn't add a new appraisal, as there is no released version of Rails 5.1 for it to use.

jaredbeck:
> I didn't add a new appraisal, as there is no released version of Rails 5.1 for it to use.

Oh right, of course, and there's no `5.1-stable` branch yet either. Plus, we already have an `appraise "ar_master"`. Sounds good.

sgrif:
CI looks to be in its expected state. ar3 failed (will fix if y'all decide not to drop support). Master with 2+ passes, fails with 1.9

jaredbeck:
Sean. We've dropped support for rails 3, please rebase, thanks.

jaredbeck:
Rebased, preserving authorship, and merged as https://github.com/airblade/paper_trail/pull/900

Thanks, Sean.

To do:

- Method-level documentation
- Possible improvements to method names
- Alphabetize methods
- Add changelog entry
- Release PT 6.0.0

sgrif:
Thanks, Jared. Sorry I didn't get to this today.

jaredbeck:
No worries. We really appreciate you going above and beyond your rails team duties on this one :)

PR #900, opened 2016-12-03
Sean's rails 5.1 work, rebased

jaredbeck:


jaredbeck:
For posterity, this was a rebase of https://github.com/airblade/paper_trail/pull/899 which was first discussed in https://github.com/airblade/paper_trail/pull/892

PR #904, opened 2016-12-04
Various 20161204181017

jaredbeck:


jaredbeck:
Ben, please see `d84dea3` and `fbe3846`. The cost of supporting ruby 1.9.3 is still manageable, but I did loose an hour or so today fighting to maintain said support. I don't think we'll be able to support it for too much longer. Maybe a few more months? @batter

batter:
Is the difficulty primarily a result of the fact that ActiveRecord 5 requires Ruby 2?

jaredbeck:
> Is the difficulty primarily a result of the fact that ActiveRecord 5 requires Ruby 2?

It's true that AR 5 requires ruby >= 2.2, but the problems I ran into in `d84dea3` and `fbe3846` were with rack and rubocop, respectively. With rack, I was able to use the same gemspec hack we used with pg, but with rubocop that will not be possible for the reasons I gave in `fbe3846`, so we are stuck on an outdated version of rubocop until we can drop support for ruby 1.9.3.

PR #905, opened 2016-12-05
only warn logger if present

andyweiss1982:
Currently it is possible to run into this error message: `private method 'warn' called for nil:NilClass`, which will crash apps that do not have `ActiveRecord::Base.logger` set.  

It's a little confusing to draw a straight line from the original source of the error (`"Attribute #{k} does not exist on #{version.item_type} (Version id: #{version.id})."`) to the error message currently displayed in the console.

I would think it would be preferable to only call `warn` on a logger if one is present, eliminating possible sources of 500 errors and allowing developers to use `paper_trail` without explicitly setting a logger class for `PaperTrail::Version` or `ActiveRecord::Base`

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -505,7 +505,7 @@ module PaperTrail
       version.logger.warn(
         "Unable to create version for #{action} of #{@record.class.name}" +
           "##{@record.id}: " + version.errors.full_messages.join(", ")
-      )
+      ) if version.logger
     end
 
     # Returns true if the given HABTM association should be saved.
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -179,7 +179,7 @@ module PaperTrail
             model[k.to_sym] = v
           elsif model.respond_to?("#{k}=")
             model.send("#{k}=", v)
-          else
+          elsif version.logger
             version.logger.warn(
               "Attribute #{k} does not exist on #{version.item_type} (Version id: #{version.id})."
             )


lib/paper_trail/record_trail.rb
```diff
@@ -502,7 +502,7 @@ def habtm_assoc_ids(habtm_assoc)
     end
 
     def log_version_errors(version, action)
-      version.logger.warn(
+      version.logger&.warn(
```
batter: What does the ampersand character do in this context?

Does it not make more sense to just wrap these statements on a `if` block that checks to see if `version.logger` is not `nil`?

batter: Ok, I think I understand now, however according to [this stackoverflow post](http://stackoverflow.com/a/36812667/995874), this functionality was introduced in Ruby 2.3, which would make the gem unusable with earlier Ruby versions, which is not desirable.

andyweiss1982: OK that is certainly a fair critique.  I can resubmit with the conditional inlined explicitly.

andyweiss1982:
@batter what do you make of the most recent change, replacing `version.logger` with `version.logger || Logger.new(STDOUT)`?

jaredbeck:
> @batter what do you make of the most recent change, replacing version.logger with version.logger || Logger.new(STDOUT)?

This will prevent people from disabling output, which I assume is the goal of setting `ActiveRecord::Base.logger` to nil. So, I don't think it's a good idea.

batter:
I suppose @jaredbeck is probably right. Personally I would prefer to see feedback but I know not everyone wants that, and we've gotten complaints in the past from some users who had difficulty silencing warnings that PaperTrail was passing, so I think the original proposed `if` statement wrap is probably the best approach.

andyweiss1982:
@batter sounds good, changed it back to plain vanilla `if version.logger`

batter:
Looks good to me, and the tests are passing. Can you squash this into a single commit, then I think it's ready to merge, cheers!

andyweiss1982:
Great, all squashed.  Thanks!

batter:
Cheers!

jaredbeck:
Released in 6.0.2

Issue #916, opened 2016-12-27
MyModel.paper_trail.disable should take a block


wconrad:
Today I did this:

    MyModel.paper_trail.disable
    ...
    MyModel.paper_trail.enable

What I really wanted to do was this:

    MyModel.paper_trail.disable do
      ...
    end

Would you accept a PR to allow #disable to take a block, such that #enable would be called implicitly when the block ended?

-----

Note: This issue started a conversation about what the API should be.  [What I am now proposing][1] is a little different than what I proposed above.

[1]: https://github.com/airblade/paper_trail/issues/916#issuecomment-269977666

batter:
This functionality already exists via the `without_versioning` method. [See the README for usage instructions and examples](https://github.com/airblade/paper_trail#per-method).

wconrad:
I think that #without_versioning is a poor substitute for allowing #disable to take a block, because:

* #without_versioning is called on an instance, but it acts upon the model (based on your comment, and me reading the code).  That's surprising.
* #without_versioning requires me to create an instance even if I don't need one yet.

Would you accept a PR for a README change to make it clear what #without_versioning affects the entire model and not just that instance?

batter:
Sure, and we could even make a class level method which has the same behavior if you feel that it's useful.

wconrad:
I do feel it would be useful.  Where would you like the class method to be?  I earlier proposed allowing `MyModel.paper_trail.disable` to take an optional block.  Do you prefer a different signature?

batter:
I feel like it would be good to stay consistent with the 'without_versioning' method name. Perhaps the disable method can be modified so if it receives a block it passes it over to the without_versioning class method.

I will say that if the 'disable' method can receive a block then it raises the question as to whether the 'enable' method should be able to receive a block with the inverse behavior, right?

jaredbeck:
> I feel like it would be good to stay consistent with the 'without_versioning' method name.

I agree. That method name will be familiar to people.

> I will say that if the 'disable' method can receive a block then it raises the question as to whether the 'enable' method should be able to receive a block with the inverse behavior, right?

I think, for that reason, that neither `enable` nor `disable` should accept blocks. Let's keep them simple, if we can.

In general, I think there are already too many ways to switch PT on and off!  See docs [section 2.d.](https://github.com/airblade/paper_trail#2d-turning-papertrail-off).

wconrad:
[...]

PR #917, opened 2016-12-28
Explain the scope of `without_versioning`

wconrad:
Per a conversation with @batter in issue #916, adds a note to the README that explains the scope of #without_versioning.


jaredbeck:
Merging, thanks Wayne.

PR #918, opened 2017-01-13
comma splice

brchristian:


batter:
`¯\_(ツ)_/¯`

brchristian:
I taught university English for several years. [Technically](https://owl.english.purdue.edu/engagement/2/1/34/) you aren’t supposed to put a comma between two independent clauses. Just sayin’. :)

batter:
I believe you that your suggestions is gramatically correct, but I don't believe it really matters in this case 😉

PR #925, opened 2017-02-06
Update RSpec matchers to work with custom version association names

kitop:
`have_a_version_with` and `have_a_version_with_changes` had the
`versions` association call hardcoded.
If `has_paper_trail` had the `versions` option set to something else
(e.g: `has_paper_trail versions: :paper_trail_versions`) the matcher
couldn't be used.

This fixes that issue by dynamically fetching the association name from
the `:versions_association_name` class attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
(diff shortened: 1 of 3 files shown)


jaredbeck:
Thanks, Esteban! Looks good. Please add a test in `spec/models/document_spec.rb`, and add an entry to the changelog.

kitop:
Sure. On it!

kitop:
Done, and squashed.

jaredbeck:
Nice work, thanks!

PR #926, opened 2017-02-08
Changelog: Enum-related breaking change in v4.1.0

olliebennett:
A breaking change from `v4.0.2` to `v4.1.0` is not documented in the [changelog](https://github.com/airblade/paper_trail/blob/master/CHANGELOG.md#410-2016-01-30).

Arguably this is the introduction of a bug, rather than a deliberate change, so perhaps it is not worth adding to the changelog.

The behaviour change is around retrieval of `enum` values from a version's changeset. After the update, the display behaviour seems to have changed from string-based to integer-based.

To reproduce this, I used the following code:

```ruby
# my_model.rb
class MyModel < ActiveRecord::Base
  enum my_enum: { foo: 4, bar: 5 }
  has_paper_trail
end

# Create the version
sample = MyModel.create(my_enum: :foo)
sample.update!(my_enum: :bar)

# displaying human-readable changeset
changeset = MyModel.last.versions.last.changeset
from = changeset['my_enum'][0]
to = changeset['my_enum'][1]
puts "Enum value changed from '#{from}' to '#{to}'."
```

Output under `paper_trail` from `4.0.2`:
> Enum value changed from 'foo' to 'bar'.

Output under `paper_trail` from `4.1.0`.
> Enum value changed from '0' to '0'.

I've not checked the behaviour of the latest `paper_trail` version; this proposal is simply to fix the changelog (even though it is for quite an outdated version).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -299,6 +299,10 @@ name conflict with your methods.
 
 ## 4.1.0 (2016-01-30)
 
+### Known Issues
+
+- Version changesets now store ENUM values incorrectly (as nulls). Previously the values were stored as strings. This only affects Rails 4, not Rails 5. See [#926](https://github.com/airblade/paper_trail/pull/926)
+
 ### Breaking Changes
 
 - None


olliebennett:
This could be related to https://github.com/airblade/paper_trail/issues/798 as it seems similar. Regardless, I suspect the problem itself is resolved in more recent versions of paper_trail.

batter:
Also related #857. I understand the issue, would be nice to understand what triggered the change between the two versions, but this seems like a reasonable warning for now since we we're now aware it's broken.

jaredbeck:
Hi Ollie. I'm fine with having this information in the changelog. Listing "known issues" in the changelog is quite uncommon in the ruby community, but common in other communities.

I'd prefer the heading "Known Issues" to "Breaking Changes", to separate the unintentional from the intentional, respectively. If that's acceptable to you, please make that change.

There have been many issues with enums over the past few versions (some due to changes in activerecord, some just bugs in PT) and I need to review the issues to confirm that this happened in v4.1.0.

olliebennett:
Thanks; I have moved it to a "Known Issues" section; good idea.

jaredbeck:
Ollie, I've tested PT 4.0.2 and 4.1.0 with your example. (https://gist.github.com/jaredbeck/e8a2b572c806bfff782ed6df12f48715) With rails 4, the output differs as you describe. However, with rails 5, they produce the same output. So, this is only an known issue for rails 4. Please update your patch to reflect this.

olliebennett:
Good observation; I should have mentioned I am using Rails 4. I've modified the commit. Thanks for the help.

jaredbeck:
Merged. Thanks, Ollie!

PR #932, opened 2017-03-01
Adds PaperTrail.whodunnit(value) { ... }

Dorian:
Allows to set whodunnit for the block, e.g.

```ruby
PaperTrail.whodunnit("me") { Booking.past.not_charged.each(&:charge) }
```

In my case I want to set `whodunnit` as the method name when a worker is at the origin of a change, so that it's easy to trace which method caused the change, also avoiding false negatives by resetting it at the end.


jaredbeck:
Hi Dorian, thanks for the PR!

Some applications will want this feature to affect all threads, but other applications will want it to only affect one thread. So, that's something we should talk about. Do we support both? Most PT config. is per-thread.

Dorian:
@jaredbeck I guess it's per-thread already (not 100% how to test threads):

```ruby
> PaperTrail.whodunnit = "main"
> Thread.new { puts PaperTrail.whodunnit }.join

> puts PaperTrail.whodunnit
main
```

Dorian:
@jaredbeck From my testing, it seems like it's thread-safe, another one I didn't included:

```ruby
def d; puts PaperTrail.whodunnit; end
d; PaperTrail.whodunnit("main") do
  d; Thread.new { d; PaperTrail.whodunnit = "other"; d }.join; d
end; d
"main

other
main
"
```

jaredbeck:
>> Some applications will want this feature to affect all threads, but other applications will want it to only affect one thread. So, that's something we should talk about. Do we support both? Most PT config. is per-thread.

> From my testing, it seems like [my implementation is] thread-safe ..

Yes, your implementation `paper_trail_store` (a RequestStore) which is a thread-local variable. 

What I would like to discuss is whether we want two versions of this feature: one that uses a per-thread variable, and another that uses per-process variable.

Dorian:
@jaredbeck So RequestStore is actually a "per-request" thread-safe store, so that's even better, nice!

jaredbeck:
> @jaredbeck So RequestStore is actually a "per-request" thread-safe store, so that's even better, nice!

Yeah, it uses `Thread.current#[]` which is actually documented as being "fiber-local". I don't know much about fibers, though.

Dorian:
@jaredbeck Anyway, then this looks good to me, like you said:

> Most PT config. is per-thread.

Anything I should change?

jaredbeck:
>> What I would like to discuss is whether we want two versions of this feature: one that uses a per-thread variable, and another that uses per-process variable.

> @jaredbeck Anyway, then this looks good to me .. Anything I should change?

So, what is your response to my question? You think that we should not offer a per-process version of this feature?

Dorian:
@jaredbeck Yup, I think this is fine, no need for per-process version for now

jaredbeck:
> @jaredbeck Yup, I think this is fine, no need for per-process version for now

OK, I was hoping to have a more in-depth discussion about this. What would we do in the future if someone wants a per-process version? What would the API look like?

Dorian:
@jaredbeck If somebody complains, then this will need to be fixed, sure.

I think the problem is deeper and they would not want to use RequestStore and provide their own store, maybe a redis instance, some filesystem based system (e.g. File.read/write('tmp/paper_trail/whodunnit'), some fancy distributed synchronization system Google-style with tokens and votes etc.

API could look like:

```
class MyPaperTrailStore
  def read(key)
    File.read("tmp/paper_trail/#{key}")
  end

  def write(key, value)
    File.write("tmp/paper_trail/#{key}", value)
  end
end

PaperTrail.paper_trail_store = MyPaperTrailStore
```

(Obviously this store is not complete, more like a PoC)

jaredbeck:
Hi Dorian, Sorry for taking so long to get back to you on this. I think you made a good point about which use case is more common.

> ~95% of paper trail use cases is in rails requests, for the remaining ~5% in scripts/workers, they often have only one thread. .. [therefore] .. no need for per-process version for now

So, I think the API you've chosen is good. Your tests look good too. Let me see if I can fix the merge conflicts.

jaredbeck:
Closed by https://github.com/airblade/paper_trail/pull/954, preserving your authorship. Thanks Dorian.

Dorian:
@jaredbeck Thanks :)

PR #934, opened 2017-03-03
Add rubocop extensions: rubocop-rspec and rubocop-thread_safety

Dorian:
So, apparently `PaperTrail.config` is not thread-safe:

```
    def config
      @config ||= PaperTrail::Config.instance
      yield @config if block_given?
      @config
    end
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,7 @@
+require:
+  - rubocop-rspec
+  - rubocop-thread_safety
+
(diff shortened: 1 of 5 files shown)


jaredbeck:
> So, apparently PaperTrail.config is not thread-safe

Yes, this is by design. We want certain configuration options to affect all threads, e.g. `PaperTrail.enabled`.

PR #938, opened 2017-03-19
Fix belongs_to reify for associations that use a different foreign_key

elju:
Paper trail currently fails to reify belongs to associations that use a non-standard foreign key. You can see this by spinning up an example rails app:
```
rails g model Person name:string
rails g model Dog name:string owner_id:integer

# app/models/person.rb
class Person < ApplicationRecord
  has_paper_trail
end

# app/models/dog.rb
class Dog < ApplicationRecord
  has_paper_trail
  belongs_to :person, foreign_key: :owner_id
end
```
and then creating a few models and attempting to do `dog.versions.last.reify(belongs_to: true)`. You get:
```
irb(main):002:0> dog.versions.last.reify(belongs_to: true)
  PaperTrail::Version Load (0.2ms)  SELECT  "versions".* FROM "versions" WHERE "versions"."item_id" = ? AND "versions"."item_type" = ? ORDER BY "versions"."created_at" DESC, "versions"."id" DESC LIMIT ?  [["item_id", 1], ["item_type", "Dog"], ["LIMIT", 1]]
  Dog Load (0.1ms)  SELECT  "dogs".* FROM "dogs" WHERE "dogs"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
NoMethodError: undefined method `person_id' for #<Dog:0x005642ccdf5290>
```

`PaperTrail::Reifiers::BelongsTo`, I believe, is checking the wrong association method to get its foreign key. It can just check `assoc.foreign_key` instead of `assoc.association_foreign_key` which I believe is designed for `has_many` situations. Happy to provide any other information or do any extra work it takes to get this merged in. Also, I'm not sure if you guys do backports, but my work is currently using paper trail 5, so it would be great to get it merged in there as well if possible!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
   matchers to work with custom version association names
 - [#929](https://github.com/airblade/paper_trail/pull/929) -
   Fix error calling private method in rails 4.0
+- [#938](https://github.com/airblade/paper_trail/pull/938) - Fix bug where
+  non-standard foreign key names broke belongs_to associations
 
 ## 6.0.2 (2016-12-13)
 
--- a/lib/paper_trail/reifiers/belongs_to.rb
+++ b/lib/paper_trail/reifiers/belongs_to.rb
@@ -6,7 +6,7 @@ module PaperTrail
       class << self
         # @api private
         def reify(assoc, model, options, transaction_id)
-          id = model.send(assoc.association_foreign_key)
+          id = model.send(assoc.foreign_key)
           version = load_version(assoc, id, transaction_id, options[:version_at])
           record = load_record(assoc, id, options, version)
           model.send("#{assoc.name}=".to_sym, record)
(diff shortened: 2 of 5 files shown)


jaredbeck:
Hi Luke, sounds like you're onto something. Please add:

1. A test. `spec/models/person_spec.rb` would probably be the easiest place, because `Person` has a non-standard foreign key already.
2. A changelog entry under Unreleased -> Fixed.

> Also, I'm not sure if you guys do backports, but my work is currently using paper trail 5, so it would be great to get it merged in there as well if possible!

We can probably manage that. Let's get it fixed in master first and then you can open a new PR for the backport, targeting the `5-stable` branch.

elju:
Ok! Just added the test and the change log. I ended up needing to add in a foreign key for `Person` since none of the records appears to have the the type of association I was looking for (`Person`s other association used a default foreign_key name). Let me know if there's anything you need me to fix!

jaredbeck:
Merged, thanks Luke!

Issue #940, opened 2017-03-27
enforce_version_limit! assumes database records are returned in correct order


danbernier:
When setting `PaperTrail.config.version_limit` to N, the expectation is that, when an update creates the (N+1)th version of a record, the _oldest_ versions for that record will be deleted.

This behavior is implemented in `PaperTrail::VersionConcern#enforce_version_limit!`, which doesn't order the versions by `created_at`, so it deletes the last versions _based on the order the database returns them._ This is very often insertion order, but it's often some other order, based on whatever is efficient for the database. You can verify this by calling `to_sql` on the relation that's used, eg, `sibling_versions.to_sql`. You get this (reformatted with whitespace):
```
SELECT "versions".*  FROM "versions" 
WHERE "versions"."item_type" = 'Widget' 
  AND "versions"."item_id" = 1
```

This behavior - the database returning records in a different order - is hard to reproduce reliably, but is a generally-understood property of RDBMs, so I felt justified in hacking to simulate it. While investigating this issue and working on a fix, I wrote a spec to verify that `#enforce_version_limit!` deletes the wrong records when they're returned out-of-insertion-order. When I noticed your bug-report template requires a reproduction, I reused that spec (only a minor port from rspec to minitest).

I'll submit a PR with the spec, and the change that fixes it (by adding an `ORDER` clause), shortly. The new SQL will be:

```
SELECT "versions".* FROM "versions" 
WHERE "versions"."item_type" = 'Widget' 
  AND "versions"."item_id" = 1  
ORDER BY "versions"."created_at" ASC
```

```
# Use this template to report PaperTrail bugs.
# Please include only the minimum code necessary to reproduce your issue.
require "bundler/inline"

# STEP ONE: What versions are you using?
gemfile(true) do
  ruby "2.3.1"
  source "https://rubygems.org"
  gem "activerecord", "4.2.7.1"
  gem "minitest", "5.9.0"
  gem "paper_trail", "5.2.0", require: false
  gem "sqlite3"
  gem "timecop"
end

require "active_record"
require "minitest/autorun"
require "logger"
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = nil
ActiveRecord::Schema.define do
  # STEP TWO: Define your tables here.
  create_table :widgets, force: true do |t|
    t.text :name, null: false
    t.timestamps null: false
  end

  create_table :versions do |t|
    t.string :item_type, null: false
[...]

PR #941, opened 2017-03-27
When enforcing version_limit, order the version records by created_at

danbernier:
The details of the issue are in #940.

I split the commits out into visible, obvious parts. If you checkout the middle commit, db4c586, you can see the failure by running `rake spec`. I'm more than happy to squash these down, if the maintainers would like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
   Fix error calling private method in rails 4.0
 - [#938](https://github.com/airblade/paper_trail/pull/938) - Fix bug where
   non-standard foreign key names broke belongs_to associations
+- [#940](https://github.com/airblade/paper_trail/pull/940) - When destroying
+  versions to stay under version_limit, don't rely on the database to
+  implicitly return the versions in the right order
 
 ## 6.0.2 (2016-12-13)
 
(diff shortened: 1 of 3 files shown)


jaredbeck:
Thanks Dan. Initial review:

1. Please add a changelog entry under Unreleased -> Fixed
1. I think that the SQL generated by `sibling_versions` should not change, should remain unordered.
1. Unless you feel strongly about keeping it, I don't think we need the "here comes the crazy part" spec. It's a bit hard to follow. If we can replace it with a simpler spec, perhaps by inserting specific values of `created_at`, that'd be fine.

danbernier:
> I think that the SQL generated by sibling_versions should not change, should remain unordered.

So, move the `order` call to the call site, in `enforce_version_limit!`? I can do that.

> ...replace it with a simpler spec, perhaps by inserting specific values of created_at, that'd be fine.

I'll give that a try. Agreed, that'll be much simpler.

danbernier:
@jaredbeck, I made the changes - the CHANGELOG, and moving the `order` out of `sibling_versions`.

I also changed the spec to create fake Version records. It's definitely shorter, but I'm not sure if it's easier to understand. Let me know what you think.

jaredbeck:
Looks good, thanks Dan!

jaredbeck:
Squashed offline into 300a16c

danbernier:
Thank you @jaredbeck! When do you think this fix will reach rubygems?

jaredbeck:
We're very close to the next release, 7.0.0. It could happen this week if all goes well.

PR #956, opened 2017-04-26
Fix ActiveRecord version check

newtrat:
Minor change: the `RAILS_GTE_5_1` version check in `record_trail.rb` looks wrong to me. It checks that ActiveRecord's major version is at least 5 and minor version is at least 1, but I think this check would exclude releases like 6.0 and 7.0.

I've changed the check to one that will match what I think was intended.

Thanks for your hard work on this gem; it's been working great for us so far!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#956](https://github.com/airblade/paper_trail/pull/956) -
+  Fix ActiveRecord >= 5.1 version check 
 
 ## 7.0.1 (2017-04-10)
 
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -1,7 +1,8 @@
 module PaperTrail
   # Represents the "paper trail" for a single record.
   class RecordTrail
-    RAILS_GTE_5_1 = ::ActiveRecord::VERSION::MAJOR >= 5 && ::ActiveRecord::VERSION::MINOR >= 1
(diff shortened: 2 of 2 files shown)


jaredbeck:
> .. I think this check would exclude releases like 6.0 and 7.0.

I think you're right, thanks.

How about `::ActiveRecord.gem_version >= ::Gem::Version.new("5.1.0.beta1")`?

newtrat:
Ooh, I didn't realize that method existed! Yeah that's much nicer.

jaredbeck:
Yeah, I think it was added in some version of rails 4. Hopefully rails 4.0, because we still try to support 4.0 (at least for a few more months).

jaredbeck:
Please also add an entry to the changelog under unreleased -> fixed. Thanks.

newtrat:
Hmm. Actually it looks like it wasn't added until 4.2, in [this giant commit](https://github.com/rails/rails/commit/2dd2fcf89673afbcf95240ecebaf34826a195164). I'm basing this on the [Github history](https://github.com/rails/rails/commits/master/activerecord/lib/active_record/gem_version.rb) for the gem_version.rb file.

Should we stick with the manual integer comparison method, then?

jaredbeck:
> Hmm. Actually it looks like it wasn't added until 4.2, in this giant commit. I'm basing this on the Github history for the gem_version.rb file.
> 
> Should we stick with the manual integer comparison method, then?

Can we do something like:

```
::ActiveRecord.respond_to?(:gem_version) && 
  ::ActiveRecord.gem_version >= ::Gem::Version.new("5.1.0.beta1")
```

jaredbeck:
Merged, thanks.

PR #976, opened 2017-07-05
Allow `whodunnit` to be a proc

narfanator:
Basically, I've got a circumstance where I'm trying to figure out *what* is changing one of my models (it's not Sidekiq jobs, and it's not going through the controllers...).

Right now I've got:

```
class PapertrailSidekiqMiddleware
  def call(worker_class, job, queue)
    if(worker_class.class == Sidekiq::Extensions::DelayedClass)
      PaperTrail.whodunnit = YAML.load(job["args"].first).first.to_s
    else
      PaperTrail.whodunnit = worker_class.class.to_s
    end
    yield
  end
end

Sidekiq.configure_server do |config|
  config.server_middleware do |chain|
    chain.add PapertrailSidekiqMiddleware
  end
end
```

This works, but it also only covers Sidekiq; if there's no place to stick the `whodunnit` initialization like this, there's no way to make the `whodunnit` dynamic and record the actual situation.

However, with a change like this, you could set a whodunnit during the creation of the paper trail, so you could evaluate anything you like (inspecting the callstack, the originally run Ruby command, etc etc).

PS - Honestly, I was 100% convinced I wanted this feature as part of tracking down this issue, but in this moment I'm honestly no remembering what the plan was that needed it.... :P

PPS - I totally don't expect this PR to get accepted, but I wanted to start the conversation and figured this'd be more illustrative than making an issue. LMK how I should proceed!

PPPS -  I also ran into a couple of issues when first running `rspec` on Papertrail, so there's a couple of Gemfile commits to fix that.

### DIFF
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -111,6 +111,8 @@ module PaperTrail
         ensure
           paper_trail_store[:whodunnit] = previous_whodunnit
         end
+      elsif paper_trail_store[:whodunnit].respond_to?(:call)
+        paper_trail_store[:whodunnit].call
       else
         paper_trail_store[:whodunnit]
       end
--- a/spec/paper_trail_spec.rb
+++ b/spec/paper_trail_spec.rb
@@ -101,6 +101,15 @@ RSpec.describe PaperTrail do
         expect(described_class.whodunnit).to be_nil
       end
     end
+
+    context "when set to a proc" do
+      it "evaluates the proc each time a version is made" do
+        call_count = 0
+        described_class.whodunnit = proc { call_count += 1 }
+        expect(described_class.whodunnit).to eq(1)
+        expect(described_class.whodunnit).to eq(2)
+      end
+    end
   end
 
   describe ".controller_info" do
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -15,10 +15,7 @@ Please use our [bug report template][1].
 
 ## Development
 
-Install gems with `bundle exec appraisal install`, run tests with `bundle exec appraisal rake`.
-This requires ruby >= 2.0.
-(It is still possible to run the `ar-4.2` gemfile locally on ruby 1.9.3, but
-not the `ar-5.0` gemfile.)
+Install gems with `bundle exec appraisal install`.
 
 Testing is a little awkward because the test suite:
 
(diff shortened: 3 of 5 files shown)


Gemfile
```diff
@@ -1,2 +1,5 @@
 source "https://rubygems.org"
 gemspec
+
+gem "rubocop-rspec", "~> 1.15.1"
+gem "rails-controller-testing"
```
jaredbeck: Please do not edit the `Gemfile`. Please edit `Appraisals` instead. But, I don't think any changes to gems are necessary.

lib/paper_trail.rb
```diff
@@ -112,7 +112,11 @@ def whodunnit(value = nil)
           paper_trail_store[:whodunnit] = previous_whodunnit
         end
       else
-        paper_trail_store[:whodunnit]
+        if paper_trail_store[:whodunnit].class == Proc
+          paper_trail_store[:whodunnit].call
+        else
+          paper_trail_store[:whodunnit]
+        end
```
jaredbeck: Should the conditional check `respond_to?(:call)` instead of `== Proc`?

spec/paper_trail_spec.rb
```diff
@@ -101,6 +101,15 @@
...
     end
+
+    context "when set to a proc" do
+      it "evaluates the proc each time a version is made" do
+        call_count = 0
+        described_class.whodunnit = proc { call_count += 1 }
+        expect(described_class.whodunnit).to eq(1)
+        expect(described_class.whodunnit).to eq(2)
+      end
+    end
```
jaredbeck: Nice test :+1:

narfanator: :D

lib/paper_trail.rb
narfanator: Yah, that makes a lot of sense; sharp knives and all that.

Gemfile
narfanator: Kk. I'm not familiar with `Appraisals`, but I can go figure it out.

narfanator: You're correct, when run via `appraisal`, everything is fine.

narfanator:
Updated as per review and failed tests; rewrote Git history of this branch to keep it super clean.

I added a bit to the documentation where I (as a new and inexperienced contributor) was confused.

jaredbeck:
Merged. Thanks for your contribution!

PR #980, opened 2017-07-25
Remove default_initializer

yhirano55:
Tough I don't know why blank initializer file exists, it looks like really an unnecessary blank file.
So I've removed it.


jaredbeck:
This file was added by @naw in https://github.com/airblade/paper_trail/pull/854. Maybe they remember the purpose of the file? If we don't hear from them, I'm OK with deleting it. I can't find any indication that it is used.

yhirano55:
Thanks for the comment. An actual initializer file is created by [this line](https://github.com/airblade/paper_trail/blob/master/lib/generators/paper_trail/install_generator.rb#L43) in generator. Therefore It's no problem to remove the blank file on this pull request.

naw:
I'll look into it and get back to you.

naw:
I agree this is an extraneous file. I think I was originally planning on copying the initializer rather than creating it, but the contents needed to be dynamic based on the `--with-associations` option, so I had to ditch that idea. Sorry for the oversight.

yhirano55:
@jarbeck Could you please merge this request, if you don't have any problems🐧

jaredbeck:
Merged, thanks!

PR #985, opened 2017-08-24
Support belongs_to_required_by_default

joelmichael:
We're trying to upgrade our app to use the Rails 5.1 defaults, but we ran into a problem with PaperTrail. Specifically, when we try to soft delete certain records (i.e., mark them as deleted without actually deleting the row from the DB), and then try to make a change to the soft-deleted record, it raises an error saying the versions are invalid. This is because of the `belongs_to :item` on VersionConcern, which makes the association required with the new defaults. To stay compatible with the old way and allow deleting and soft deleting of records, we should make this association optional.

An unrelated commit adds the `rails-controller-testing` gem to the Gemfile. I tried adding it as a development dependency in the gemspec, but couldn't get it to work. Without this gem, a lot of the tests fail for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#985](https://github.com/airblade/paper_trail/pull/985) - Fix RecordInvalid error on nil item
+  association when belongs_to_required_by_default is enabled.
 
 ## 7.1.1 (2017-08-18)
 
--- a/spec/dummy_app/config/application.rb
+++ b/spec/dummy_app/config/application.rb
@@ -31,7 +31,12 @@ module Dummy
       if v >= Gem::Version.new("4.2") && v < Gem::Version.new("5.0.0.beta1")
         config.active_record.raise_in_transactional_callbacks = true
       end
-      if v >= Gem::Version.new("5.0.0.beta1")
+      if v >= Gem::Version.new("5.0.0.beta1") && v < Gem::Version.new("5.1")
+        config.active_record.belongs_to_required_by_default = true
+        config.active_record.time_zone_aware_types = [:datetime]
+      end
+      if v >= Gem::Version.new("5.1")
+        config.load_defaults "5.1"
         config.active_record.time_zone_aware_types = [:datetime]
       end
     end
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -12,7 +12,11 @@ module PaperTrail
     extend ::ActiveSupport::Concern
 
     included do
-      belongs_to :item, polymorphic: true
+      if ActiveRecord.gem_version >= Gem::Version.new("5.0")
+        belongs_to :item, polymorphic: true, optional: true
+      else
+        belongs_to :item, polymorphic: true
+      end
 
       # Since the test suite has test coverage for this, we want to declare
       # the association when the test suite is running. This makes it pass when
--- a/spec/dummy_app/app/models/callback_modifier.rb
+++ b/spec/dummy_app/app/models/callback_modifier.rb
@@ -20,9 +20,14 @@ class BeforeDestroyModifier < CallbackModifier
   paper_trail.on_destroy :before
 end
 
-class AfterDestroyModifier < CallbackModifier
-  has_paper_trail on: []
-  paper_trail.on_destroy :after
+if ActiveRecord.gem_version < Gem::Version.new("5") ||
+    !ActiveRecord::Base.belongs_to_required_by_default
+
+  class AfterDestroyModifier < CallbackModifier
+    has_paper_trail on: []
+    paper_trail.on_destroy :after
+  end
+
 end
 
 class NoArgDestroyModifier < CallbackModifier
--- a/spec/dummy_app/app/models/fluxor.rb
+++ b/spec/dummy_app/app/models/fluxor.rb
@@ -1,3 +1,7 @@
 class Fluxor < ActiveRecord::Base
-  belongs_to :widget
+  if ActiveRecord.gem_version >= Gem::Version.new("5.0")
+    belongs_to :widget, optional: true
+  else
+    belongs_to :widget
+  end
 end
--- a/spec/dummy_app/app/models/person.rb
+++ b/spec/dummy_app/app/models/person.rb
@@ -1,7 +1,13 @@
 class Person < ActiveRecord::Base
   has_many :authorships, foreign_key: :author_id, dependent: :destroy
   has_many :books, through: :authorships
-  belongs_to :mentor, class_name: "Person", foreign_key: :mentor_id
+
+  if ActiveRecord.gem_version >= Gem::Version.new("5.0")
+    belongs_to :mentor, class_name: "Person", foreign_key: :mentor_id, optional: true
+  else
+    belongs_to :mentor, class_name: "Person", foreign_key: :mentor_id
+  end
+
   has_paper_trail
 
   # Convert strings to TimeZone objects when assigned
--- a/spec/dummy_app/app/models/whatchamajigger.rb
+++ b/spec/dummy_app/app/models/whatchamajigger.rb
@@ -1,4 +1,9 @@
 class Whatchamajigger < ActiveRecord::Base
   has_paper_trail
-  belongs_to :owner, polymorphic: true
+
+  if ActiveRecord.gem_version >= Gem::Version.new("5.0")
+    belongs_to :owner, polymorphic: true, optional: true
+  else
+    belongs_to :owner, polymorphic: true
+  end
 end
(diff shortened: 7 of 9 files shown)


joelmichael:
Test failures are due to the CI using older versions of the gems (Rails 4.2), due to there being no Gemfile.lock.

jaredbeck:
Hi Joel, thanks for the contribution! 

- Re: The `rails-controller-testing` gem: Why do we need it? We haven't needed it before. I just did a build a week ago and it passed: https://travis-ci.org/airblade/paper_trail/builds/266018448
- Re: `belongs_to(optional:)` I have not used `optional` before and I'll have to do some research. Does it just test that `item.present?` or does it query the database? The docs are a bit unclear, they just say:

> :optional
> When set to true, the association will not have its presence validated.

joelmichael:
@jaredbeck The gem is needed for Rails 5 backwards compatibility with the controller tests, specifically the `assigns` feature which was removed.

The `optional: true` is required by Rails 5 as well, because Rails 5 makes `required: true` the default on all `belongs_to` associations.

This isn't showing up in CI because, it seems to me, Travis is including the Rails 4.2 gems instead of Rails 5.0 or 5.1.

Gemfile
```diff
@@ -1,2 +1,3 @@
 source "https://rubygems.org"
 gemspec
+gem "rails-controller-testing"
```
jaredbeck: > Re: The rails-controller-testing gem: Why do we need it? We haven't needed it before. ...

Oops, I was wrong about this, we already are using the `rails-controller-testing` gem. Please see our `Appraisals` file. No changes to our `Gemfile` are necessary.

joelmichael:
Sorry about my confusion. I removed that bit from Gemfile and added some version conditionals where necessary. Seems to pass in all versions for me locally, so I suspect it'll pass on CI soon.

joelmichael:
I updated the commit message and renamed this PR for improved clarity.

joelmichael:
Note that it now prints out this warning message in AR 5.0 and 5.1 appraisal specs:

```
DEPRECATION WARNING: paper_trail.on_destroy(:after) is incompatible with ActiveRecord's
belongs_to_required_by_default and has no effect. Please use :before
or disable belongs_to_required_by_default.
 (called from <class:AfterDestroyModifier> at /Users/joel/Code/paper_trail/spec/dummy_app/app/models/callback_modifier.rb:25)
```

jaredbeck:
Joel, thanks for addressing my concerns re: the `rails-controller-testing` gem. I know the test suite is hard to run (because we test against so many rails' and dbs) and I hope you found the contributing guide helpful. Please open a new PR if you have any suggestions for the contributing guide.

Now, let me see if I understand the changes that remain in this PR.

> We're trying to upgrade our app to use the Rails 5.1 defaults, but we ran into a problem with PaperTrail.
> Specifically, when we try to soft delete certain records (i.e., mark them as deleted without actually deleting the row from the DB), and then try to make a change to the soft-deleted record, it raises an error saying the versions are invalid. 

Do you get a `RecordInvalid` error? Are you referring to the following change, which happened in rails 5.0?

> belongs_to will now trigger a validation error by default if the association is not present. You can turn this off on a per-association basis with optional: true. Also deprecate required option in favor of optional for belongs_to. (Pull Request)
> http://guides.rubyonrails.org/5_0_release_notes.html

Earlier, I asked:

> I have not used `optional` before and I'll have to do some research. Does it [`optional: false`] just test that `item.present`? or does it query the database?

Did this question get answered?

> .. when we try to soft delete certain records (i.e., mark them as deleted without actually deleting the row from the DB), and then try to make a change to the soft-deleted record, it raises an error saying the versions are invalid. 

Please include a test that demonstrates this issue. Use an existing test model in `spec/dummy_app` if at all possible. If not, it's ok to create a new one.

> To stay compatible with the old way and allow deleting and soft deleting of records, we should make this association optional.

Based only on the release notes I quoted above, I agree that adding `optional: true` improves compatibility. I think we're on the right track.

Please add an entry to the changelog, under Unreleased -> Fixed.

Do we need to add `optional: true` to as many of our test models as you have done? I'd rather change as few tests as possible.

> Note that it now prints out this warning message in AR 5.0 and 5.1 appraisal specs:
> DEPRECATION WARNING: paper_trail.on_destroy(:after) is incompatible with ActiveRecord's
belongs_to_required_by_default and has no effect. Please use :before
or disable belongs_to_required_by_default. ..

Well that's a problem. We don't want our test suite printing any deprecation warnings. First, why is the warning printed? Clearly, because of the changes to `spec/dummy_app/config/application.rb`. Now, what is the warning telling us? Our specs are using `on_destory(:after)`, I assume. So, what options do we have? Change those specs to use `:before`? That seems easiest, but then we lose test coverage of `:after`. WDYT?

Thanks again for your work on this.

jaredbeck:
> Please include a test that demonstrates this issue.

On second thought, is it possible that *many* tests demonstrate this issue once your change to `spec/dummy_app/config/application.rb` is applied?

joelmichael:
>I know the test suite is hard to run (because we test against so many rails' and dbs) and I hope you found the contributing guide helpful.

It wasn't too hard after I figured out how to use appraisal. I didn't look at the guide until you mentioned it just now; it seems fairly straight-forward to me.

>Do you get a RecordInvalid error? Are you referring to the following change, which happened in rails 5.0?

Yes, I get that error, and I'm referring to that change.

>Did this question get answered?

I didn't answer it. According to the API documentation: "When set to `true`, the association will not have its presence validated."

>On second thought, is it possible that many tests demonstrate this issue once your change to spec/dummy_app/config/application.rb is applied?

Yes, that is the case. It's why I changed the models I did, so that the tests would pass. If you remove any of those changes, there will be test failures. That's the only reason I didn't add a new test, but I'm not totally against making a new one anyway. Just not sure if it's necessary.

>Well that's a problem. We don't want our test suite printing any deprecation warnings. First, why is the warning printed?

It's printed because we now have this configuration option enabled. I'm not really sure why the `:after` function is incompatible with it. One possible solution is to check for the configuration option prior to defining the `:after` functionality in the specs, and also before testing it. This way, PaperTrail could continue to support this functionality for older ActiveRecord versions, or for Rails 5.x users who do not have this configuration option enabled.

>Thanks again for your work on this.

I am happy to do it, and it's in my interest, as this is acting as a blocker for enabling this default in our code. Thank you for your interest in this PR.

joelmichael:
I've made some updates to this PR. To avoid the deprecation warning, it will now check the version and config setting prior to defining and testing the :after callback. I also added a description of the fix to the Changelog.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+
```
jaredbeck: Please remove blank line

CHANGELOG.md
```diff
@@ -15,7 +16,7 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- The "item" association will no longer be required when belongs_to_required_by_default is enabled.
```
jaredbeck: Please follow the convention of this file and put in a link to the PR, so people can find out more about the change if they have questions.

spec/dummy_app/config/application.rb
```diff
@@ -31,7 +31,12 @@ class Application < Rails::Application
       if v >= Gem::Version.new("4.2") && v < Gem::Version.new("5.0.0.beta1")
         config.active_record.raise_in_transactional_callbacks = true
       end
-      if v >= Gem::Version.new("5.0.0.beta1")
+      if v >= Gem::Version.new("5.0.0.beta1") && v < Gem::Version.new("5.1")
+        config.active_record.belongs_to_required_by_default = true
```
jaredbeck: Why not `load_defaults("5.0")`?

spec/dummy_app/config/application.rb
```diff
@@ -31,7 +31,12 @@ class Application < Rails::Application
...
+        config.active_record.time_zone_aware_types = [:datetime]
+      end
+      if v >= Gem::Version.new("5.1")
+        config.load_defaults 5.1
```
jaredbeck: This is bordering on annoyingly pedantic, I'm sure, but we should not use a `Float` here. `Float#to_s` is not predictable. Please use string `"5.1"`. Thanks.

```
5.1.to_s 
# => "5.1"
5.99999999999999999999999999999999999999999999999999999.to_s
=> "6.0"
```

I know, I know, it's not relevant, how likely are we to hit a floating point issue in a rails version number? But, I can't help it. :)

spec/dummy_app/config/application.rb
joelmichael: Apparently this command was added in Rails 5.1.

joelmichael:
I believe I have corrected all of the reported issues. Please let me know if anything else is necessary. Thanks!

jaredbeck:
> I believe I have corrected all of the reported issues. Please let me know if anything else is necessary. Thanks!

Looks good, thanks Joel.

PR #988, opened 2017-09-05
Fix ActiveRecord version check in Rails 4.0

magnusvk:
Rails 4.0 does not have the `ActiveRecord.gem_version` method, so this
check will fail there. However, we know that if `gem_version` is not
available, we are not dealing with Rails >= 5.0, so this will work in
both cases.

See #956 for an equivalent case elsewhere in the codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#988](https://github.com/airblade/paper_trail/pull/988) - Fix ActiveRecord
+  version check in `VersionConcern` for Rails 4.0
+
 
 ## 7.1.2 (2017-08-30)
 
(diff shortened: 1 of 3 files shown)


magnusvk:
@jaredbeck added as requested.

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 - [#985](https://github.com/airblade/paper_trail/pull/985) - Fix RecordInvalid
   error on nil item association when belongs_to_required_by_default is enabled.
+- [#988](https://github.com/airblade/paper_trail/pull/988) - Fix ActiveRecord
+  version check in `VersionConcern` for Rails 4.0
```
jaredbeck: Please put your changelog entry under Unreleased -> Fixed. Here you have it under 7.1.2 -> Fixed. Thanks.

magnusvk:
Oops. @jaredbeck fixed.

PR #990, opened 2017-09-14
Add whodunnit for with_versioning method in RSpec unit tests

lazebny:
Added ability to set whodunnit in RSpec unit tests.

Use cases: service and model tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Added
 
-- None
+- [#990](https://github.com/airblade/paper_trail/pull/990) Add whodunnit for with_versioning method in RSpec unit tests
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -1098,9 +1098,9 @@ Generates (but does not run) a migration to add a versions table.  Also generate
 
 As of version 6, PT no longer supports rails 3 or the [protected_attributes][17]
 gem. If you are still using them, you may use PT 5 or lower. We recommend
-upgrading to [strong_parameters][18] as soon as possible.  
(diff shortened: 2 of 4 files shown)


jaredbeck:
Hi Vadim, are you still working on this? Want me to keep it open?

lazebny:
Hey, thank you for feedback and sorry I forgot about this pr. I'll try to complete your recommendations at weekend.

lazebny:
@jaredbeck i need you help,
If we don't use threads in our units than Thread.current works fine (spec covers it), but with threads it don't:
```
# code
Thread.current[:name] = 'test';
Thread.new { p "From child: #{Thread.current[:name]}" }.join;
p "From current: #{Thread.current[:name]}";

# output
"From child: "
"From current: test"
```
I think that for unit tests is better to have whodunnit globally but it can break paper_trail behaviour on a request level.
Could you suggest what we can do with this?

jaredbeck:
Hi Vadim, thanks for adding the changelog entry and test.

> ::PaperTrail.enabled= affects all threads, but ::PaperTrail.whodunnit= only affects the current thread (see RequestStore). Does this matter?

I'm worried that people will be confused if `with_versioning` sets both `enabled` and `whodunnit`. We will get a bug report saying that `with_versioning` "works for enabled but doesn't work for whodunnit".

Options:

1. We can close this PR, and you can have your own method that does this.
2. We can have another method in PT like `with_whodunnit`, but the nesting might look ugly

```
with_versioning do
  with_whodunnit("Vadim") do
    # ..
```

3. We can try to clarify the issue with argument naming:

```
with_versioning(this_thread: { whodunnit: "Vadim" }) do
  # ..
```

What do you think? Any other ideas?

lazebny:
Hi Jared, I think the simpler solution is to close this PR :) Because other approaches make dsl a bit complicated.

PR #994, opened 2017-09-20
Drop support for ruby 2.1

jaredbeck:
Ruby 2.1 reached EoL on 2017-04-01.
https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

PT continued to support it for 6 months after EoL.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,7 +15,7 @@ AllCops:
     - spec/dummy_app/db/schema.rb # Generated, out of our control
 
   # Set to lowest supported version
-  TargetRubyVersion: 2.1
+  TargetRubyVersion: 2.2
 
 Layout/AlignParameters:
   EnforcedStyle: with_fixed_indentation
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: ruby
 cache: bundler
 rvm:
-  - 2.3.4
-  - 2.1.10
+  - 2.4.2
+  - 2.2.8
 env:
   global:
     - TRAVIS=true
@@ -31,11 +31,6 @@ gemfile:
 
 matrix:
   fast_finish: true
-  exclude:
-    - gemfile: gemfiles/ar_5.0.gemfile
-      rvm: 2.1.10
-    - gemfile: gemfiles/ar_5.1.gemfile
-      rvm: 2.1.10
 
 addons:
   postgresql: "9.4"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Breaking Changes
 
+- Drop support for ruby 2.1, whose EoL was 2017-04-01
 - [#803](https://github.com/airblade/paper_trail/issues/803) -
   where_object_changes no longer supports reading json from a text column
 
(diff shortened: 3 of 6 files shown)


jaredbeck:
In https://github.com/airblade/paper_trail/pull/995 Ben asked:

> Is there something that Ruby 2.1 breaks, just out of curiosity?

Good question, and I'm glad you're asking it :) The way I phrase the question is, what burden does supporting an EoL ruby put on the library? Then, we can weigh that against the needs of people using EoL rubies.

First, supporting ruby 2.1 makes testing slightly more difficult because rails no longer supports ruby 2.1, as of rails 5. TravisCI makes this easy though, so that's only a slight burden.

Second, I'd like to eventually use the new features of ruby, like `Hash#dig`, frozen string literals, and the safe navigation operator. (Those are all from 2.3. Ruby 2.2 and 2.4 are mostly performance improvements). Supporting EoL rubies prevents us from using these new features.

People using EoL rubies can always continue to use PT <= 7, and we can continue to backport bug fixes to the 7 series. I'm happy to do that.

Basically, I'm happy to support EoL rails and ruby versions for a little while (6 months to a year) but I'd argue for a policy where we eventually drop support.

batter:
I was just saying, that I don't know that we necessarily need to officially support it, however, leaving the `required_ruby_version` minimum threshold at `2.1` likely doesn't hurt, and I'm guessing people could still continue using the modern versions without much problem.  That being said, I know that some of the modern versions of `ActiveRecord` / rails, which this gem depends on, also require more modern versions of Ruby, so perhaps we should match what they do in terms of required ruby version?

jaredbeck:
> .. leaving the required_ruby_version minimum threshold at 2.1 likely doesn't hurt ..

I agree that we could probably get away with `required_ruby_version("2.1")` for a while, even if we're not testing against 2.1. I'd prefer not to, because I would be concerned about claiming a level of compatibility that we don't test for, but if you want to it's OK with me.

In the future, when we begin using ruby 2.3 syntax (e.g. safe nav. operator) we will have to specify `required_ruby_version("2.3")`.

> That being said, I know that some of the modern versions of ActiveRecord / rails, which this gem depends on, also require more modern versions of Ruby, so perhaps we should match what they do in terms of required ruby version?

Makes sense to me. :+1: PT is a rails plugin, so it makes sense to track the same versions of ruby that rails depends on, within reason. One exception: rails 4.2, which is still living, supports ruby 1.9.3 and we already dropped support for 1.9.3. But otherwise, I think that policy makes sense going forward.

batter:
> I agree that we could probably get away with required_ruby_version("2.1") for a while, even if we're not testing against 2.1. I'd prefer not to, because I would be concerned about claiming a level of compatibility that we don't test for, but if you want to it's OK with me.

You know what, I actually don't really care that much either way. Historically my opinion would be to say leave it unless we know something's broken, but on the other hand if someone wants to use an older ruby version they can always fork the gem and change it themselves and the specification will at least have warned them there could be an issue.

> In the future, when we begin using ruby 2.3 syntax (e.g. safe nav. operator) we will have to specify required_ruby_version("2.3").

Agreed, and that is a nice operator that they've added 😄 

👍  to everything else

jaredbeck:
> .. if someone wants to use an older ruby version they can always fork the gem and change it themselves ...

Or just continue to use PT 7.

> 👍 to everything else

Cool, thanks.

PR #995, opened 2017-09-20
Drop support for rails 4.0 and 4.1

jaredbeck:
EoL for both was 2016-06-30
http://weblog.rubyonrails.org/2016/6/30/Rails-5-0-final/

PT continued to support them for 15 months after EoL.


aried3r:
Will this result in a 8.0.0 release of paper_trail?

jaredbeck:
> Will this result in a 8.0.0 release of paper_trail?

Yes, the release that includes this change will be a major version bump. There are three breaking changes lined up for 8.0.0 (see changelog under "Unreleased"). If you have other suggestions for breaking changes, now's the time :)

batter:
Is there something that Ruby 2.1 breaks, just out of curiosity?

jaredbeck:
> Is there something that Ruby 2.1 breaks, just out of curiosity?

Hi Ben, Good question, and I'm glad you're asking it :) I'll answer over at https://github.com/airblade/paper_trail/pull/994 as this PR is for dropping rails 4.0 and 4.1 support.

Issue #803, opened 2016-05-11
where_object_changes returns more records than expected


jaredbeck:
The readme says:

> Known issue: inconsistent results for numeric values due to limitations of SQL wildcard matchers against the serialized objects.

I asked Ben about this in https://github.com/airblade/paper_trail/pull/801/files

> > does this "known issue" still exist, re: inconsistent results for numeric values? In the YAML serializer, the value is succeeded by a newline, and in the JSON serializer, care has been taken to succeed the value by a comma or brace.
> 
> Yes I believe this issue does still exist. I'm not super knowledgable on wildcard matchers, but I did some investigation and this dates back to 1ad7838, which I believe shows evidence of issues when you are using numbers such as 10, 100, 1000, etc, since I think certain flavors of SQL have difficulty matching on those values.

What should our next step be in troubleshooting this? Can we get a failing test by reverting 1ad7838, or should we write a new test?

jaredbeck:
## Bug

`with_object_changes` matches more records than it should when json is stored in a text column. This is not the default configuration, and there's no way to know how few people are affected by this bug.

## Reproduction

This bug can be reproduced in the existing test suite by reverting part of 1ad78383a.

```diff
# spec/models/version_spec.rb
+          let(:int) { 1 }
-          let(:int) { rand(5) + 2 }
```

This patch will cause `with_object_changes` to return more records than expected and the tests will fail.

## Analysis

Consider `where_object_changes(an_integer: 1)`, which generates the following SQL:

```sql
"versions"."object_changes" ILIKE '%"an_integer":[1,%' 
OR "versions"."object_changes" ILIKE '%"an_integer":[%,1]%'
```

See the problem yet? It's in the second half of the disjunction. Think about what those wildcards might match. What about a version record with `object_changes` like this?

```plain
{"name":[null,"Retta"],"an_integer":[null,0],"id":[null,1]}
```

We don't want to match this record. We don't want `where_object_changes(an_integer: 1)` to return this record. You see why it does though, right? That `%` wildcard matches everything between `"an_integer":[` and the `1]` at the very end.

## Suggestions

I don't know how to fix this, so I might add a runtime warning to the method asking people to come here and join the discussion. Here are some ideas to get the conversation going.

### Have `where_object_changes` raise an error if `object_changes` is a text column

The simplest solution. I think it works fine for `json` and `jsonb` columns, so there's no need to remove the `where_object_changes` method entirely.

### A Fix for Postgres only, using `::json->`

[...]

PR #997, opened 2017-09-29
clarify the issue with `where_object_changes` for numeric values

sergey-alekseev:
History
---
[The first problem mention](https://github.com/airblade/paper_trail/commit/1ad78383aa58d6b74e237f84310f7440712710ab).
[The major issue](https://github.com/airblade/paper_trail/issues/803) with relevant discussions.

Details
---
I see the problem only with one specific case – when the method is `where_object_changes` and the serializer is YAML. Which is a default configuration.

I see there are 6 types of queries with a numeric value for `where_object_changes` and `where_object` (grabbed from `version_spec.rb`):
```sql
# where_object_changes
# 1
SELECT "versions".* FROM "versions" WHERE "versions"."item_id" = 1 AND "versions"."item_type" = 'Widget' AND (object_changes @> '{"an_integer":[1]}')  ORDER BY "versions"."created_at" ASC, "versions"."id" ASC

# 2
SELECT "versions".* FROM "versions" WHERE "versions"."item_id" = 1 AND "versions"."item_type" = 'Widget' AND ("versions"."object_changes" ILIKE '%
an_integer:
- 1
%' OR "versions"."object_changes" ILIKE '%
an_integer:
-%
- 1
%')  ORDER BY "versions"."created_at" ASC, "versions"."id" ASC

# 3
SELECT "versions".* FROM "versions" WHERE "versions"."item_id" = 1 AND "versions"."item_type" = 'Widget' AND (((object_changes->>'an_integer' ILIKE '[1,%') OR (object_changes->>'an_integer' ILIKE '[%,1]%')))  ORDER BY "versions"."created_at" ASC, "versions"."id" ASC

# where_object
# 4
SELECT "versions".* FROM "versions" WHERE (object @> '{"an_integer":1}')

# 5
SELECT "versions".* FROM "versions" WHERE (object->>'an_integer' = '1')

# 6
SELECT "versions".* FROM "versions" WHERE ("versions"."object" ILIKE '%
an_integer: 1
%')
```

The only problematic one is `#2`. It incorrectly matches `object_changes` like
```yaml
---
name:
-
- Kimberli
an_integer:
-
- 0
created_at:
-
- 2017-09-28 18:30:13.369889000 Z
updated_at:
-
- 2017-09-28 18:30:13.369889000 Z
id:
-
- 1
---
name:
- foobar
- Hayes
an_integer:
- 77
- 1
updated_at:
- 2017-09-28 18:30:13.383966000 Z
- 2017-09-28 18:30:13.395539000 Z
```

@batter @jaredbeck correct me if I'm wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Added
 
-- None
+- [#997](https://github.com/airblade/paper_trail/pull/997)
+  Deprecate `where_object_changes` when reading YAML from a text column
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -248,9 +248,11 @@ version.index
 version.event
 
 # Query the `versions.object` column (or `object_changes` column), by
-# attributes, using the SQL LIKE operator. Known issue: inconsistent results for
-# numeric values due to limitations of SQL wildcard matchers against the
-# serialized objects.
+# attributes, using the SQL LIKE operator.
+#
+# Known issue: `where_object_changes` with the default YAML serializer and
+# an `object_changes` text column may return incorrect results for numeric values
+# due to limitations of SQL wildcard matchers against the serialized objects.
 PaperTrail::Version.where_object(attr1: val1, attr2: val2)
 PaperTrail::Version.where_object_changes(attr1: val1)
 ```
--- a/lib/paper_trail/serializers/yaml.rb
+++ b/lib/paper_trail/serializers/yaml.rb
@@ -4,6 +4,13 @@ module PaperTrail
   module Serializers
     # The default serializer for, e.g. `versions.object`.
     module YAML
+      E_WHERE_OBJ_CHANGES = <<-STR.squish.freeze
+        where_object_changes has a known issue. When reading YAML from a text
+        column, it may return more records than expected. Instead of a warning,
+        this method may raise an error in the future. Please join the discussion
+        at https://github.com/airblade/paper_trail/pull/997
+      STR
+
       extend self # makes all instance methods become module methods as well
 
       def load(string)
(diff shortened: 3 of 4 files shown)


jaredbeck:
Hi Sergey, Thanks for the contribution!

It sounds like where_object_changes+YAML has the same issue as where_object_changes+JSON-in-text-column, as described in https://github.com/airblade/paper_trail/issues/803. I think the same solution (https://github.com/airblade/paper_trail/pull/993) would be appropriate here. 

What do you think?

jaredbeck:
> It sounds like where_object_changes+YAML has the same issue as where_object_changes+JSON-in-text-column, as described in #803. I think the same solution (#993) would be appropriate here.

To clarify, I am suggesting that we raise an error when someone is storing YAML in a `text` column and they call `where_object_changes`. Sound good?

sergey-alekseev:
@jaredbeck I think applying a "solution" like #979 + #993 is better than not applying it.  

Ideally such cases should be fixed, but I'm not interested in a text column.  

---
FYI I fixed a failed spec example by changing   
```ruby
let(:int) { column_datatype_override ? 1 : rand(5) + 1 }
```

to
```ruby
let(:int) { column_datatype_override ? 1 : rand(5) + 2 }
```

Initially I didn't take into account `rand(5) + 1` may return 1 :flushed:

jaredbeck:
>> It sounds like where_object_changes+YAML has the same issue as where_object_changes+JSON-in-text-column, as described in #803. I think the same solution (#993) would be appropriate here.

>> To clarify, I am suggesting that we raise an error when someone is storing YAML in a `text` column and they call `where_object_changes`. Sound good?

> @jaredbeck I think applying a "solution" like #979 + #993 is better than not applying it.

OK, raising an error is a breaking change, so we should start with a deprecation warning , just like we did in #979. If you agree, will you please add such a warning, Sergey? Thanks!

sergey-alekseev:
@jaredbeck I added a deprecation warning. Let me know if you think commits should be squashed.

jaredbeck:
Looks perfect! Thanks Sergey.

jaredbeck:
We'll wait a few months after this deprecation warning is released to give people a chance to discuss the issue here. Then, if no concerns are raised we will replace the warning with an error.

PR #1003, opened 2017-10-19
Check for presence of Rails.env not just ::Rails

WojtekKruszewski:
Sometimes Rails module is defined without Rails being loaded (e.g. rails-html-sanitizer defines this module). Instead let's check for Rails.env.

Solution taken from rails/activeresource#216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1003](https://github.com/airblade/paper_trail/pull/1003) - detect Rails
+  presence with `Rails.application` 
 
 ## 8.0.0 (2017-10-04)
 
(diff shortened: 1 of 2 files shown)


WojtekKruszewski:
One item in the test matrix failed (https://travis-ci.org/airblade/paper_trail/builds/289990814). It's passing on my machine - I'm guessing it's a random error.

jaredbeck:
CI job #1516.11 restarted

jaredbeck:
https://github.com/rails/activeresource/pull/216 checks for `Rails.application`, but this PR checks for `Rails.env`. Any difference?

WojtekKruszewski:
Ah, I saw `Rails.env` used in some other project. Let's go with `Rails.application` since that's what ActiveResource uses.

jaredbeck:
Closing via https://github.com/airblade/paper_trail/pull/1006 which preserves your authorship, and adds a warning instead of silently failing to load.

PR #1005, opened 2017-10-23
Load activesupport early so we can squish big strings

jaredbeck:
PT usually loads AS, unless something goes really wrong;
we're just loading it earlier.

This is not a new dependency.


jaredbeck:
HULK SQUISH HEREDOCS! 💪

PR #1006, opened 2017-10-23
Print warning when PT cannot be loaded ..

jaredbeck:
.. because rails is not loaded yet.

In some situations, when loading PT, `::Rails` has been defined,
perhaps by a library like rails-html-sanitizer, but rails itself
has not been loaded yet.

Supersedes https://github.com/airblade/paper_trail/pull/1003

### DIFF
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -14,6 +14,12 @@ require "paper_trail/serializers/yaml"
 # An ActiveRecord extension that tracks changes to your models, for auditing or
 # versioning.
 module PaperTrail
+  E_RAILS_NOT_LOADED = <<-EOS.squish.freeze
+    PaperTrail has been loaded too early, before rails is loaded. This can
+    happen when another gem defines the ::Rails namespace, then PT is loaded,
+    all before rails is loaded. You may want to reorder your Gemfile, or defer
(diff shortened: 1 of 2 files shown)


jaredbeck:
@WojtekKruszewski Hi Wojtek, please review, thanks.

lib/paper_trail.rb
```diff
@@ -190,8 +196,14 @@ def version
...
 # Require frameworks
-if defined?(::Rails) && ActiveRecord::VERSION::STRING >= "3.2"
-  require "paper_trail/frameworks/rails"
+if defined?(::Rails)
+  # Rails module is sometimes defined by gems like rails-html-sanitizer
+  # so we check for presence of Rails.application.
+  if defined?(::Rails.application)
+    require "paper_trail/frameworks/rails"
+  else
+    ::Kernel.warn(E_RAILS_NOT_LOADED)
```
WojtekKruszewski: Missing namespace, should be: `::Kernel.warn(PaperTrail::E_RAILS_NOT_LOADED)`

jaredbeck: Good catch, thanks.

jaredbeck:
Released as 8.0.1

PR #1009, opened 2017-11-16
End with newline created config/initializers/paper_trail.rb

Yuki-Inoue:
Currently, config/initializers/paper_trail.rb file created by
    rails generate paper_trail:install
does not end with newline.
Such text files may result in trouble because in unix, because text files
generally should end with newline.

So, make it put newline at the end.


jaredbeck:
Looks good. Please add an entry to the changelog under Unreleased -> Fixed. Thanks!

Yuki-Inoue:
@jaredbeck Done!  Please check if it is done correctly!

jaredbeck:
Thanks!

PR #1016, opened 2017-12-03
Discussing the API for request variables eg. whodunnit, controller_info

AlexEvanczuk:
I originally wanted to just have `controller_info` be able to take a block like `whodunnit`, and realized that this would be a good opportunity to remove some of the duplication between the two, and allow arbitrary settings to accept a block.  Let me know what you think.  I'm not attached to the idea of the private `config_handler`, and can put that logic back into the methods if you think it would be better.


AlexEvanczuk:
I spent some more time looking through the code base.  I wonder if you think it would be better to have a 
`PaperTrail::RequestStoreManager` that handles all interactions with the request store.  It could be a private class that is delegated to by the public `PaperTrail` API.  Basically wherever we use `paper_trail_store` could use this class.  It might not be necessary yet, but just a thought.

jaredbeck:
> I originally wanted to just have controller_info be able to take a block like whodunnit,  and realized that this would be a good opportunity to remove some of  the duplication between the two, and allow arbitrary settings to accept a  block.

I like this idea. Let's discuss it. Our request_store has the following keys:

| *key* | *versions table* | *api* |
| --- | --- | --- |
| :request_enabled_for_controller | no | public |
| :"enabled_for_#{model}" | no | public |
| :whodunnit | yes | public |
| :controller_info | yes | public |
| :transaction_id | yes | private |

Some keys are data, to be saved in the `versions` table, others are configuration. Most are public, but `:transaction_id` is private.  The only thing they have in common is that they affect the current request.

>  I wonder if you think it would be better to have a
PaperTrail::RequestStoreManager that handles all interactions with the request store.

I like this idea too. It could simply be called `PaperTrail::Request`.

>  It could be a private class that is delegated to by the public PaperTrail API. 

Yes, delegation will be necessary to prevent breaking changes. I'm picturing an API like this:

```
PaperTrail.request.enabled?
PaperTrail.request.whodunnit = 'Alex'
PaperTrail.request = PaperTrail::Request.new(
  whodunnit = 'Alex',
  controller_info = {}
}
PaperTrail.request(whodunnit: 'Alex', controller_info: {}) do
  # ...
end
```

This would be the preferred API and the old methods could eventually be deprecated. What do you think?

Initial code review, just trivial stuff for now:

- Please don't change version_number.rb. I'll change it when I do the release.
- Keep the changelog entry small, a single sentence if possible. The docs should go in various places in the readme.

AlexEvanczuk:
Good ideas -- I'll update this PR with the discussed changes hopefully within the next week and we can take it from there.

AlexEvanczuk:
@jaredbeck I decided to go ahead and take a first pass at this to get your thoughts before polishing the API and adding specs.

I did not change any existing specs yet, except to get rid of deprecation warnings.

If you like this approach, I'll use the `request_key_handler` or something like it on every method and clean up the new class's internal implementation, as well as make `PaperTrail.request(options) do; ... end;` functional.

AlexEvanczuk:
All tests are passing -- by the way it looks like there is a bit of flakiness in the test suite.

lib/paper_trail.rb
```diff
@@ -47,32 +48,46 @@ def enabled?
...
+      controller_info=
+      controller_info
+      transaction_id
+      transaction_id=
+    ].freeze
+
+    def method_missing(method, *args, &block)
+      if REQUEST_STORE_DEPRECATED_METHODS.include?(method)
+        msg = format("Use paper_trail.request.%s instead of paper_trail.%s", method, method)
+        ::ActiveSupport::Deprecation.warn(msg, caller)
```
jaredbeck: I know I said " .. the old methods could eventually be deprecated" but let's not do it yet.

lib/paper_trail.rb
```diff
@@ -88,61 +103,6 @@ def timestamp_field=(_field_name)
       raise(E_TIMESTAMP_FIELD_CONFIG)
     end
 
-    # Sets who is responsible for any changes that occur. You would normally use
-    # this in a migration or on the console, when working with models directly.
-    # In a controller it is set automatically to the `current_user`.
-    # @api public
-    def whodunnit=(value)
-      paper_trail_store[:whodunnit] = value
-    end
```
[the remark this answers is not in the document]
AlexEvanczuk: 👍 okay

lib/paper_trail/request.rb
```diff
@@ -0,0 +1,161 @@
+require "singleton"
```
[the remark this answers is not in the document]
AlexEvanczuk: This shouldn't  be there.

lib/paper_trail.rb
AlexEvanczuk: :+1:

AlexEvanczuk:
Addressed your comments:
- [x] Using plain ruby
- [x] Left the deprecation as a comment rather than as an actual deprecation in the code
- [x] Removed unnecessary require

Let me know if you think the general approach is good and I'll go ahead and move the tests over from `paper_trail_spec` to request (and have the ones in `paper_trail_spec` simply test that we are delegating to `request`.

CHANGELOG.md
```diff
@@ -21,6 +21,21 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 - None
 
+### Fixed
+
+- None
+
+## 8.1.1 (2017-12-02)
```
jaredbeck: Please put your changelog entry under "unreleased". Do not specify a version number. I'll set the version number when I do the release.

README.md
```diff
@@ -11,6 +11,7 @@ has been destroyed.
 | Version        | Documentation |
 | -------------- | ------------- |
 | Unreleased     | https://github.com/airblade/paper_trail/blob/master/README.md |
+| 8.1.1          | https://github.com/airblade/paper_trail/blob/v8.1.1/README.md |
```
jaredbeck: Please do not edit this table.

lib/paper_trail.rb
```diff
@@ -47,32 +48,46 @@ def enabled?
...
+    # @api public
+    def request(options = nil)
+      if options
+        old_request = @request
+        @request ||= PaperTrail::Request.new
+        yield
+        @request = old_request
+      else
+        @request ||= PaperTrail::Request.new
+      end
```
jaredbeck: I don't think it makes sense use an instance variable here. There's nothing to store. The `Request` instance doesn't have any data of its own, it's just an interface to a `request_store`.

lib/paper_trail/request.rb
```diff
@@ -0,0 +1,152 @@
...
+      end
+
+      begin
+        yield
+      ensure
+        previous_config.each do |config_key, config_value|
+          paper_trail_store[config_key] = config_value
+        end
+      end
+    end
```
jaredbeck: Do we need both this *and* `PaperTrail.request({}) do`?

lib/paper_trail/request.rb
```diff
@@ -0,0 +1,152 @@
...
+    #   PaperTrail.request.controller_info({ ip: '127.0.0.1' }) do
+    #     puts PaperTrail.request.controller_info # => { ip: '127.0.0.1' }
+    #   end
+    #
+    # @api public
+    def controller_info(value = nil)
+      config_handler(:controller_info, value) do
+        yield
+      end
+    end
```
jaredbeck: I'm not certain we want to add block support to all of these individual config options, given that we have `PaperTrail.request({}) do`

lib/paper_trail/request.rb
AlexEvanczuk: I was thinking that `PaperTrail.request({}) do` would just delegate to this method.  Ideally with this new API, people would actually just call the paper trail request class API directly rather than through paper trail, so maybe `PaperTrail.request(...)` doesn't make sense as a public API, and instead they should do `PaperTrail.request.with_config({...})`.  What do you think?

lib/paper_trail/request.rb
AlexEvanczuk: Okay -- so you're thinking just leave the `whodunnit` with block support and the have the rest use the old behavior, but can be set with blocks using the new method?  That sounds good to me.  Under the hood I can just have `whodunnit` call the new method as well for readability.

jaredbeck:
Alex, I think I'm going to open a new PR with what I'd like to see, based on your work here (preserving your authorship). I think that'll be easier than trying to specify, in detail here, the API I'm imagining.

AlexEvanczuk:
Sounds good!

PR #1017, opened 2017-12-03
Add sidekiq tagging middleware

AlexEvanczuk:
Base branch: 
https://github.com/airblade/paper_trail/pull/1016/files


jaredbeck:
Thanks for the feature suggestion Alex. I think it will be helpful to sidekiq users!

This year, I've been working to keep PT simple by moving integration into other gems.

- sinatra controller integration was moved to [paper_trail-sinatra](https://github.com/jaredbeck/paper_trail-sinatra)
- [globalize-versioning](https://github.com/globalize/globalize-versioning) provides an interface between globalize and PT
- [paper_trail-globalid](https://github.com/ankit1910/paper_trail-globalid) enhances `whodunnint` by adding an `actor` method ".. to instances of PaperTrail::Version that will return you the ActiveRecord object who was responsible for change."
- a few others

So, my first instinct here is to suggest that you create a new gem for this. The correct name for this gem would be `paper_trail-sidekiq`, according to rubygems.org (http://guides.rubygems.org/name-your-gem/)

Sound good?

AlexEvanczuk:
Makes sense, I think it's a good call!

jaredbeck:
>> So, my first instinct here is to suggest that you create a new gem for this.

> Makes sense, I think it's a good call!

Great. I'll close this. If there's anything you need in PT itself to make integration easier, please open a new PR. (I'm planning to review https://github.com/airblade/paper_trail/pull/1016 next and I'm wondering if it's related)

When your new gem is published, please add a link to our readme under [8. Integration with Other Libraries](https://github.com/airblade/paper_trail#8-integration-with-other-libraries)

PR #1018, opened 2017-12-04
Issue 015/serializing postgres arrays

hubertpompecki:
This PR addresses https://github.com/airblade/paper_trail/issues/1015

Starting from Rails version 5.0.2 the default serializer of PostgreSQL columns returns an `ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Array` object instead of a string. This new object is not suitable for JSON encoding and breaks versioning of any array fields backed by Postgres.

Whenever a PostgreSQL array is used, instead of asking Active Record for a serializer we introduce our own, which simply returns the underlying array without any modifications.


lib/paper_trail/attribute_serializers/array_serializer.rb
```diff
@@ -0,0 +1,28 @@
+module PaperTrail
+  module AttributeSerializers
+    # :nodoc:
```
jaredbeck: Please provide documentation for all new classes.

spec/dummy_app/db/migrate/20110208155312_set_up_test_tables.rb
```diff
@@ -61,6 +61,15 @@ def up
       t.timestamps null: true
     end
 
+    if ENV["DB"] == "postgres"
+      create_table :postgres_users, force: true do |t|
+        t.string     :name
+        t.integer    :post_ids,    array: true
+        t.datetime   :login_times, array: true
```
jaredbeck: I like how you are testing two different types of arrays here. Also, it looks like this is the first test table to have arrays at all! Kind of surprising.

spec/paper_trail/attribute_serializers/object_attribute_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "spec_helper"
+
+RSpec.describe PaperTrail::AttributeSerializers::ObjectAttribute do
```
jaredbeck: Please use the nested-module style.

```
module PaperTrail
  module AttributeSerializers
    RSpec.describe ObjectAttribute
```

spec/spec_helper.rb
```diff
@@ -2,6 +2,7 @@
 ENV["DB"] ||= "sqlite"
 
 require "byebug"
+require "support/database_specific_specs"
```
jaredbeck: Please create a separate PR for this. It's a good idea, but I'd like to handle it separately, and more comprehensively. For this PR, please use plain ruby conditionals the way other tests currently do.

lib/paper_trail/attribute_serializers/attribute_serializer.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module PaperTrail
+  module AttributeSerializers
+    # Values returned by some Active Record serializers are
+    # not suited for JSON encoding. This factory
+    # replaces certain default Active Record serializers
+    # with custom PaperTrail ones.
+    module AttributeSerializer
+      def self.for(klass, attr)
+        case active_record_serializer = klass.type_for_attribute(attr)
```
jaredbeck: Please do not use the assignment operator in a `case`.

lib/paper_trail/attribute_serializers/attribute_serializer.rb
```diff
@@ -0,0 +1,22 @@
...
+        when ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Array then
+          ArraySerializer.new(
+            active_record_serializer.subtype,
+            active_record_serializer.delimiter
+          )
+        else active_record_serializer
```
jaredbeck: Please move this onto its own line:

```
else
  active_record_serializer
end
```

lib/paper_trail/attribute_serializers/attribute_serializer.rb
```diff
@@ -0,0 +1,22 @@
+require "paper_trail/attribute_serializers/array_serializer"
+
+module PaperTrail
+  module AttributeSerializers
+    # Values returned by some Active Record serializers are
+    # not suited for JSON encoding. This factory
+    # replaces certain default Active Record serializers
+    # with custom PaperTrail ones.
+    module AttributeSerializer
```
jaredbeck: I see what you were going for with this name. `AttributeSerializer.for` does read nicely, but I'd think I'd prefer `AttributeSerializerFactory.for`. It doesn't read as well, but you can tell the purpose of the module from its name alone.

lib/paper_trail/attribute_serializers/attribute_serializer.rb
```diff
@@ -0,0 +1,22 @@
+require "paper_trail/attribute_serializers/array_serializer"
+
+module PaperTrail
+  module AttributeSerializers
+    # Values returned by some Active Record serializers are
+    # not suited for JSON encoding. This factory
+    # replaces certain default Active Record serializers
+    # with custom PaperTrail ones.
```
jaredbeck: A good, helpful comment. I'd change "JSON encoding" to "writing JSON to a text column.".

This phrasing disabiguates JSON-in-text-column from native postgres json(b).

lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
```diff
@@ -32,15 +34,15 @@ def defined_enums
       # This implementation uses AR 5's `serialize` and `deserialize`.
       class CastAttributeSerializer
         def serialize(attr, val)
-          @klass.type_for_attribute(attr).serialize(val)
+          PaperTrail::AttributeSerializers::AttributeSerializer.for(@klass, attr).serialize(val)
```
jaredbeck: It's not necessary to fully-qualify the constant here. `AttributeSerializer` alone is sufficient.

lib/paper_trail/attribute_serializers/array_serializer.rb
```diff
@@ -0,0 +1,28 @@
...
+    class ArraySerializer
```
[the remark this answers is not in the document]
hubertpompecki: Good point and I'd also lean towards no. 1.

Given that the original implementation obtains an AR serializer by calling `#type_for_attribute(attr)` then perhaps `PaperTrail::TypeSerializers::PostgresArraySerializer` would be a better name?

jaredbeck:
Thanks for your work on this Hubert!

PR #1019, opened 2017-12-07
Improves performance for large datasets by outsourcing subselects

Quantenradierer:
On a dataset of ~300 000 entries, one query took about 20 seconds, which could be reduced, with the committed changes, to less than one ms.
This solution may cause problems when the subselect returns a large set of indices, therefore I tested this:
On the same dataset an query "SELECT * FROM versions WHERE ID in (30 000 different ids)" took about 112ms, but it was a direct sql statement so network latency is not included. But even if it were, it's still better than 20 seconds.

However, if the subselect returns 30 000 different ids, the versions table may be much larger than 300 000 entries and therefore even slower when using an subselect.

### DIFF
--- a/lib/paper_trail/reifiers/has_many.rb
+++ b/lib/paper_trail/reifiers/has_many.rb
@@ -81,9 +81,10 @@ module PaperTrail
         # @return A `Hash` mapping IDs to `Version`s
         #
         def versions_by_id(klass, version_id_subquery)
+          ids = version_id_subquery.map(&:min_id).reject(&:nil?)
           klass.
             paper_trail.version_class.
-            where("id IN (#{version_id_subquery})").
+            where("id IN (?)", ids).
             inject({}) { |a, e| a.merge!(e.item_id => e) }
         end
 
@@ -95,13 +96,12 @@ module PaperTrail
         def load_versions_for_hm_association(assoc, model, version_table, tx_id, version_at)
           version_id_subquery = ::PaperTrail::VersionAssociation.
             joins(model.class.version_association_name).
-            select("MIN(version_id)").
+            select("MIN(version_id) as min_id").
             where("foreign_key_name = ?", assoc.foreign_key).
             where("foreign_key_id = ?", model.id).
             where("#{version_table}.item_type = ?", assoc.class_name).
             where("created_at >= ? OR transaction_id = ?", version_at, tx_id).
-            group("item_id").
-            to_sql
+            group("item_id")
           versions_by_id(model.class, version_id_subquery)
         end
       end
--- a/lib/paper_trail/reifiers/has_many_through.rb
+++ b/lib/paper_trail/reifiers/has_many_through.rb
@@ -72,7 +72,7 @@ module PaperTrail
         # @api private
         def load_versions_for_hmt_association(assoc, ids, tx_id, version_at)
           version_id_subquery = assoc.klass.paper_trail.version_class.
-            select("MIN(id)").
+            select("MIN(id) as min_id").
             where("item_type = ?", assoc.class_name).
             where("item_id IN (?)", ids).
             where(
@@ -80,8 +80,7 @@ module PaperTrail
               version_at,
               tx_id
             ).
-            group("item_id").
-            to_sql
+            group("item_id")
           Reifiers::HasMany.versions_by_id(assoc.klass, version_id_subquery)
         end
       end


jaredbeck:
Thanks for the PR, Bernd! Please provide a benchmark using our [bug report template](https://github.com/airblade/paper_trail/blob/master/doc/bug_report_template.rb). I recommend using ruby's `Benchmark` module.

Quantenradierer:
It seems to be a MySQL Problem. I already tested it with different databases, but my following results are only for mysql. I want to test it with the same dataset but it takes time.

I will add Postgres and SQLite3 results tomorrow.

## configuration 1
```
      physicians = 100
      patients = 100
      appointments = 100_0

      changes_per_transaction = 10
      transactions = 10_000
```
329_931 entries in version, 219_817 entries in version_associations

### MySQL
c0a90c2892f922b93179041857706cbb5668a78f
Patient has_many: 1.246860933334877 has_many_through: 0.012706000241450965
Physician has_many: 1.2306154333346058 has_many_through: 0.01338899993384257

9eedb4210414383e0ac869974b0c1775c6e99131 (my changes)
Patient has_many: 0.20238596666798306 has_many_through: 0.012047999887727201
Physician has_many: 0.17144210000211993 has_many_through: 0.01249700017767027

## configuration 2

```
      physicians = 10
      patients = 10_000
      appointments = 10_000

      changes_per_transaction = 10
      transactions = 10_000
```
Lesser changes for appointments, still a high amount of changes for physicians. Also each physician is related to a lot of patients over the appointments.

c0a90c2892f922b93179041857706cbb5668a78f
Patient has_many: 0.9767984333292892 has_many_through: 0.013881999999284744
Physician has_many: 4.128579433338018 has_many_through: 0.012463000137358904

9eedb4210414383e0ac869974b0c1775c6e99131 (my changes)
Patient has_many: 0.06817433332519916 has_many_through: 0.011987000121735036
Physician has_many: 3.0315577000010916 has_many_through: 0.01278599991928786

here we see, the time difference for physician isn't much. That's because physicians have a lot of relations (the issue I already mentioned at the top of the PR).

Quantenradierer:
Here it is. So to explain my benchmark: I have three tables with are connected by many and many through. I am creating and updating entries, depending on the configuration. If there is already data in the DB, it doesn't add additional. It takes a while to create, so I want to keep it for further tests.

When the data has been created, I pick 30 entries by random and then call reify on them (once with many, once with many_through). I take the time reify needs (with benchmark) and divide it by 30 (number of entries, so I get the average time it needs to get one).

The lines you see here is how long one call takes in seconds on average.

I recommend to let it run a second time, after the data has been created. Otherwise caching/internal behavior of the databases could create weird results.

## PosgreSQL
## configuration 1:
c0a90c2
Patient has_many: 0.08592150000040419 has_many_through: 0.014680000022053719
Physician has_many: 0.08508950000395998 has_many_through: 0.01321399980224669

9eedb42 (my changes)
Patient has_many: 0.08292269999704634 has_many_through: 0.012589000049047172
Physician has_many: 0.07952753333956934 has_many_through: 0.014558999857399613

## configuration 2:
c0a90c2
Patient has_many: 0.02805400000458273 has_many_through: 0.010950999974738806
Physician has_many: 2.952003366665061 has_many_through: 0.012887999881058931

9eedb42 (my changes)
Patient has_many: 0.02875359999986055 has_many_through: 0.01119199994718656
Physician has_many: 2.8140798666650273 has_many_through: 0.011221000226214528

## SQLite3
## configuration 1:
c0a90c2
Patient has_many: 0.07273769999737852 has_many_through: 0.010308999801054597
Physician has_many: 0.07027653332334012 has_many_through: 0.011400000017601997

9eedb42 (my changes)
Patient has_many: 0.07223619999519239 has_many_through: 0.010948999843094498
Physician has_many: 0.07078573333759171 has_many_through: 0.011802999768406153

## configuration 2:
c0a90c2
Patient has_many: 0.041685933326759066 has_many_through: 0.010714999923948199
Physician has_many: 2.359536100004334 has_many_through: 0.011008000292349607

9eedb42 (my changes)
Patient has_many: 0.047034733338902394 has_many_through: 0.011614999908488244
Physician has_many: 2.3564312999990458 has_many_through: 0.011145999887958169


As you can see, there are a few small positive but also negative differences. In my opionen, most changes are so small that it may be a measuring error. Expect the results for MySQL, where it simply indicates that MySQL has issues with subselects - which get's worst when having more data.

jaredbeck:
Thanks Bernd. I compared `origin/master` with and without your changes and got the following results:

```
# master
Patient has_many: 0.06572743333341956 has_many_through: 0.011231999998926767
Physician has_many: 2.6373713666666543 has_many_through: 0.01072199999998702

# your changes
Patient has_many: 0.07809873333338449 has_many_through: 0.010785000001305889
Physician has_many: 2.8108651333333 has_many_through: 0.009502999997494044
```

So, I'm not seeing any improvement.

I am using MariaDB 10.2.11.  What version of MySQL are you using?

Quantenradierer:
mysql  Ver 14.14 Distrib 5.7.20, for osx10.13 (x86_64), which is the latest available version. (2017-10-16)
Also, it's not Mac OS related, since it's occurring on Linux based systems too.

I downloaded MariaDB to test and well, it has no performance issue. Interesting.  

I need to think about this.

Quantenradierer:
So first of all, thank you for your time Jared.

I tried several ideas, how to solve this in sql (and asked google), but I can't find another better solution.

The only other way I found was:
`where("id IN (#{version_id_subquery})").`
can be replaced with
`JOIN (#{version_id_subquery}) as t on t.min_id = id`
(the subselect must also 'MIN(version_id) as min_id', but that's just naming)

This solved the issue in MySQL, but was ~10% slower in most cases in all databases, compared to my changes.

So, what now? I see those options:
- Accepting the PR
- Adding new methods which get's called for MySQL only
- Or everyone who uses MySQL should switch to MariaDB (until someone figured out another better way)

jaredbeck:
> So, what now? I see those options (after you confirmed that it's an MySQL issue, of course):
> - Adding new methods which get's called for MySQL only

This is my least favorite solution. I'd like to avoid RDBMS-specific code if possible.

> - Accepting the PR

Assuming I can reproduce your results on MySQL, this is my second-favorite solution. But, I really don't like performing two queries instead of one. To merge this PR, I'd also have to be convinced that the problem can't be fixed by configuring MySQL differently. This investigation will take many hours, so I don't know when I'll be able to get around to it. If you can recruit a third person, preferably a MySQL expert, that would be helpful.

> - Or everyone who uses MySQL should switch to MariaDB

This is my favorite solution. 😀  Seriously though, MariaDB seems better in every way and makes [incredible claims about compatibility](https://mariadb.com/kb/en/library/mariadb-vs-mysql-compatibility/). They even claim to "do monthly merges with the MySQL code base".

Quantenradierer:
> If you can recruit a third person, preferably a MySQL expert, that would be helpful.
Unfortunately, I can't. At least no one with experience in this field.
But I am going to make a few more tests in the next days, maybe I can tell you more about this bug or publish / find it on the mysql bug tracker if I know more.

PR #1023, opened 2017-12-10
Fix NameError when pg is not loaded

jaredbeck:
Most apps only load a single db adapter gem.

A MySQL app, for example will not load the pg gem.

See https://github.com/airblade/paper_trail/issues/1022


jaredbeck:
NameError introduced by https://github.com/airblade/paper_trail/pull/1018 ..

.. and we didn't catch it because our test suite loads all three adapter gems. See https://github.com/airblade/paper_trail/issues/1022

Issue #594, opened 2015-08-26
Reify on associations fails if using Single Table Inheritance


ksornberger:
It appears as though if Single Table Inheritance is being used on an association, if you try to reify with has_many: true or has_one: true, it will not find and build the previous version of the association and instead always use the current version of the association when browsing previous versions.

I believe this is happening because if STI is used, the item_type is always stored as the base class. But when calling the method reify_has_many_directly, it is using the main class of the association. This builds a WHERE clause that is looking for the wrong value in the item_type column.

ex:

```
class Document < ActiveRecord::Base
  has_paper_trail
  has_one :authorship
end

class Assignment < ActiveRecord::Base
  belongs_to :document 
end

class Authorship < Assignment
  has_paper_trail
end

@document.reify(has_many: true, has_one: true)
```

The last line there will produce a SQL query like this: 

```
SELECT  `versions`.* FROM `versions` 
INNER JOIN `version_associations` ON `version_associations`.`version_id` = `versions`.`id` 
WHERE (version_associations.foreign_key_name = 'document_id') 
AND (version_associations.foreign_key_id = 1) 
AND (versions.item_type = 'Authorship') 
AND (created_at >= '2015-08-26 20:07:28.000000' OR transaction_id = 933)  
ORDER BY versions.id ASC LIMIT 1
```

But item_type will never be 'Authorship', it is always stored as Assignment. I'm still working on figuring out a solution to this if I can, but this seems to be what is happening from my understanding.

Modifying the database manually and changing the item_type column from Assignment to Authorship results in being able to view the correct versions of the association as I browse through the document history.

jaredbeck:
Sounds like you're onto something.  Thanks for the bug report.

> I'm still working on figuring out a solution to this ..

Any luck?  Even a failing test would be welcome, thanks.

jaredbeck:
Still working on this, Kevin? Were you able to write a failing test? Thanks.

ksornberger:
Hey, sorry.
I'm still planning on looping back to take another look at this, just haven't had the time to get my head into the right space to try to sort this out again. I'll try to make it a priority soon and keep you updated!

jaredbeck:
I recently adapted the AR bug report template for use in PaperTrail.  Maybe it will help you write your failing test? Please check it out: https://github.com/airblade/paper_trail/blob/master/doc/bug_report_template.rb

I hope it helps!

jaredbeck:
[...]

PR #1026, opened 2017-12-20
Use base class name for reification of associations

andre1810:
Use base class name for reification instead of class name to reify polymorphic models properly.

Refers to #594 but won´t fix all issues. If you have more `has_one` associations to an polymorphic model you won´t get the correct associated models because sub type reference is missing in query.

### DIFF
--- a/lib/paper_trail/reifiers/belongs_to.rb
+++ b/lib/paper_trail/reifiers/belongs_to.rb
@@ -39,7 +39,7 @@ module PaperTrail
         # @api private
         def load_version(assoc, id, transaction_id, version_at)
           assoc.klass.paper_trail.version_class.
-            where("item_type = ?", assoc.klass.name).
+            where("item_type = ?", assoc.klass.base_class.name).
             where("item_id = ?", id).
             where("created_at >= ? OR transaction_id = ?", version_at, transaction_id).
             order("id").limit(1).first
--- a/spec/dummy_app/db/migrate/20110208155312_set_up_test_tables.rb
+++ b/spec/dummy_app/db/migrate/20110208155312_set_up_test_tables.rb
@@ -41,13 +41,13 @@ class SetUpTestTables < (
       t.string :name, null: false
       t.string :type, null: false
       t.integer :owner_id
-      t.timestamps null: false
+      t.timestamps null: false, limit: 6
     end
 
     create_table :skippers, force: true do |t|
       t.string     :name
       t.datetime   :another_timestamp, limit: 6
-      t.timestamps null: true
+      t.timestamps null: true, limit: 6
     end
 
     create_table :widgets, force: true do |t|
@@ -61,7 +61,7 @@ class SetUpTestTables < (
       t.date      :a_date
       t.boolean   :a_boolean
       t.string    :type
-      t.timestamps null: true
+      t.timestamps null: true, limit: 6
     end
 
     if ENV["DB"] == "postgres"
@@ -69,7 +69,7 @@ class SetUpTestTables < (
         t.string     :name
         t.integer    :post_ids,    array: true
         t.datetime   :login_times, array: true, limit: 6
-        t.timestamps null: true
+        t.timestamps null: true, limit: 6
       end
     end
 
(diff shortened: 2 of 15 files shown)


lib/paper_trail/reifiers/belongs_to.rb
```diff
@@ -39,7 +39,7 @@ def load_record(assoc, id, options, version)
         # @api private
         def load_version(assoc, id, transaction_id, version_at)
           assoc.klass.paper_trail.version_class.
-            where("item_type = ?", assoc.klass.name).
+            where("item_type = ?", assoc.klass.base_class.name).
```
jaredbeck: Nice. I was hoping AR provided something like this `base_class`, but I couldn't find it.

jaredbeck:
This looks good André. Please add:

1. at least one test, perhaps based on https://gist.github.com/andre1810/3cc1cbfaa8344a8ad8a840fbb9f91890, but let's use existing test models if we can. How about `Person`, `Animal`, `Cat`, and `Dog`? Person has many pets?
2. a changelog entry, under 9.0.0 -> Fixed (once this is in `master` we can discuss a backport if needed)

> .. [this] won´t fix all issues. If you have more has_one associations to an polymorphic model you won´t get the correct associated models because sub type reference is missing in query.

Do we have a bug report demonstrating that? Is there a gist I can run? Should we open a new issue for that or continue discussing it in https://github.com/airblade/paper_trail/issues/594?

andre1810:
@jaredbeck I added a test to verify the changes regarding `has_many` associations and another one to show the still open issue #594 with `has_one` associations which is currently skipped. If you don´t want to include the issue I can remove it but I think it will help solving the issue later. Maybe we should keep #594 open because it´s not fixed at all.

I also added a changelog entry. Backward compatibility would be nice as I currently use 8.1.1.

spec/models/person_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+require "spec_helper"
+
+RSpec.describe Person, type: :model, versioning: true do
+  it "baseline test setup" do
+    expect(Person.new).to be_versioned
+  end
+
+  describe "#cars and bicycles" do
+    it "can be reified", skip: "Known Issue #594" do
```
jaredbeck: This is great, thanks for including it.

spec/models/pet_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    last_version = person.reload.versions.last.reify(has_many: true)
+    expect(last_version.pets.length).to(eq(2))
+    expect(last_version.animals.length).to(eq(2))
+    expect(last_version.animals.map { |a| a.class.name }).to(eq(%w[Dog Cat]))
+    expect(last_version.pets.map { |p| p.animal.class.name }).to(eq(%w[Dog Cat]))
+    expect(last_version.animals.first.name).to(eq("Beethoven"))
+    expect(last_version.dogs.first.name).to(eq("Beethoven"))
+    expect(last_version.animals.second.name).to(eq("Sylvester"))
+    expect(last_version.cats.first.name).to(eq("Sylvester"))
```
jaredbeck: This test reads well. :+1:

spec/dummy_app/db/migrate/20110208155312_set_up_test_tables.rb
```diff
@@ -36,17 +36,18 @@ def up
...
+      t.integer :owner_id
+      t.timestamps null: false, limit: 6
     end
 
     create_table :skippers, force: true do |t|
       t.string     :name
-      t.datetime   :another_timestamp
-      t.timestamps null: true
+      t.datetime   :another_timestamp, limit: 6
+      t.timestamps null: true, limit: 6
```
jaredbeck: Ugh. MySQL's lack of Fractional Seconds Precision (FSP) has been a big headache on this project. I'm OK with this change, but can you remind me what datatype this produces in MySQL? Is it a `DATETIME(6)`? Does that mean six decimal points of FSP?

I wonder if, after this change, we can use less `timecop` in our tests .. but that's a story for another time. :smile:

jaredbeck:
Can you organize this into two commits, the first adding fractional seconds precision and the second doing everything else? If it's a pain, don't worry about it, I'll just squash it into one commit when I merge.

PR #1028, opened 2017-12-21
Use base class name for reification and add fractional second precision to test db

andre1810:
- Use base class name for reification of associations
- Add explicit precision on datetime columns to enable fractional second precision on mysql

PR #1030, opened 2017-12-21
Backport #1028: Use base class name for reification and add fractional second precision to test db

andre1810:
- Use base class name for reification of associations
- Add explicit precision on datetime columns to enable fractional second precision on mysql

Backport to 8.2.0


jaredbeck:
Released as 8.1.2

PR #1033, opened 2017-12-30
Discussing the API for request variables, part 2

jaredbeck:
In https://github.com/airblade/paper_trail/pull/1016, Alex and I discussed a new API for request variables.

### Current request variables

.. copied from #1016 for convenience:

| *key* | *versions table* | *api* |
| --- | --- | --- |
| :request_enabled_for_controller | no | public |
| :"enabled_for_#{model}" | no | public |
| :whodunnit | yes | public |
| :controller_info | yes | public |
| :transaction_id | yes | private |

> Some keys are data, to be saved in the `versions` table, others are configuration. Most are public, but `:transaction_id` is private.  The only thing they have in common is that they affect the current request.

### API Goals

1. Alex would like a convenient way to set a request variable temporarily, for which I propose `PaperTrail.request(whodunnit: 'Alex') do .. end`.
2. I have always wanted a better way to communicate which PT settings affect the current request vs. the whole thread, vs. all threads, and deprecating eg. `PaperTrail.whodunnit` in favor of `PaperTrail.request.whodunnit` communicates this unambiguously.

Let me know what you think, Alex. And @batter as always I'd appreciate your opinion too.


jaredbeck:
@AlexEvanczuk and @batter please review, thanks!

AlexEvanczuk:
Thanks @jaredbeck!  I should be able to look at this this week.

lib/paper_trail/request.rb
```diff
@@ -0,0 +1,123 @@
...
+      def set(options)
+        options.each do |k, v|
+          store[k] = v
+        end
+      end
+
+      # Returns a deep copy of the internal hash from our RequestStore.
+      # Keys are all symbols. Values are mostly primitives, but
+      # whodunnit can be a Proc.
+      # @api private
```
AlexEvanczuk: This is a private method, but it is called in `PaperTrail#request`.  I am not sure how you think about private methods (is it just for gem clients, or for service classes within the gem)?  In either case, I think that `def request` when called with `options` should pass the block to `Request` and the request itself knows how to set config for the duration of the block.

lib/paper_trail.rb
```diff
@@ -46,7 +48,11 @@ module PaperTrail
   class << self
     # @api private
     def clear_transaction_id
```
AlexEvanczuk: Should this and other `# @api private` methods, such as `transaction_id=`, have a deprecation warning?  Is your rationale that even if they're private, clients might be using them, and we want to deprecate them in the same way?

lib/paper_trail.rb
```diff
@@ -103,59 +138,54 @@ def timestamp_field=(_field_name)
...
+        ::ActiveSupport::Deprecation.warn(
+          "PaperTrail.whodunnit is deprecated, use PaperTrail.request.whodunnit"
+        )
+        request.whodunnit
+      else
+        ::ActiveSupport::Deprecation.warn(
+          "Passing a block to PaperTrail.whodunnit is deprecated, " \
+          'use PaperTrail.request(whodunnit: "John") do .. end'
+        )
+        before = request.whodunnit
```
AlexEvanczuk: Why not use method as suggested in the deprecation warning?

e.g.

```ruby
request(whodunnit: value)
  yield
end
```

lib/paper_trail.rb
```diff
@@ -103,59 +138,54 @@ def timestamp_field=(_field_name)
...
 
-    # Returns any information from the controller that you want
-    # PaperTrail to store.
-    #
-    # See `PaperTrail::Rails::Controller#info_for_paper_trail`.
+    # See PaperTrail::Request for information
     # @api public
-    def controller_info
-      paper_trail_store[:controller_info]
+    def controller_info(value = nil, &block)
```
AlexEvanczuk: Why did you add these parameters?  They aren't used in `Request`.  The `Request` method signature is `def controller_info`.  Wouldn't you get a `ArgumentError: wrong number of arguments` if you call this?  If so, can you add test coverage for this?

lib/paper_trail.rb
```diff
@@ -174,21 +204,20 @@ def transaction?
       ::ActiveRecord::Base.connection.open_transactions.positive?
     end
 
-    # @api public
+    # @api private
```
AlexEvanczuk: Is this intentional?  Again I'm not sure how you like to do this maintenance-wise, but I think that maybe we should add a deprecation notice that this is no longer a public method.

lib/paper_trail/request.rb
```diff
@@ -0,0 +1,123 @@
...
+      def to_h
+        store.store.deep_dup
```
AlexEvanczuk: I think that `store.store` is a little confusing.  Maybe something should have a different name here.

lib/paper_trail.rb
```diff
@@ -62,32 +68,61 @@ def enabled?
...
+    # All methods that handle request-level configuration and are now handled
+    # by PaperTrail::Request.
+    # @api public
+    def request(options = nil)
+      if options.nil?
+        Request
+      else
+        begin
+          before = Request.to_h
+          Request.set(options)
```
AlexEvanczuk: Do we want to limit what options are settable, and perhaps validate the options?  If they set:
```ruby
PaperTrail.request(some_nonexistant_option: true) do
 ...
end
```

What should happen?

What if they try to set a private attribute?  Such as `transaction_id`?

lib/paper_trail/version_number.rb
```diff
@@ -9,7 +9,7 @@ module PaperTrail
   module VERSION
     MAJOR = 8
     MINOR = 1
-    TINY = 1
+    TINY = 0
```
AlexEvanczuk: Shouldn't this be going in the other direction?

lib/paper_trail.rb
jaredbeck: Going forward, all transaction id methods should be `@api private`. I'll fix that in the depr. warnings and comments.

lib/paper_trail.rb
jaredbeck: Good suggestion, thanks.

lib/paper_trail.rb
jaredbeck: Seems like a mistake. I didn't mean to change the method signature. Good catch.

lib/paper_trail.rb
jaredbeck: Yes, intentional. I'll be changing the deprecation warning to say "deprecated without replacement".

jaredbeck:
@AlexEvanczuk and @batter ready for another review, thanks!

airblade:
Looks great!

lib/paper_trail/record_trail.rb
```diff
@@ -485,25 +497,25 @@ def without_versioning(method = nil)
         yield @record
       end
     ensure
-      @record.class.paper_trail.enable if paper_trail_was_enabled
+      PaperTrail.request.enable_model(@record.class) if paper_trail_was_enabled
     end
 
     # Temporarily overwrites the value of whodunnit and then executes the
     # provided block.
     def whodunnit(value)
```
AlexEvanczuk: Why is this method here?  Shouldn't it be deprecated in favor of using `PaperTrail.request.with(whodunnit: value) do ... end`?

spec/paper_trail/request_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+          call_count = 0
+          described_class.whodunnit = proc { call_count += 1 }
+          expect(described_class.whodunnit).to eq(1)
+          expect(described_class.whodunnit).to eq(2)
+        end
+      end
+    end
+
+    describe ".with" do
+      context "block given" do
```
AlexEvanczuk: Can you add tests for the validator here through the public `with` test?  That is, test that whodunnit, enabled for work (maybe enabled for is a good new test case since it's a regex), transaction id, and an invalid option.

Also can you change the initial value to be non nil so we can test something more useful, e.g. instead of `expect(described_class.whodunnit).to be_nil`, we can have `expect(described_class.whodunnit).to 'previous_whodunnit`

spec/paper_trail_spec.rb
```diff
@@ -94,37 +94,8 @@
   end
 
   describe ".version" do
-    it { expect(described_class).to respond_to(:version) }
```
AlexEvanczuk: I'd probably prefer to have some delegation/deprecation tests, e.g. to test deprecation issues such as changing the method signature to `controller_info`.

This could be done tersely with a shared example if you want to write less code:
```ruby
shared_example 'it delegates to request' do |method, *args, &block|
  it do
    expect(PaperTrail::Request).to receive(:method).with(args, &block).and_return(:some_return_value)
    expect(described_class.send(:method, args, &block).to_return(:some_return_value)
 end
end
```
```ruby
it_behaves_like 'it delegates to request', :controller_info, [], nil
it_behaves_like 'it delegates to request', :whodunnit, [:some_whodunnit] { 'some_block' }
```

AlexEvanczuk: Not sure if the above will work exactly as is, but that's the idea.  Could be just as straightforward to do it more explicitly for each method, but I think it would be less error-prone this way.

spec/paper_trail/request_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+module PaperTrail
+  ::RSpec.describe(Request, versioning: true) do
```
AlexEvanczuk: If you have the time, I'd probably prefer to have the other `@api public` methods tested as well here, such as `controller_info`, `controller_info=`, `enabled_for_controller?`, `enabled_for_controller=`, and `whodunnit=`.

spec/paper_trail/request_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        expect(PaperTrail.request.enabled_for_model?(Widget)).to eq(true)
+      end
+
+      after do
+        PaperTrail.request.enable_model(Widget)
+      end
+    end
+
+    describe ".whodunnit" do
+      context "when set to a proc" do
```
AlexEvanczuk: Can you add a context when not set to a proc?

README.md
```diff
@@ -703,34 +699,44 @@ PaperTrail::Version.delete_all ['created_at < ?', 1.week.ago]
...
+
+#### Setting `whodunnit` Temporarily
+
+To set whodunnit temporarily, for the duration of a block, use
+`PaperTrail.request.with`:
 
 ```ruby
-PaperTrail.whodunnit = proc do
-  caller.first{ |c| c.starts_with? Rails.root.to_s }
+PaperTrail.request.with(whodunnit: 'Dorian Marié') do
```
batter: I like the simplicity of this

lib/paper_trail.rb
```diff
@@ -46,124 +47,158 @@ module PaperTrail
   class << self
     # @api private
```
batter: Does this comment actually do anything or it's merely a note to help developers remember which methods are public / private?

lib/paper_trail/frameworks/cucumber.rb
```diff
@@ -3,9 +3,9 @@
 # before hook for Cucumber
 Before do
   PaperTrail.enabled = false
-  PaperTrail.enabled_for_controller = true
-  PaperTrail.whodunnit = nil
-  PaperTrail.controller_info = {} if defined? Rails
+  PaperTrail.request.enabled_for_controller = true
+  PaperTrail.request.whodunnit = nil
+  PaperTrail.request.controller_info = {} if defined? Rails
```
batter: Should this be `if defined?(::Rails)` to be consistent with the RSpec helper? I don't think there is a Rails module in this library yet but there could be one day

batter:
My first impression is that I like it and it seems like an eloquent solution that makes the gem nicely flexible to configure in a thread-safe manner or across all threads depending on what the user desires for various conditions. I'd like to spend some more time unpacking all the changes, will try to check back to do a deeper review later this week.

lib/paper_trail.rb
jaredbeck: It specifies whether something is part of our public API, ie. whether we would consider changing it to be a breaking change. Notably, not all `public` methods are public API, if that makes sense.

jaredbeck:
Thanks everyone for the reviews! 

Alex, I'd really appreciate your help writing some more tests. Would you like to? You can make a PR into this one, eg. "Alex wants to merge .. commits into request_variables from alex_branch" Thanks!

AlexEvanczuk:
@jaredbeck 

Added specs: https://github.com/airblade/paper_trail/pull/1052

lib/paper_trail/record_trail.rb
jaredbeck: Good catch. Yes, I will deprecate it.

lib/paper_trail/frameworks/cucumber.rb
jaredbeck: Sure, I'm always happy to use absolute constant paths, it's one less thing to worry about.

jaredbeck:
> I'd like to spend some more time unpacking all the changes, will try to check back to do a deeper review later this week.

Ben, I think I'm ready to merge this. Do you still want to do another review?

Andy, can you un-protect this branch? I want to do a rebase and tidy up my commits, but force-pushing is disabled. I'd prefer if only master and the stable branches (eg. 8-stable, 7-stable) were protected. Thanks.

batter:
LGTM, I'm sure there will be time to tweak functionality if needed before a release. Nice work by @AlexEvanczuk and @jaredbeck!

jaredbeck:
> LGTM, I'm sure there will be time to tweak functionality if needed before a release. Nice work by @AlexEvanczuk and @jaredbeck!

Thanks Ben! Yeah, I'm planning on releasing 9.0.0 in a few weeks, as soon as ruby 2.2 EoL is announced (https://www.ruby-lang.org/en/news/2017/09/14/ruby-2-2-8-released/)

PR #1036, opened 2018-01-12
No upper-bound dependency for AR

maschwenk:
These upper bound dependencies make it so that every time I go to upgrade a Rails version, I need to point to a fork of paper_trail until you guys release a new version. I don't see the major benefit of having an upper bound dependency. It just creates noise/busy work for you guys. If people have issues with newer versions of Rails they will report them.

In my project of ~100 gems, paper_trail and 2-3 others always come up when I go to do Rails upgrades for this reason

What do you guys think?

### DIFF
--- a/paper_trail.gemspec
+++ b/paper_trail.gemspec
@@ -28,7 +28,7 @@ has been destroyed.
   s.required_ruby_version = ">= 2.3.0"
 
   # Rails does not follow semver, makes breaking changes in minor versions.
-  s.add_dependency "activerecord", [">= 4.2", "< 5.2"]
+  s.add_dependency "activerecord", ">= 4.2"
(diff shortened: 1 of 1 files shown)


jaredbeck:
Hi Max,

The constraint matches which versions of `activerecord` a given version of PT has been tested against, and thus ensures compatibility. **Every new minor version of rails has required changes to PT**, some big (like the new attributes API), and some small (like optional `belongs_to`). Said another way, **each new version of rails is going to be incompatible with PT**, and the constraint helps people avoid incompatibility.

It sounds like you're an active Rails and PT user, and you upgrade sooner than the rest of the community does, and that's great. I'm a pretty aggressive updater myself. So, it'd be awesome if you could help work on PT, and make the necessary changes, each time a new minor version of rails comes out.

Thanks!

PR #1037, opened 2018-01-15
Add paper_trail.update_columns

TylerRick:
Add paper_trail.update_columns so you can record a version when using
`update_columns` (which skips all callbacks, including the `after_update`
callback).

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -302,16 +302,16 @@ module PaperTrail
       merge_metadata_into(data)
     end
 
+    # @api private
     def record_update_columns(changes)
-      if enabled?
-        versions_assoc = @record.send(@record.class.versions_association_name)
-        version = versions_assoc.create(data_for_update_columns(changes))
-        if version.errors.any?
-          log_version_errors(version, :update)
-        else
-          update_transaction_id(version)
-          save_associations(version)
-        end
+      return unless enabled?
+      versions_assoc = @record.send(@record.class.versions_association_name)
+      version = versions_assoc.create(data_for_update_columns(changes))
+      if version.errors.any?
+        log_version_errors(version, :update)
+      else
+        update_transaction_id(version)
+        save_associations(version)
       end
     end
 
@@ -323,7 +323,6 @@ module PaperTrail
         object: recordable_object,
         whodunnit: PaperTrail.whodunnit
       }
-      data[:created_at] = Time.now
       if record_object_changes?
         data[:object_changes] = recordable_object_changes(changes)
       end
@@ -437,13 +436,18 @@ module PaperTrail
 
     # Like the `update_column` method from `ActiveRecord::Persistence`, but also
     # creates a version to record those changes.
+    # @api public
     def update_column(name, value)
       update_columns(name => value)
     end
 
     # Like the `update_columns` method from `ActiveRecord::Persistence`, but also
     # creates a version to record those changes.
+    # @api public
     def update_columns(attributes)
+      # `@record.update_columns` skips dirty tracking, so we can't just use `@record.changes` or
+      # @record.saved_changes` from `ActiveModel::Dirty`. We need to build our own hash with the
+      # changes that will be made directly to the database.
       changes = {}
       attributes.each do |k, v|
         changes[k] = [@record[k], v]
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -226,7 +226,7 @@ module PaperTrail
         data[:created_at] = @record.updated_at
       end
       if record_object_changes? && changed_notably?
-        data[:object_changes] = recordable_object_changes
+        data[:object_changes] = recordable_object_changes(changes)
       end
       add_transaction_id_to(data)
       merge_metadata_into(data)
@@ -300,7 +300,7 @@ module PaperTrail
         data[:created_at] = @record.updated_at
       end
       if record_object_changes?
-        data[:object_changes] = recordable_object_changes
+        data[:object_changes] = recordable_object_changes(changes)
       end
       add_transaction_id_to(data)
       merge_metadata_into(data)
@@ -348,7 +348,7 @@ module PaperTrail
     # otherwise the column is a `text` column, and we must perform the
     # serialization here, using `PaperTrail.serializer`.
     # @api private
-    def recordable_object_changes(changes = changes())
+    def recordable_object_changes(changes)
       if @record.class.paper_trail.version_class.object_changes_col_is_json?
         changes
       else
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -271,20 +271,24 @@ module PaperTrail
     def record_update(force)
       @in_after_callback = true
       if enabled? && (force || changed_notably?)
-        versions_assoc = @record.send(@record.class.versions_association_name)
-        version = versions_assoc.create(data_for_update)
-        if version.errors.any?
-          log_version_errors(version, :update)
-        else
-          update_transaction_id(version)
-          save_associations(version)
-        end
+        create_update_version(data_for_update)
       end
     ensure
       @in_after_callback = false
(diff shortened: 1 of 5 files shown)


jaredbeck:
Hi Tyler. I don't have time to review this right now, but a few quick thoughts:

- Please don't change linter config
- Can this be done without a method extraction, or was that necessary because of `AbcSize` cop? Generally, if you can put refactoring in a separate PR, it makes the review easier.

Sorry, that's not much of a review, but it's all I have time for right now.

TylerRick:
> Can this be done without a method extraction, or was that necessary because of AbcSize cop? 
No, it was only necessary to allow reuse without duplication. As you wish. I've added back the duplication.

spec/models/widget_spec.rb
```diff
@@ -285,6 +285,20 @@
...
+    it "creates a version record" do
+      widget = Widget.create
+      expect(widget.versions.count).to eq(1)
+      Timecop.travel 1.second.since # because MySQL lacks fractional seconds precision
+      widget.paper_trail.update_columns(name: "Bugle")
+      # widget.update_attributes(name: "Bugle")
+      expect(widget.versions.count).to eq(2)
+      expect(widget.versions.last.event).to(eq("update"))
+      expect(widget.versions.last.changeset[:name]).to eq([nil, "Bugle"])
+      expect(widget.versions.last.created_at.to_i).to eq(Time.now.to_i)
```
jaredbeck: We can't guarantee that this expectation will pass. The operation could take more than a second, or cross a second-boundary.

spec/models/widget_spec.rb
```diff
@@ -285,6 +285,20 @@
     end
   end
 
+  describe ".paper_trail.update_columns", versioning: true do
+    it "creates a version record" do
+      widget = Widget.create
+      expect(widget.versions.count).to eq(1)
+      Timecop.travel 1.second.since # because MySQL lacks fractional seconds precision
+      widget.paper_trail.update_columns(name: "Bugle")
+      # widget.update_attributes(name: "Bugle")
```
jaredbeck: Please delete all commented-out code.

spec/models/widget_spec.rb
```diff
@@ -285,6 +285,20 @@
...
+      expect(widget.versions.count).to eq(2)
+      expect(widget.versions.last.event).to(eq("update"))
+      expect(widget.versions.last.changeset[:name]).to eq([nil, "Bugle"])
```
jaredbeck: Good, detailed test.

spec/models/on/empty_array_spec.rb
```diff
@@ -21,6 +21,15 @@ module On
       end
     end
 
+    describe ".paper_trail.update_columns" do
+      it "creates a version record" do
+        widget = Widget.create
+        assert_equal 1, widget.versions.length
+        widget.paper_trail.update_columns(name: "Bugle")
+        assert_equal 2, widget.versions.length
```
jaredbeck: Just FYI, there is a nice way to "expect a change" in RSpec. You don't have to use it, what you have is fine, just thought I'd share.

```
expect {
  widget.paper_trail.update_columns(name: "Bugle")
}.to(change { widget.versions.length }).from(1).to(2)
```

lib/paper_trail/record_trail.rb
```diff
@@ -406,6 +435,23 @@ def touch_with_version(name = nil)
...
     end
 
+    # Like the `update_column` method from `ActiveRecord::Persistence`, but also
+    # creates a version to record those changes.
+    def update_column(name, value)
+      update_columns(name => value)
+    end
+
+    # Like the `update_columns` method from `ActiveRecord::Persistence`, but also
+    # creates a version to record those changes.
```
jaredbeck: Good comment. Please add:

```
# @api public
```

lib/paper_trail/record_trail.rb
```diff
@@ -406,6 +435,23 @@ def touch_with_version(name = nil)
       @record.save!(validate: false)
     end
 
+    # Like the `update_column` method from `ActiveRecord::Persistence`, but also
+    # creates a version to record those changes.
```
jaredbeck: Good comment. Please add:

```
# @api public
```

lib/paper_trail/record_trail.rb
```diff
@@ -296,7 +296,36 @@ def data_for_update
...
       if record_object_changes?
-        data[:object_changes] = recordable_object_changes
+        data[:object_changes] = recordable_object_changes(changes)
+      end
+      add_transaction_id_to(data)
+      merge_metadata_into(data)
+    end
+
+    def record_update_columns(changes)
+      if enabled?
```
jaredbeck: I'd prefer a "guard clause" here.

```
return unless enabled?
```

lib/paper_trail/record_trail.rb
```diff
@@ -296,7 +296,36 @@ def data_for_update
         data[:created_at] = @record.updated_at
       end
       if record_object_changes?
-        data[:object_changes] = recordable_object_changes
+        data[:object_changes] = recordable_object_changes(changes)
+      end
+      add_transaction_id_to(data)
+      merge_metadata_into(data)
+    end
+
```
jaredbeck: Please comment all methods. This method's comment should include:

```
# @api private
```

lib/paper_trail/record_trail.rb
```diff
@@ -296,7 +296,36 @@ def data_for_update
...
+    def record_update_columns(changes)
+      if enabled?
+        versions_assoc = @record.send(@record.class.versions_association_name)
+        version = versions_assoc.create(data_for_update_columns(changes))
+        if version.errors.any?
+          log_version_errors(version, :update)
```
jaredbeck: Nice job including details like this error logging, and the `enabled?` check. :+1:

lib/paper_trail/record_trail.rb
```diff
@@ -296,7 +296,36 @@ def data_for_update
...
+
+    # Returns data for record_update_columns
+    # @api private
+    def data_for_update_columns(changes)
+      data = {
+        event: @record.paper_trail_event || "update",
+        object: recordable_object,
+        whodunnit: PaperTrail.whodunnit
+      }
+      data[:created_at] = Time.now
```
jaredbeck: Why do we have to set created_at? Won't AR do this?

lib/paper_trail/record_trail.rb
```diff
@@ -406,6 +435,23 @@ def touch_with_version(name = nil)
...
+      update_columns(name => value)
+    end
+
+    # Like the `update_columns` method from `ActiveRecord::Persistence`, but also
+    # creates a version to record those changes.
+    def update_columns(attributes)
+      changes = {}
+      attributes.each do |k, v|
+        changes[k] = [@record[k], v]
+      end
```
jaredbeck: Let's add a comment somewhere here explaining that we are constructing a hash that is compatible with the hash returned by `ActiveModel::Dirty#changes`.

spec/models/widget_spec.rb
TylerRick: True... should I just use a `Timecop.freeze` to keep `Time.now` fixed? (That's what I did.)

Looks like there's at least one other expectation with the same problem:
```
        expect(widget.paper_trail.version_at(Time.now)).to be_nil
```

TylerRick:
Thanks for the detailed review, @jaredbeck. I've pushed up another commit to fix the things you suggested.

jaredbeck:
Merged. Thanks for your contribution, Tyler!

> .. maybe we should follow up with some documentation in the readme also, or else people will not know this feature exists.

Please follow up with a docs PR, so other people can discover this feature. Thanks!

Jakanapes:
This is excellent, we were just looking at how to solve this issue cleanly.

thejosk:
@jaredbeck Can you please provide documentation for this?

jaredbeck:
> @jaredbeck Can you please provide documentation for this?

As I said above, contributions of documentation are welcome, thanks!

PR #1040, opened 2018-01-22
Raise helpful error when CVC is abstract

### LINKED ISSUE #961: TypeError (no implicit conversion of nil into String)
I'm trying to use the experimental feature (track associations) but I'm facing a issue:

I have the following models:

**UserVersion**:
```

class UserVersion < PaperTrail::Version
  self.table_name = :user_versions

  belongs_to :user, foreign_key: :whodunnit
end
```

**User**:

```
class User < ApplicationRecord
  has_paper_trail class_name: 'UserVersion'

  has_many :addresses, inverse_of: :user
  accepts_nested_attributes_for :addresses, allow_destroy: true
end
```

**Address**:

```
class Address < ApplicationRecord
  has_paper_trail # <- The error is here

  belongs_to :user, inverse_of: :addresses
end
```

So, the issue is: associations aren't being tracked when I have a "parent" custom version class... 

Always when I try to create/update "addresses", it show the following error:

> TypeError (no implicit conversion of nil into String):

If I remove the line has_paper_trail from Address model it works 100% without errors (but, of course, without versioning).

It seems like it searches for the `versions` (default) table instead of go to `user_versions`...

I opened a [question](http://stackoverflow.com/questions/43906353/track-associations-using-paper-trail) in SO, however it seems like it's a bug (it's normal in a experimental feature)..

I'm just trying to find a way to do this... even with a workaround.

Thanks in advance.

jaredbeck:
Instead of crashing when misconfigured Custom Version Classes are used, an error will be raised earlier, with a much more helpful message.

[Fixes #961]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Added
 
-- None
+- [#961](https://github.com/airblade/paper_trail/issues/961) - Instead of
+  crashing when misconfigured Custom Version Classes are used, an error will be
+  raised earlier, with a much more helpful message.
 
 ### Fixed
 
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -12,6 +12,16 @@ module PaperTrail
       or disable belongs_to_required_by_default.
     STR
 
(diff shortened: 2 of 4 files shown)


jaredbeck:
I'm kind of surprised I was able to write a test for this. Ruby so crazy.

PR #1052, opened 2018-02-25
Add specs to request variables branch

AlexEvanczuk:
@jaredbeck 

Added specs that I asked you to add in the original PR, let me know what you think and if I can merge it into your branch.


.gitignore
```diff
@@ -4,6 +4,7 @@
 .idea
 .rbenv-gemsets
 .rbenv-version
+.rspec_results
```
AlexEvanczuk: I added this so I could use `--next-failure` and `--only-failures` when running `rspec`, which I use a lot.

spec/spec_helper.rb
```diff
@@ -10,6 +10,8 @@
 end
 
 RSpec.configure do |config|
+  config.example_status_persistence_file_path = '.rspec_results'
```
AlexEvanczuk: I added this so I could use `--next-failure` and `--only-failures` when running `rspec`, which I use a lot.

.gitignore
jaredbeck: Nice. I use `--only-failures` on other projects, but I didn't know about `--next-failure` :+1: Thanks

jaredbeck:
This looks great, Alex, thanks!

I don't normally bother testing delegation of deprecated methods, but I'm happy to have the tests. Later this year, when we remove the deprecated methods, the test will be easy to remove.

I also usually avoid `shared_examples`, because they are usually in separate files, which makes them harder to follow, but you have used it in the same file and it is relatively self-contained.

However, these are very minor points, and I really appreciate you adding these tests, thanks.

AlexEvanczuk:
Sure thing!

PR #1053, opened 2018-02-25
Fix issues with touch_with_version & conditional versioning

westonganger:
Solves [Issue 1051](https://github.com/airblade/paper_trail/issues/1051)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 - PaperTrail now uses `frozen_string_literal`, so you should assume that all
   strings it returns are frozen.
 - Using `where_object_changes` to read YAML from a text column will now raise
+- [#1051](https://github.com/airblade/paper_trail/issues/1051) Fix `touch_with_version` does not work with conditional options
   error, was deprecated in 8.1.0.
 
 ### Breaking Changes, Minor
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -475,7 +475,7 @@ module PaperTrail
       @record.paper_trail.without_versioning do
         @record.save!(validate: false)
       end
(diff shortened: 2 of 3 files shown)


westonganger:
As of the first commit there is only 1 test failing, related to enum values. Not quite sure how to work around this at this point maybe someone can point me in the right direction.

jaredbeck:
> As of the first commit there is only 1 test failing, related to enum values in Rails v5.1. Not quite sure how to work around this at this point maybe someone can point me in the right direction.

Ugh. The code that handles enums is some of the most confusing code in PT. It'll be difficult to discuss asynchronously here, but if you want to pair on this, I might be able to help. Please email me and we can set something up. jared at jaredbeck dot com Thanks!

westonganger:
I `puts` debugged the crap out of the serializer system to no avail. Then I decided to re-order `record_update` and `record.save` and then all tests passed. So simple.

I did notice that `touch_with_version` only accepts a single attribute as an argument whereas the ActiveRecord `touch` method supports multiple attributes. However I will create a separate issue and PR for this.

Another question, I have to add `gem "rails-controller-testing"` to the `Gemfile` before testing every time. Should this not be added in the `Gemfile` or other dependency file you use for tests?

jaredbeck:
> Another question, I have to add gem "rails-controller-testing" to the Gemfile before testing every time. Should this not be added in the Gemfile or other dependency file you use for tests?

I think you're not running the tests correctly. Have you read `CONTRIBUTING.md`? I think you're not using `appraisal`.

jaredbeck:
> I puts debugged the crap out of the serializer system to no avail.

Ouch. Been there. It feels like there are too many layers, doesn't it?

> Then I decided to re-order record_update and record.save and then all tests passed. So simple.

I like it. I'm especially happy to get rid of `will_record_after_update?`. It's much easier to understand without a callback.

Thank you for your work on this.

PR #1058, opened 2018-02-28
Regenerate ToC

aried3r:
Using https://github.com/jonschlinkert/markdown-toc
`markdown-toc -i --maxdepth 3 --bullets='-'  README.md`

This is merely a proposition because I saw it was out of date and incomplete ("chapter" 9 was missing completely) and wanted to find a method that can semi-automate the job for you. No worries if you'd rather do it manually, then I'll apply the changes you want. Or just close this PR if you don't want it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ has been destroyed.
 
 ## Table of Contents
 
+<!-- toc -->
+
 - [1. Introduction](#1-introduction)
   - [1.a. Compatibility](#1a-compatibility)
   - [1.b. Installation](#1b-installation)
@@ -42,10 +44,9 @@ has been destroyed.
 - [4. Saving More Information About Versions](#4-saving-more-information-about-versions)
   - [4.a. Finding Out Who Was Responsible For A Change](#4a-finding-out-who-was-responsible-for-a-change)
   - [4.b. Associations](#4b-associations)
-    - [4.b.1. Known Issues](#4b1-known-issues)
-  - [4.c. Storing metadata](#4c-storing-metadata)
+  - [4.c. Storing Metadata](#4c-storing-metadata)
 - [5. ActiveRecord](#5-activerecord)
-  - [5.a. Single Table Inheritance](#5a-single-table-inheritance-sti)
+  - [5.a. Single Table Inheritance (STI)](#5a-single-table-inheritance-sti)
   - [5.b. Configuring the `versions` Association](#5b-configuring-the-versions-association)
   - [5.c. Generators](#5c-generators)
   - [5.d. Protected Attributes](#5d-protected-attributes)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,23 +44,31 @@ has been destroyed.
 - [4. Saving More Information About Versions](#4-saving-more-information-about-versions)
   - [4.a. Finding Out Who Was Responsible For A Change](#4a-finding-out-who-was-responsible-for-a-change)
   - [4.b. Associations](#4b-associations)
-    - [4.b.1. Known Issues](#4b1-known-issues)
```
aried3r: 4.b.1. disappears, because I defined `--max-depth 3`, with a higher max-depth, even more items would appear, 4.b.1. is the anomaly here, being the only ToC entry that deep.

README.md
```diff
@@ -42,23 +44,31 @@ has been destroyed.
...
-  - [7.a Minitest](#7a-minitest)
-  - [7.b RSpec](#7b-rspec)
-  - [7.c Cucumber](#7c-cucumber)
-  - [7.d Spork](#7d-spork)
-  - [7.e Zeus or Spring](#7e-zeus-or-spring)
+  - [7.a. Minitest](#7a-minitest)
+  - [7.b. RSpec](#7b-rspec)
+  - [7.c. Cucumber](#7c-cucumber)
+  - [7.d. Spork](#7d-spork)
+  - [7.e. Zeus or Spring](#7e-zeus-or-spring)
```
aried3r: The periods are correct I guess, they are present in all the other entries.

README.md
```diff
@@ -42,23 +44,31 @@ has been destroyed.
...
 - [8. Integration with Other Libraries](#8-integration-with-other-libraries)
+- [9. Related Libraries and Ports](#9-related-libraries-and-ports)
+- [Articles](#articles)
+- [Problems](#problems)
+- [Contributors](#contributors)
+- [Contributing](#contributing)
+- [Inspirations](#inspirations)
+- [Intellectual Property](#intellectual-property)
```
aried3r: I'm not sure if you want all these entries in the ToC, I find them useful because I was looking for where to report issues, usage questions etc (the latter not being answered by the README as far as I can see. Should these go to StackOverflow? GitHub issues?).

Also, they can't easily be ignored by the generator :D

aried3r: Nevermind the usage questions part, just found out it's not in the README but in the issue template.

jaredbeck:
Thanks Anton. It's a good idea to automate the ToC, because it can be a pain to maintain. 

This doesn't change any URLs, right? I often reference specific parts of the ToC from a StackOverflow answers.

Otherwise, I'm kind of ambivalent about this. I wish it didn't include the sections at the end like "Inspirations", but I can live with it.

aried3r:
> This doesn't change any URLs, right? I often reference specific parts of the ToC from a StackOverflow answers.

I think it doesn't as long as the headlines stay the same. Once you want something like `[Foo](#bar)` where the `#anchor` is not the same as the headline, automation becomes a problem.

For this I'd recommend using either a specific revision, tag or branch to link to, so they kinda stay the same, for example for 7.1.3 https://github.com/airblade/paper_trail/blob/v7.1.3/README.md#5-activerecord, which might not exist in, i.e. paper_trail 10.

That way links should stay valid and relevant to the SO answer.

> Otherwise, I'm kind of ambivalent about this. I wish it didn't include the sections at the end like "Inspirations", but I can live with it.

Yeah, I can understand that. But I was actually looking for the _Problems_ part in the ToC, didn't find it and thus led me to open this PR, heh.

jaredbeck:
>> This doesn't change any URLs, right?

> I think it doesn't as long as the headlines stay the same.

Cool, then I'm fine with merging this.

> For this I'd recommend using either a specific revision, tag or branch to link to ..

Good point, I'll try to remember that for future SO answers.

Thanks for your continued contributions to PT!

PR #1063, opened 2018-03-13
Integrate versioning into AR touch method

westonganger:
Related to [Issue 1061](https://github.com/airblade/paper_trail/issues/1061)

### DIFF
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -128,7 +128,7 @@ module PaperTrail
     # "class attributes", instance methods, and more.
     # @api private
     def setup(options = {})
-      options[:on] ||= %i[create update destroy]
+      options[:on] ||= %i[create update destroy touch]
       options[:on] = Array(options[:on]) # Support single symbol
       @model_class.send :include, ::PaperTrail::Model::InstanceMethods
       setup_options(options)
@@ -204,8 +204,6 @@ module PaperTrail
       options_on.each do |event|
         public_send("on_#{event}")
       end
-
-      public_send("on_touch")
     end
 
     def setup_habtm_change_callbacks(assoc)
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -116,6 +116,14 @@ module PaperTrail
       @model_class.paper_trail_options[:on] << :update
     end
 
+    # Adds a callback that records a version after a "touch" event.
+    # @api public
+    def on_touch
+      @model_class.after_touch { |r|
+        r.paper_trail.record_update(force: true, in_after_callback: true)
+      }
+    end
+
     # Set up `@model_class` for PaperTrail. Installs callbacks, associations,
     # "class attributes", instance methods, and more.
     # @api private
@@ -196,6 +204,8 @@ module PaperTrail
       options_on.each do |event|
         public_send("on_#{event}")
       end
+
+      public_send("on_touch")
     end
 
     def setup_habtm_change_callbacks(assoc)
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -570,7 +570,7 @@ module PaperTrail
     # Rails 5.1 changed the API of `ActiveRecord::Dirty`. See
     # https://github.com/airblade/paper_trail/pull/899
     #
-    # Event can be any of the following (create, update, destroy, touch).
+    # Event can be any of the three (create, update, destroy).
     #
     # @api private
     def attribute_in_previous_version(attr_name)
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -7,6 +7,12 @@ module PaperTrail
       my_model_instance.paper_trail.whodunnit('John') is deprecated,
       please use PaperTrail.request(whodunnit: 'John')
     STR
+
+    DPR_TOUCH_WITH_VERSION = <<-STR.squish.freeze
+      my_model_instance.paper_trail.touch_with_version is deprecated,
+      please use my_model_instance.touch
+    STR
+
     RAILS_GTE_5_1 = ::ActiveRecord.gem_version >= ::Gem::Version.new("5.1.0.beta1")
 
     def initialize(record)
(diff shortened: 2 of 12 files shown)


westonganger:
Something to consider with this PR is the following lines:

https://github.com/westonganger/paper_trail/blob/master/lib/paper_trail/model_config.rb#L209-L208

Currently it will attach to the `touch` event always. Maybe this should be added as one of the available `:on` events and it included by default or must be specified?

lib/paper_trail/model_config.rb
```diff
@@ -196,6 +204,8 @@ def setup_callbacks_from_options(options_on = [])
       options_on.each do |event|
         public_send("on_#{event}")
       end
+
+      public_send("on_touch")
```
jaredbeck: I think you can just call `on_touch` here, right? The `public_send` is not necessary.

jaredbeck:
> Currently it will attach to the touch event always. Maybe this should be added as one of the available :on events and it included by default or must be specified?

Good thinking. I think the callback should be installed by default, but configurable via the `:on` option.

jaredbeck:
This will need a changelog entry and probably some significant changes to documentation in comments and in the readme.

westonganger:
I guess `touch_with_version` could be deprecated. Should that be a part of a separate PR? or would you want to do this yourself?

westonganger:
Okay the following items have been added to this PR:

- Add `touch` to list of `:on` events (by default it is included)
- Deprecate `touch_with_version`
- Updated changelog
- Updated readme documentation to use `touch` instead of `touch_with_version`

lib/paper_trail/record_trail.rb
```diff
@@ -461,8 +467,9 @@ def source_version
     # version records are inserted. It's unclear under which specific
     # circumstances this technique should be adopted.
     #
-    # @api public
+    # @deprecated
     def touch_with_version(name = nil)
+      ::ActiveSupport::Deprecation.warn(DPR_TOUCH_WITH_VERSION, caller(1))
```
jaredbeck: Nice deprecation implementation

lib/paper_trail/record_trail.rb
```diff
@@ -563,17 +570,16 @@ def attribute_changed_in_latest_version?(attr_name)
     # Rails 5.1 changed the API of `ActiveRecord::Dirty`. See
     # https://github.com/airblade/paper_trail/pull/899
     #
-    # Event can be any of the three (create, update, destroy).
+    # Event can be any of the following (create, update, destroy, touch).
```
jaredbeck: "touch" is not an event. There are still only three events: create, update, and destroy. `touch` will cause the insertion of an "update" event. See `record_update`.

I think this will be confusing to many people, and I'm wondering how we can document it better.

spec/models/on/update_spec.rb
```diff
@@ -5,9 +5,10 @@
 
 module On
   RSpec.describe Update, type: :model, versioning: true do
+    let(:record) { described_class.create(name: "Alice") }
```
jaredbeck: Please do not use `let`. I believe it makes tests harder to understand, especially when it's too far away, lexically, from the test.

spec/models/on/update_spec.rb
```diff
@@ -17,13 +18,19 @@ module On
...
         expect(record.versions.last.event).to(eq("banana"))
       end
     end
+
+    describe "#touch" do
+      it "does not create a version for the touch event" do
+        count = record.versions.count
+        expect(count).to eq(count)
+      end
+    end
```
jaredbeck: Good test. Thanks for finding and using the existing models.

spec/models/widget_spec.rb
```diff
@@ -232,21 +232,39 @@
...
+      expect { widget.touch }.to change {
+        widget.versions.count
+      }.by(+1)
+    end
+
+    it "does not create a version using without_versioning" do
+      count = widget.versions.count
+      widget.paper_trail.without_versioning do
+        expect(count).to eq(count)
+      end
```
jaredbeck: I don't understand the implementation of this test. It's like saying `expect(7).to eq(7)`, right? The local variable `count` is unchanging, right?

lib/paper_trail/record_trail.rb
westonganger: Well should we make `touch` an event? I would simply add another optional argument to `record_update(force:, in_after_callback:, event: "update")` or add a `record_touch` method.

Otherwise I can simply add documentation in the readme next to the touch documentation stating that the event will be recorded as `update`

jaredbeck: > should we make touch an event?

I don't know. 

Reasons against: A `touch` is simply an update. It updates timestamps. If we had a touch event, the lifecycle would be harder to understand. Instead of eg. create-update-update-destroy, it's eg. create-update-touch-touch-update-touch-destroy. The "middle part" of the lifecycle becomes heterogeneous. This heterogeneity would be a major breaking change.

Reasons for: Maintains a one-to-one correspondence between callbacks and events, which makes callbacks easier to understand I guess. It's not a big advantage.

So I guess I'm leaning against making touch an event.

westonganger: Okay then in that case, I simply updated the readme stating the event will be saved as `update`

jaredbeck:
Thanks, Weston.

gduprey:
Is there a way to globally disable the tracking of the touch event?  Or more generically, set the default /create/update/destroy/touch global 'on:' defaults ?

As implemented, it breaks when using the delay_touching gem (specific case is when a delayed touch is for a record that was marked for destruction, the attempt to create the version  association in record_update() in record_trail.rb throws an exception because the @record no longer is persisted (as it was deleted before the touch events are fired - touch having been delayed as the function of the delay_touching gem).

Given this was not the default behavior before and the alternative is updating potentially hundreds of models (we have a pretty large app) to disable on: :touch, I'd much rather have a PaperTrail.config.xxxx setting once.

We upgraded from 7.x to 10.x and had this break (it worked before, likely because it was before touch versioning was done)

ghiculescu:
@gduprey 	https://github.com/paper-trail-gem/paper_trail/pull/1166 adds this

ghiculescu:
.... of course it does, you wrote it.

spec/models/on/update_spec.rb
joshuapinter: Reading this from 2020, and I completely agree. It can be useful for dynamic values to DRY up tests but if it's overused, it becomes a nightmare. Additionally, the `let` block is only executed once it is called so that alone can be confusing and have detrimental side effects with tests.

PR #1064, opened 2018-03-15
Move without_versioning to model

westonganger:
The current implementation of`without_versioning` is incorrectly set as an instance method when it should be part of the class which the current behaviour clearly shows. This PR fixes this and deprecates the old behavior.

The only difference with the new method is that it no longer accepts a method name, as its on the class not the instance. I didnt want to add this to the new implementation because I believe it is an unnecessary addition to the API.

This was somewhat related to Issue #916

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Deprecated
 
+- [#1064](https://github.com/airblade/paper_trail/pull/1064) - `without_versioning` has
+  now been moved to the model eg. `MyModel.paper_trail.without_versioning` and the old way,
+  `my_instance.paper_trail.without_versioning` is deprecated. Behaviour remains unchanged except 
+  that it no longer accepts a method name, it only accepts a block.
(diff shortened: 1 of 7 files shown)


jaredbeck:
> The current implementation of without_versioning is incorrectly set as an instance method when it should be part of the class ..

I agree it's misleading. However, instead of moving it, we could just drop it and people could use `PaperTrail.request(enabled_for_controller: false) do ..`, right?

Please rebase.

westonganger:
Rebased.

Your suggestion could work but `enabled_for_controller` seriously needs to be renamed its awful, especially for this use case. Its not even a good name for the controller method because it says enabled for controller but really its enabled for the request/action.

Why not change it to `without_versioning`. Ex. 
```
PaperTrail.request(without_versioning: true) do .. end
```

and the controller method name

```
class MyController < ApplicationController

def paper_trail_request_without_versioning

end
```

westonganger:
Another note is that, there would no longer be an option to disable only 1 model as its now disabled for all models within that block. But I'm not sure that even matters because anyone should be able to easily organize the code around this.

jaredbeck:
Closing via #1069

PR #1067, opened 2018-03-20
Add support to Rails 5.2

rafaelfranca:
Rails 5.2 is going to be released really soon, and our application depends on paper_trail so we thought about adding Rails 5.2 to the test matrix and allowing installing with it.


paper_trail.gemspec
```diff
@@ -28,7 +28,7 @@ has been destroyed.
   s.required_ruby_version = ">= 2.3.0"
 
   # Rails does not follow semver, makes breaking changes in minor versions.
-  s.add_dependency "activerecord", [">= 4.2", "< 5.2"]
+  s.add_dependency "activerecord", [">= 4.2", "< 6"]
```
jaredbeck: Because rails makes breaking changes in minor versions, I'd prefer `< 5.3` here.

rafaelfranca: There is no 5.3. After 5.2 it will be 6.0.

rafaelfranca:
Added the changelog. The build is failing because of what I believe is a bug in bundler or rubygems. Not sure how to fix this. Locally I have to comment the rspec-rails dependency, run `bundle install`, uncomment `rspec-rails` and run `bundle install` again.

I didn't change the constraint because the next Rails release will be 6.0, not 5.3.

paper_trail.gemspec
jaredbeck: > There is no 5.3. After 5.2 it will be 6.0.

Of course, if you say this, I believe you :) You are the person who would know. But, it doesn't change my policy for this version constraint. What advantage does PaperTrail gain by specifiying `< 6` rather than `< 5.3`? Maybe I am missing something.

rafaelfranca:
I just opened an issue report to bundler https://github.com/bundler/bundler/issues/6449

paper_trail.gemspec
rafaelfranca: I think the only advantage is make sure that we are using a version constraint that actually exists, but it doesn't matter if we use 6.0, so I'll change.

rafaelfranca:
Changed the constraint.

jaredbeck:
Changes approved; 2b3824e29ca37b2fbb9875da7c1b15a6a1af4883 looks good. 

Once CI passes we can merge.

jaredbeck:
Closed via https://github.com/airblade/paper_trail/pull/1068, preserving your authorship. Thanks Rafael.

PR #1074, opened 2018-04-12
PaperTrail.request { } should return the value of the block to aid in passive wrapping

krainboltgreene:
So I just spent the last 3 hours debugging because (ultimately) I had overwritten a method that some library depended on and I assumed that the below would be passively wrapping:

``` ruby
def valid_for_authentication?(*)
  PaperTrail.request(whodunnit: "The Machine") do
    super
  end
end
```

Turns out it doesn't, because it's forced to return nil.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Breaking Changes
 
-- None
+- `PaperTrail.request do ... end` now returns whatever value is returned from the given block. For example:
+``` ruby
+raise PaperTrail.request(whodunnit: "Michael Bay") do
+  ApplicationError
+end
+# => ApplicationError has been raised!
+```
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,7 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Breaking Changes
 
-- `PaperTrail.request do ... end` now returns whatever value is returned from the given block. For example:
-``` ruby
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,7 +7,13 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
...
 ### Breaking Changes
 
-- None
+- `PaperTrail.request do ... end` now returns whatever value is returned from the given block. For example:
+``` ruby
+raise PaperTrail.request(whodunnit: "Michael Bay") do
+  ApplicationError
+end
+# => ApplicationError has been raised!
+```
```
jaredbeck: As I said, please put this under "Added". I don't see it as a breaking change.

Also, please omit the code example. The single sentence description is sufficient.

jaredbeck:
>> Document the return type in the method documentation

> @jaredbeck The whole point of this PR is that the return type could be anything.

To clarify, I'd like the method documentation to say something like "If a block is given, returns value from block"

CHANGELOG.md
krainboltgreene: We're changing the return type from always nil to sometimes nil, are you sure that's not a breaking change?

krainboltgreene:
Roger.

CHANGELOG.md
jaredbeck: > We're changing the return type of a public method from always nil to sometimes nil, are you sure that's not a breaking change?

Yes, you're right (both correct and righteous) to question this :grin: I don't consider it a breaking change, because I don't think anyone is depending on the fact that it is returning nil. What do you think? Do you think people will be depending on the nil?

jaredbeck:
Closing via https://github.com/airblade/paper_trail/pull/1077, preserving your authorship. Thanks Kurtis.

PR #1077, opened 2018-04-23
Return value of block from PaperTrail.request

jaredbeck:
.. making it easier to use at the call site.


jaredbeck:
Supersedes #1074

Issue #1076, opened 2018-04-17
Touch not saving changes. touch_with_version did


key88sf:
Simple example of the difference:
```
> p = Person.last
> p.name = "New Name"
> p.touch
# UPDATE people SET updated_at = ....
```
^^ This **does not** update the name in either the database or the versions table.
```
> p.paper_trail.touch_with_version
# UPDATE people SET updated_at=..., name='New Name' ...
```
^^ This updates **both** the name in the database, **and** adds a correct version row for the updated name in the versions table.

This is a huge difference in behavior, and one which is relied upon likely by many people besides us -- it is **integral** to the behavior of the [counter_culture](https://github.com/magnusvk/counter_culture) gem's support of the `with_papertrail` option.

Please either do not deprecate this, or fix the behavior of `touch` to be equivalent. Right now they are two completely different behaviors.

westonganger:
Okay so touch_with_version should almost be called save_with_version

key88sf:
Sure, that would be fine. I definitely do not think this should just be deprecated though, as it's completely different behavior.

jaredbeck:
Closing via https://github.com/airblade/paper_trail/pull/1078

> Okay so touch_with_version should almost be called save_with_version

I like the idea of having a `save_with_version` method, but if that's what it's called, I'd rather have it just do a `save` and not (simulate) a `touch`. So, that led me to the implementation in https://github.com/airblade/paper_trail/pull/1078. I hope it works for you, @key88sf.

key88sf:
Yes thanks!

Issue #1073, opened 2018-04-02
[v9.0.0] Difference between `#touch` and `#paper_trail.touch_with_version`?


fukayatsu:
Is this an expected behavior?

### my_bug_report.rb

```ruby
# frozen_string_literal: true

# Use this template to report PaperTrail bugs.
# Please include only the minimum code necessary to reproduce your issue.
require "bundler/inline"

# STEP ONE: What versions are you using?
gemfile(true) do
  ruby "2.5.0"
  source "https://rubygems.org"
  gem "activerecord", "5.1.4"
  gem "minitest", "5.10.3"
  gem "paper_trail", "9.0.0", require: false
  gem "sqlite3"
end

require "active_record"
require "minitest/autorun"
require "logger"

# Please use sqlite for your bug reports, if possible.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = nil
ActiveRecord::Schema.define do
  # STEP TWO: Define your tables here.
  create_table :users, force: true do |t|
    t.text :first_name, null: false
    t.timestamps null: false
  end

  create_table :versions do |t|
    t.string :item_type, null: false
    t.integer :item_id, null: false
    t.string :event, null: false
    t.string :whodunnit
    t.text :object, limit: 1_073_741_823
    t.text :object_changes, limit: 1_073_741_823
    t.integer :transaction_id
    t.datetime :created_at
  end
  add_index :versions, %i[item_type item_id]
  add_index :versions, [:transaction_id]

  create_table :version_associations do |t|
    t.integer  :version_id
    t.string   :foreign_key_name, null: false
    t.integer  :foreign_key_id
  end
  add_index :version_associations, [:version_id]
  add_index :version_associations, %i[foreign_key_name foreign_key_id],
    name: "index_version_associations_on_foreign_key"
end
ActiveRecord::Base.logger = Logger.new(STDOUT)
require "paper_trail/config"

# STEP THREE: Configure PaperTrail as you would in your initializer
PaperTrail::Config.instance.track_associations = true

require "paper_trail"

# STEP FOUR: Define your AR models here.
class User < ActiveRecord::Base
  has_paper_trail
end

# STEP FIVE: Please write a test that demonstrates your issue.
class BugTest < ActiveSupport::TestCase
  def test_touch_with_version
    user = User.create(first_name: "Jane")
    user.paper_trail.touch_with_version
[...]

PR #1084, opened 2018-05-14
Fix saving changes on #touch

fukayatsu:
from: https://github.com/paper-trail-gem/paper_trail/issues/1073#issuecomment-388766519

Related to #1076

### DIFF
--- a/spec/models/on/touch_spec.rb
+++ b/spec/models/on/touch_spec.rb
@@ -20,6 +20,12 @@ module On
         )
         expect(record.versions.last.event).to eq("update")
       end
+
+      it "saves a object" do
+        record = described_class.create(name: "Alice")
+        record.touch
+        expect(record.versions.last.reify.name).to eq("Alice")
+      end
     end
 
     describe "#update" do
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -120,7 +120,7 @@ module PaperTrail
     # @api public
     def on_touch
       @model_class.after_touch { |r|
-        r.paper_trail.record_update(force: true, in_after_callback: true)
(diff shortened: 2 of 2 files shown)


spec/models/on/touch_spec.rb
```diff
@@ -20,6 +20,12 @@ module On
         )
         expect(record.versions.last.event).to eq("update")
       end
+
+      it "saves a object" do
+        record = described_class.create(name: "Alice")
+        record.touch
+        expect(record.versions.last.reify.name).to eq("Alice")
+      end
```
fukayatsu: On master branch, this spec passes on AR4.2 and AR5.0 but fails on AR5.1 and AR 5.2

Please see https://travis-ci.org/fukayatsu/paper_trail/builds/378643661 .

jaredbeck:
Thanks for the PR, Fukaya-san. The test you added is good.

My only concern is the `in_after_callback:` argument. 

```diff
-        r.paper_trail.record_update(force: true, in_after_callback: true)
+        r.paper_trail.record_update(force: true, in_after_callback: false)
```

PT installs an `after_touch` callback. It is, *actually* an after callback. So, if we say `in_after_callback: false`, that would be confusing. 

First, I thought maybe we should change the name of the argument? But, it's not easy to come up with a better name.

How about just passing another argument?

```diff
-        r.paper_trail.record_update(force: true, in_after_callback: true)
+        r.paper_trail.record_update(force: true, in_after_callback: true, is_touch: true)
```

Please see https://github.com/paper-trail-gem/paper_trail/pull/1085

fukayatsu:
@jaredbeck Thank you for the quick response. 
#1085 looks good to me!

jaredbeck:
Fixed by #1085, released in 9.0.2. Thanks, Fukaya-san.

PR #1085, opened 2018-05-14
Alternative fix for #1084

jaredbeck:
See discussion in https://github.com/paper-trail-gem/paper_trail/pull/1084


westonganger:
Why not make the `is_touch` argument on all the internal methods default to false? That would make for a lot less code changes in this PR.

PR #1091, opened 2018-05-19
Add PaperTrail.config.association_reify_error_behaviour

### LINKED ISSUE #1090: Add Papertrail initializer option to switch PaperTrail::Reifiers::HasOne::FoundMoreThanOne for a warning instead
After trying to upgrade from v8.1.2 to v9+. I started recieving this error `PaperTrail::Reifiers::HasOne::FoundMoreThanOne`. I believe the old paper trail just kinda said "I don't care about being totally correct, just carry on". So now I cannot upgrade because of this exception. Can we add an option to the initializer to instead just warn and continue.

https://github.com/paper-trail-gem/paper_trail/blob/06a5e30ab40581ee4f035989c961dd520f48e29a/lib/paper_trail/reifiers/has_one.rb#L77

We could change this to something like

```ruby
if PaperTrail.config.associations_just_warn_enabled
  # do warning
  versions.first
else
  # do raise
end
```

westonganger:
Fixes #1090 

The added tests do pass when configured appropriately in the dummy app initializer before running the tests. However I am not sure exactly how to dynamically test this option so I have commented out the tests, please advise.


lib/paper_trail/reifiers/has_one.rb
```diff
@@ -74,7 +74,18 @@ def load_version(assoc, model, transaction_id, version_at)
           when 1
             versions.first
           else
-            raise FoundMoreThanOne.new(base_class_name, versions.length)
+            case PaperTrail.config.association_reify_error_behaviour.to_s
+            when "warn"
+              version = versions.first
+              version.logger&.warn(
+                format(FoundMoreThanOne::MESSAGE_FMT, base_class_name, versions.length)
+              )
```
jaredbeck: Nice idea to use `version.logger`.

I think I'd prefer:

```ruby
version.logger&.warn(
  FoundMoreThanOne.new(base_class_name, versions.length).message
)
```

What do you think?

spec/models/person_spec.rb
```diff
@@ -34,5 +34,53 @@
...
+    #   person.update_attributes(name: "Peter")
+
+    #   expect(person.reload.versions.length).to(eq(3))
+
+    #   expect(person.versions.first.logger).to_not(
+    #     receive(:warn).with(/Unable to reify has_one association/)
+    #   )
+
+    #   person.reload.versions.second.reify(has_one: true)
+    # end
```
jaredbeck: If you're having trouble writing this type of test, I'd be happy with testing `load_version` in isolation, even using mocks. I'd be happier with a test that used real database records, but for now an isolated test is OK with me.

README.md
```diff
@@ -948,8 +948,9 @@ Associations are an **experimental feature** and have the following known
 issues, in order of descending importance.
 
 1. PaperTrail only reifies the first level of associations.
-1. Does not fully support STI (For example, see `spec/models/person_spec.rb` and
-   `PaperTrail::Reifiers::HasOne::FoundMoreThanOne` error)
+1. Sometimes the has_one association will find more than one possible candidate and will raise a `PaperTrail::Reifiers::HasOne::FoundMoreThanOne` error. For example, see `spec/models/person_spec.rb`
+  - If you are not using STI, you may want to just assume the first result (of multiple) is the correct one and continue. Versions pre v8.1.2 and below did this without error or warning. To do so add the following line to your initializer: `PaperTrail.config.association_reify_error_behaviour = :warn`. Valid options are: `[:error, :warn, :ignore]`
+  - When using STI, even if you enable :warn you will likely still end up recieving an `ActiveRecord::AssociationTypeMismatch` error in which case your SOL
```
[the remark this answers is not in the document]
westonganger: Im not going to recommend the :error option to STI users in case they are also struggling with the non-STI version of this problem. Beside the :error option is already the default.

lib/paper_trail/reifiers/has_one.rb
westonganger: Meh. I'd prefer not to because that creates an extra error object for no reason.

jaredbeck: > Meh. I'd prefer not to because that creates an extra error object for no reason.

The reason is to avoid duplicating the implementation of `FoundMoreThanOne#message`. By duplicating the implementation, if the message changes, you have to change it in two places.

> .. that creates an extra error object ..

For me, the duplication of code seems worse than a few object allocations, but that's not based on any benchmark of course, just a gut instinct.

lib/generators/paper_trail/install_generator.rb
```diff
@@ -45,7 +45,8 @@ def create_migration_file
     def create_initializer
       create_file(
         "config/initializers/paper_trail.rb",
-        "PaperTrail.config.track_associations = #{!!options.with_associations?}\n"
+        "PaperTrail.config.track_associations = true\n",
+        "PaperTrail.config.association_reify_error_behaviour = :error"
```
jaredbeck: Nice thinking to include this in the generator

spec/models/association_error_behaviour/ignore.rb
```diff
@@ -0,0 +1,40 @@
...
+      allow(logger).to receive(:warn)
+
+      person.reload.versions.second.reify(has_one: true)
+
+      expect(logger).not_to(
+        have_received(:warn).with(/Unable to reify has_one association/)
+      )
+    end
+  end
+end
```
jaredbeck: I like how you put these specs in different files. 

Can we put them in `spec/paper_trail/association_error_behaviour` though? Because, `spec/models` is intended to correlate to `app/models`.

spec/paper_trail/config_spec.rb
```diff
@@ -32,6 +32,14 @@ module PaperTrail
       end
     end
 
+    # describe "association_reify_error_behaviour" do
+    #   it "accepts a value" do
+    #     config = described_class.instance
+    #     config.association_reify_error_behaviour = :warn
+    #     expect(config.association_reify_error_behaviour).to eq(:warn)
+    #   end
+    # end
```
jaredbeck: Given the other specs you've added, it looks like we can delete this commented-out spec?

jaredbeck:
Thanks Weston, I'd like to tidy a few very minor things up (eg. changelog format, nothing major) and then I'll cut a release.

jaredbeck:
Released in 9.1.0

Issue #1082, opened 2018-05-12
Can I store a custom version of object_changes?


hashwin:
When using paper_trail, is there a more efficient way to store object_changes? Especially for jsonb columns, a large amount of data is stored in the table as it stores the entire value of the field before and after the change was made, even if the change was one key in a large object.

Right now, when I send it to the view I am using Hashdiff (https://github.com/liufengyun/hashdiff) to find the specific things that were changed, but I have to run this every time a view is loaded, plus there is a large load on the db as the entire field change is stored every time - versions is by far my largest table.

Is there an option of storing the Hashdiff directly in the object_changes and dumping that into the view every time it is loaded? It would make things a lot faster and less load on the db.

jaredbeck:
Hi Ashwin, 

> .. versions is by far my largest table.

The best way to manage table size is with `version_limit`, though it's not appropriate for all use cases.

> .. is there a more efficient way to store object_changes?

First, as you probably know, `object_changes` is an optional column, so you could just drop it.

Now, is there a more efficient way? From the perspective of disk space, yes, we could be storing a patch instead of full values. For example, if you updated `users.full_name` from 'Alice Anderson' to 'Alice Bowman', we could store 's/Anderson/Bowman'. (That's just an example, it's not a practical patch representation'). For very large strings this could save a few blocks on disk. (A typical block size is 4 kibibytes, so the string would need to be quite long, much longer than a person's name.)

I have two concerns about any change to what we store on disk. The first concern is that PT has to support two different formats, and de/serialization is already one of the most complicated parts of PT. Second, I'd want to see benchmarks showing that the change doesn't negatively impact the performance of A) queries like `where_object_changes`, or B) reification.

If you'd like to continue to work on this, I think the first thing we should discuss is your proposed patch format.

jaredbeck:
> If you'd like to continue to work on this, I think the first thing we should discuss is your proposed patch format.

Is this something you are still working on? Want me to keep this open?

seanlinsley:
Jared and I previously discussed this a bit here: https://github.com/paper-trail-gem/paper_trail/issues/1046

FWIW, I have it on my to-do list to give [logidze](https://github.com/palkan/logidze) a try, which only stores changes as opposed to the whole object, and relies on database triggers instead of ActiveRecord callbacks.

jaredbeck:
[...]

Issue #1046, opened 2018-02-08
Skip object on update


seanlinsley:
Around 50% of a long-lasting `versions` table of mine is storing the `object` on updates:

```sql
select event,
  pg_size_pretty(sum(pg_column_size(object))) as object_size,
  pg_size_pretty(sum(pg_column_size(object_changes))) as object_changes_size
from versions
group by event;

  event  | object_size | object_changes_size 
---------+-------------+---------------------
 create  | 4736 bytes  | 7418 MB
 update  | 13 GB       | 4246 MB
 destroy | 830 MB      | 
```

In this application that's not really storing useful information; the only time `object` becomes useful is when a record needs to be reified after being destroyed.

Would a PR be welcome to add the ability to skip `object` on update? It'd essentially be the inverse of #507.

jaredbeck:
Hi Sean, 

If the problem is table size, one could use `version_limit`.

You might also see some small savings by using a `jsonb` column instead of `text`.

If neither of those helps, and you want to proceed with your original idea, I think I'd prefer it be a separate gem at first, to see how many people have the same problem. It sounds like an uncommon problem. We can mention the new gem in various places in the readme.

seanlinsley:
It's important to track every update, so `version_limit` wouldn't work.

`jsonb` is already being used. I started looking at options to compress the data when I realized that half of the storage space used isn't hold useful data.

How would you imagine the gem would be structured? The existing code doesn't seem easy to extend.

I doubt we're going to get many people interested in trying this out; time-crunched app developers are much more likely to pay for a bigger database or delete old data.

Unless there's something I'm missing, it seems like this should be the default behavior of Paper Trail. Storing duplicates of all the columns on every update just isn't scalable over long periods of time. Though of course that'd be a breaking change, so the better choice would be to recommend it in the documentation.

jaredbeck:
> How would you imagine the gem would be structured? The existing code doesn't seem easy to extend.

I think it could be done as-is by re-opening classes or via module-prepend, but those techniques are brittle and I'm open to suggestions on how make it easier to extend without those techniques.

> Unless there's something I'm missing, it seems like this should be the default behavior of Paper Trail. Storing duplicates of all the columns on every update just isn't scalable over long periods of time. Though of course that'd be a breaking change, so the better choice would be to recommend it in the documentation.

I agree, I've always wondered why Andy chose to design PT that way. Some of his original comments about that decision can still be found in the readme:

[...]

PR #1093, opened 2018-05-21
Allow storing diffed changes in object_changes column

hashwin:
Based on the discussion in https://github.com/paper-trail-gem/paper_trail/issues/1082

This allows you to store changes in the format specified by https://github.com/liufengyun/hashdiff
This is especially useful when dealing with serialized json columns in postgres.

For example, say you have a jsonb column as follows:
```
{
  custom_values: {
    name: 'abc'
    age: 5,
    tags: ['a', 'b']
  }
}
```
and it is changed to
```
{
  custom_values: {
    name: 'def'
    age: 5,
    tags: ['a', 'c']
  }
}
```

Previously the object changes column would have been stored as follows:
```
{
  object_changes: {
    custom_values: [
      {
        name: 'abc',
        age: 5,
        tags: ['a', 'b']
      },
      {
        name: 'def',
        age: 5,
        tags: ['a', 'c']
      }
    ]
  }
}

```

With this change, it will be stored as follows:
```
{
  object_changes: {
    custom_values: [
      ["~", [:name], "abc", "def"], ["-", [:tags, 1], "b"], ["+", [:tags, 1], "c"]
    ]
  }
}
```

As you can see, the keys that were not changed (i.e. in this case, age) will not appear in the object changes at all, thus increasing storage efficiency.

If you are on board with this, I can add tests and a migration path to this format.

### DIFF
--- a/spec/models/version_spec.rb
+++ b/spec/models/version_spec.rb
@@ -16,7 +16,7 @@ module PaperTrail
         end
       end
 
-      context " with object_changes_adapter" do
+      context "with object_changes_adapter" do
         let(:adapter) { instance_spy("CustomObjectChangesAdapter") }
 
         before do
--- a/spec/paper_trail/model_spec.rb
+++ b/spec/paper_trail/model_spec.rb
@@ -66,7 +66,7 @@ RSpec.describe(::PaperTrail, versioning: true) do
           PaperTrail.config.object_changes_adapter = nil
         end
 
-        it "calls the adapter's where_object_changes method" do
+        it "calls the adapter's load_changeset method" do
           changeset = @widget.versions.last.changeset
           expect(changeset[:a]).to eq("b")
           expect(changeset[:c]).to eq("d")


hashwin:
Also, FWIW HashDiff supports "Patch/Unpatch" which could possibly be used for reification, removing the need for an object column, which could be used to solve the problem described here:  https://github.com/paper-trail-gem/paper_trail/issues/1046.

Though this would probably entail a much more holistic change in the way paper_trail works. Just a thought, definitely don't have the time to work on that one 😄

jaredbeck:
Hi Ashwin,

> If you'd like to continue to work on this, I think the first thing we should discuss is your proposed patch format.

Thanks for the clear PR description, I now understand your recommended patch format.

What do you think about using a plugin design? Something like this:

```ruby
PaperTrail.config.object_diff_adapter = PaperTrail::ObjectDiffAdapters::HashDiff.new

# lib/paper_trail/record_trail.rb
def recordable_object_changes(changes)
  if PaperTrail.config.object_diff_adapter.present?
    changes = PaperTrail.config.object_diff_adapter.diff(changes)
  end
  # ...
end
```

We would also use the adapter in `VersionConcern#load_changeset`.

`config.object_changes_adapter` would be another possible name.

So, the adapter interface would have at least three methods:

1. Given the a hash of `changes` (as is currently passed to `recordable_object_changes`) return something that can be assigned to the `object_changes` attribute (as we currently do in `recordable_object_changes`).
2. Whatever `VersionConcern#load_changeset` needs
3. Implement `where_object_changes`, just as `Queries::Versions::WhereObjectChanges` currently does.

That way, the new adapter (and others to come) can live in separate gems. The gem should probably be named `paper_trail-hashdiff`, per the rubygems.org [Name your gem](https://guides.rubygems.org/name-your-gem/) guide. (See the "Use dashes for extensions" section).

jaredbeck:
Hi Ashwin, Are you still working on this? Want me to keep it open?

hashwin:
@jaredbeck I have updated based on your suggestions. Right now I'm not 100% sure about details within custom attribute serialization/deserialization as I have not gone too deep into load_changeset. The method already works fine if there is no custom serialization.

I think this could be moved forward and people who are doing the following could use it:
1. Not using any Custom Attribute serialization 
2. Using a jsonb column for object_changes

What do you think?

jaredbeck:
Hi Ashwin, thanks for the update. 

I see you have added a file `hash_diff_adapter.rb`. I was hoping you would put that in a separate gem, named `paper_trail-hashdiff`. What do you think?

> .. I'm not 100% sure about details within custom attribute serialization/deserialization as I have not gone too deep into load_changeset. The method already works fine if there is no custom serialization.

I think `load_changeset` must support the adapter, otherwise the semi-popular `changeset` method will not work, right?

When you say "custom attribute", is that [`ActiveRecord::Attributes::ClassMethods#attribute`](http://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html)?
I think it's OK for the `paper_trail-hashdiff` gem to lack support for that. If someone needs it, they can build it later.

> I think this could be moved forward and people who are doing the following could use it: .. Using a jsonb column for object_changes

I am OK with the `paper_trail-hashdiff` gem only supporting jsonb, but the code in `paper_trail` itself needs to support all three data types:

1. `WhereObjectChanges#text` - JSON in text column
2. `WhereObjectChanges#json` - actual `json` column
3. `WhereObjectChanges#jsonb` - `jsonb` column

So far, it looks like you have only implemented `jsonb`. Can you add adapter support to all three please?

hashwin:
Hi @jaredbeck 
I've made the requested changes and extracted the adapter into a gem here: https://rubygems.org/gems/paper_trail_hashdiff

jaredbeck:
Hi Ashwin, This is getting close.

Have we chosen the best name for this config option? Do you like `object_diff_adapter` or `object_changes_adapter` or something else?

You can name your gem whatever you want, of course, but the correct spelling according to rubygems.org is `paper_trail-hashdiff` (with a hyphen), as I said above.

>
> The gem should probably be named `paper_trail-hashdiff`, per the rubygems.org [Name your gem](https://guides.rubygems.org/name-your-gem/) guide. (See the "Use dashes for extensions" section).
>

Please add to the readme:
- Section 6.c.1., explaining how to install your adapter
- Section 6.c.2., explaining how to write a different adapter (specify the interface)

Please add an entry to the changelog under Unreleased -> Added.

hashwin:
Hi Jared, I have updated with the changes you requested.

jaredbeck:
> Hi Jared, I have updated with the changes you requested.

Hi Ashwin. Everything looks great. However, I just realized we don't have any tests. I'm sorry I didn't mention tests earlier. It should be OK to have some very simple tests. If you want to have a simple little adapter in `spec/support`, you can do that, or you could use a [mock](https://relishapp.com/rspec/rspec-mocks/v/3-7/docs) adapter. I'm happy with either.

PR #1096, opened 2018-05-30
Add consolidated list of plugins

westonganger:
This may also come in handy for things like #1093


jaredbeck:
Looks good, but please regenerate the Table of Contents. We are using markdown-toc these days. See https://github.com/paper-trail-gem/paper_trail/pull/1058

PR #1102, opened 2018-06-04
Introduce config.object_changes_adapter

jaredbeck:
Allows users to write third-party adapters giving them complete control over the contents of the `object_changes` column.

See https://github.com/paper-trail-gem/paper_trail/pull/1093


fameoflight:
Is there anyway to specify this at model level rather than global level. My usecase is I have large jsonb which I want track when it's changed with md5 don't actually want to store full values. 

But I definitely don't want this to be global behavior for all jsonb

PR #1108, opened 2018-06-07
Fix for issue #594, reifying sub-classed models that use STI

lorint:
This update provides a fix for [Issue #594](https://github.com/paper-trail-gem/paper_trail/issues/594) -- Reify on associations fails if using Single Table Inheritance.

In order to properly reify a version of a model using STI, item_type can refer directly to the class name instead of being base_class. When making this change, the five reifiers in the gem [paper_trail-association_tracking](https://github.com/westonganger/paper_trail-association_tracking) that refer to `base_class` can also be updated, which is covered in [PT-AT PR #5](https://github.com/westonganger/paper_trail-association_tracking/pull/5). With these changes, the previously failing example in person_spec.rb will pass.

One key benefit from this change is to allow PT to represent STI model names. So if you have for instance the model "Person", there could be subclasses for "Doctor" and "Patient". The subclassed name gets put into `item_type` so that audit logs can be more clear.


jaredbeck:
Given that this changes what is stored in the database, do we need to provide a convenient way for people convert their existing database records? If they did not convert their records, then reification (without `version_reification_class`) would produce incorrect results, right? Is there any practical way to detect when someone has not converted their existing records, and warn them?

spec/models/family/celebrity_family_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+require "spec_helper"
+
+module Family
+  RSpec.describe CelebrityFamily, type: :model, versioning: true do
+    describe "#reify" do
+      context "belongs_to" do
+        it "uses the correct item_type in queries" do
+          parent = described_class.new(name: "Jermaine Jackson")
+          parent.path_to_stardom = "Emulating Motown greats such as the Temptations and "\
+                                   "The Supremes"
```
jaredbeck: 😀

lorint: Ah yes -- I have a distinct interest in making a quasi-boring topic (auditing) into something enduring.

And actually I would hope that things like email notifications and cool reporting with pretty charts could piggy-back on top of what we store in `versions`!

lorint: > ...do we need to provide a convenient way for people convert their existing database records?

We could, and in our project I actually have a migration that does this.  It's possible to include with this a generator that creates a migration to cycle through existing PaperTrail::Version entries and update item_type appropriately.

-= But =-  wonderfully enough this is not entirely essential because ActiveRecord just does the right thing; reified objects get instantiated as their proper subclassed type.  With the Person / Doctor / Patient example, if for a Doctor you have a combo of entries from the past (<= v9.1) where the item_type was stored as Person, and now there are other entries (>= v9.2) some changes stored as Doctor, the reification always brings back a Doctor.

The AT part currently does not rehydrate all the historic attributes properly when this update is applied, so that's the reason for the couple notes in `spec/models/pet_spec.rb`. During the coming week I have high hopes to solve that and then have the best of all worlds.  The biggest "win" in my mind becomes more specific reporting of the item_type, and certainly having the example in `spec/models/person.rb` work out is a wonderful benefit as well.

jaredbeck:
>>  ...do we need to provide a convenient way for people convert their existing database records?

> We could, and .. I actually have a migration that does this .. But wonderfully enough this is not entirely essential because ActiveRecord just does the right thing; reified objects get instantiated as their proper subclassed type. With the Person / Doctor / Patient example, if for a Doctor you have a combo of entries from the past (<= v9.1) where the item_type was stored as Person, and now there are other entries (>= v9.2) some changes stored as Doctor, the reification always brings back a Doctor.
>

Great. If both `item_type`s reify into a Doctor, then it sounds like we don't have to help people convert existing records, and I'm less inclined to treat this as a breaking change. We'll definitely need a test that demonstrates old-style records reifying correctly.

I don't think this PR will make it into 9.2. Maybe 9.3? It's hard to predict the contents of future releases.

> The AT part currently does not rehydrate all the historic attributes properly when this update is applied, so that's the reason for the couple notes in spec/models/pet_spec.rb. During the coming week I have high hopes to solve that and then have the best of all worlds. The biggest "win" in my mind becomes more specific reporting of the item_type, and certainly having the example in spec/models/person.rb work out is a wonderful benefit as well.

Sweet! Thanks for your work on this tricky issue.

lib/paper_trail/has_paper_trail.rb
```diff
@@ -78,6 +78,19 @@ module InstanceMethods
...
+          type_column = self.class.inheritance_column
+          @paper_trail_type_name =
+            (respond_to?(type_column) ? send(type_column) : nil) ||
+            self.class.name
+          if @paper_trail_type_name == self.class.base_class.name
+            @paper_trail_type_name = ""
+          end
+        end
+        @paper_trail_type_name
+      end
```
jaredbeck: - Can this method be moved into `RecordTrail`?
- Why use an instance variable? I'd prefer a local variable if possible.

lib/paper_trail/model_config.rb
```diff
@@ -168,6 +168,31 @@ def cannot_record_after_destroy?
         ::ActiveRecord::Base.belongs_to_required_by_default
     end
 
+    def type_aware_has_many(klass)
```
jaredbeck: In keeping with the name of the method `setup_associations`, I would call this `setup_versions_association`.

lib/paper_trail/model_config.rb
```diff
@@ -168,6 +168,31 @@ def cannot_record_after_destroy?
...
+        klass.versions_association_name,
+        lambda do |object|
+          # Support Single Table Inheritance (STI) with custom inheritance columns.
+          #
+          # For example, imagine a `version` whose `item_type` is "Animal". The
+          # `animals` table is an STI table (it has cats and dogs) and it has a
+          # custom inheritance column, `species`. If `attrs["species"]` is "Dog",
+          # type_name is set to `Dog`. If `attrs["species"]` is blank, type_name
+          # is set to `Animal`. You can see this particular example in action in
+          # `spec/models/animal_spec.rb`.
```
jaredbeck: - Assuming `paper_trail_type_name` can be moved into `RecordTrail`, and renamed to something like `versions_association_item_type`, this would be `object.paper_trail.versions_association_item_type`.
- Extract a local variable `item_type = object.paper_trail.versions_association_item_type` and use in the two places below to avoid calculating it twice.

lib/paper_trail/record_trail.rb
```diff
@@ -237,7 +237,10 @@ def record_create
       @in_after_callback = true
       return unless enabled?
       versions_assoc = @record.send(@record.class.versions_association_name)
-      versions_assoc.create! data_for_create
+      version = versions_assoc.new data_for_create
+      version.item_type = @record.class.name
```
jaredbeck: Can `item_type` be moved into `data_for_create`?

CHANGELOG.md
```diff
@@ -24,7 +24,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#594](https://github.com/paper-trail-gem/paper_trail/issues/594) -
+  In order to properly reify a version of a model using STI, item_type can refer to the class instead of base_class. When making this change, the five reifiers in the gem [paper_trail-association_tracking] that refer to base_class can also be updated, which is covered in [PT-AT PR #5](https://github.com/westonganger/paper_trail-association_tracking/pull/5). With these changes, the previously failing example in person_spec.rb passes.
```
jaredbeck: - Can you boil this down a bit, maybe into a single sentence?
- Should we link to #594 or this PR (#1108) or both?

lib/paper_trail/has_paper_trail.rb
lorint: This caches @paper_trail_type_name -- although perhaps people would not often call .versions more than once on a versioned object.

If it moves anywhere then I'd prefer to have it back in model_config where the has_many is established.

(TBH this started because the Rubocop ABC is set at 22, and I couldn't get under 22.69 after breaking out the has_many stuff in model_config, and with `type_name` only being a simple string it seemed like a fairly lightweight addition to each object.)

lib/paper_trail/model_config.rb
lorint: Good suggestion!

lib/paper_trail/record_trail.rb
lorint: Unfortunately when initializing the new Version, ActiveRecord uses the polymorphic association to override whatever you specify for `item_type`, so we need to `new` it and then reset the `item_type` on our own.

CHANGELOG.md
lorint: Perhaps linking to this [PR](https://github.com/paper-trail-gem/paper_trail/pull/1108) is best, and I'll make reference to [#594](https://github.com/paper-trail-gem/paper_trail/issues/594) at the top.

lorint:
> Did you want to wait for [PT-AT #5](https://github.com/westonganger/paper_trail-association_tracking/pull/5) to be released ... and then we update our gemspec dependency?

I believe the other way around is best -- if Weston were to update his side now then it would break lots of tests.  Once this change is put into place then those 5 reifiers in PT-AT can be updated from referencing `class.base_class` back to simply`class`.

jaredbeck:
What if we leave `item_type` as-is and add another column called e.g. `item_reification_type`?

Now that I understand this proposal better, I'm uncomfortable about taking over the `item_type` column. As long as we're going to use a polymorphic association (`as: :item`), I think rails should control that column.

jaredbeck:
> Now that I understand this proposal better, I'm uncomfortable about taking over the item_type column. As long as we're going to use a polymorphic association (as: :item), I think rails should control that column.
>

To make my concerns a bit more concrete, here are two possible future scenarios:

1. rails decides to add some validation to polymorphic-STI associations, starts raising error when we call `item`
2. rails decides to build new feature X, based on polymorphic-STI associations, which requires `item_type` to contain base class. PaperTrail gets bug reports saying "X doesn't work".

Sorry for not raising this concern earlier, it did not occur to me.

> What if we leave item_type as-is and add another column called e.g. item_reification_type?

What do you think?

lorint:
Have had some free cycles again to dig back in on this, and I think we're finally out of the woods; some VERY good things to report!  We don't have to force ActiveRecord's hand in the least.

Remarkably when calling .create!() as we had been doing previously the result is different than if we call .new() and then .save().  Thankfully the latter causes `item_type` to receive the subclassed model name!  I would not anticipate this to cause any kind of breaking change in the future, especially in light of how `Inheritance#find_sti_class` works:

https://apidock.com/rails/ActiveRecord/Inheritance/ClassMethods/find_sti_class

Putting together a commit now for you to see this in action.

jaredbeck:
> Remarkably when calling .create!() as we had been doing previously the result is different than if we call .new() and then .save(). Thankfully the latter causes item_type to receive the subclassed model name! 

I cannot confirm this. Here is what I tried:

```ruby
  it "saves the expected item_type" do
    cat = Cat.create(name: "Buddy")
    expect(cat.versions.first.item_type).to eq('Animal')
    cat2 = Cat.new(name: "Buddy")
    cat2.save
    expect(cat2.versions.first.item_type).to eq('Cat') # => fails: Expected Cat, got Animal
  end
```

The above demonstrates that create (or create!) produces the same item_type as new/save.

lorint:
> I cannot confirm this

Ah -- the trick is not in how the Cat object gets built out, but how RecordTrail builds out the associated Version object:

    it "saves the expected item_type" do
      buddy_cat = Cat.create(name: "Buddy")

      # Let's make a couple versions in two possible ways that RecordTrail could:
      version1 = buddy_cat.versions.create!(buddy_cat.paper_trail.data_for_create)
      version2 = buddy_cat.versions.new(buddy_cat.paper_trail.data_for_create)
      version2.save
      expect(version1.item_type).to eq("Animal")
      expect(version2.item_type).to eq("Cat")
    end

jaredbeck:
> Ah -- the trick is not in how the Cat object gets built out, but how RecordTrail builds out the associated Version object:

Ah ha! Because of the `lambda` this PR would add to the `has_many`?

lorint:
> Because of the lambda this PR would add to the has_many?

That is a necessary additional piece.  First the three places with .new() .save() in RecordTrail are able to record the real class name.  Then when referencing `versions`, the original lambda had just `order(model.timestamp_sort_order)`.  In this updated `has_many` this behaviour is preserved for non-STI tables, and for STI tables there's the .unscope and .where to always find based on their real class name.

jaredbeck:
I raised some concerns about us tampering with the `item_type` column.

> Now that I understand this proposal better, I'm uncomfortable about taking over the item_type column. As long as we're going to use a polymorphic association (as: :item), I think rails should control that column.
>
>
> To make my concerns a bit more concrete, here are two possible future scenarios:
>
> 1. rails decides to add some validation to polymorphic-STI associations, starts raising error when we call `item`
> 2. rails decides to build new feature X, based on polymorphic-STI associations, which requires `item_type` to contain base class. PaperTrail gets bug reports saying "X doesn't work".
>

Can you address these concerns, please? Do you not share them? Thanks.

lorint:
> Can you address these ... concerns about us tampering with the `item_type` column.  Do you not share them?

I fail to see what kind of concern we can have.  I mean, things operate just as before such that when we build a Version for a `create` or `update` it fully relies on the polymorphic association `as: :item` to establish `item_type` and `item_id`.  The nice benefit of this PR is that it cleans up other tampering that we previously had to do, digging into `base_class` in order to reify.  I think all of these changes can now be considered win-win.

The three places in RecordTrail that use ActiveRecord code to make everything work each go a little something like this:

    version = versions_assoc.new data_for_update_columns(changes)
    version.save

It is curious that calling versions.create() through the has_many only gets the base class, and instead doing versions.new() + version.save() arrives upon the real class, but that's not something that gives me concern.  I figure the battle to try to get the kind of update to fix .create() rolled into Rails would be a significant uphill battle, especially since many people may have coded around this nuance, and rely on this behaviour to remain unmodified.

I'm very happy with how it all works, and feel that further refinement may only be to improve performance slightly by caching `versions_association_item_type`.

jaredbeck:
Thanks for the response. I've given some specific concerns,

> .. here are two possible future scenarios:
>
> 1. rails decides to add some validation to polymorphic-STI associations, starts raising error when we call item
> 2. rails decides to build new feature X, based on polymorphic-STI associations, which requires item_type to contain base class. PaperTrail gets bug reports saying "X doesn't work".

Could you please address these specific concerns, one at a time? Thanks.

lorint:
While approaching the answers below please bear in mind that with such general concerns surrounding the possibility of changes to ActiveRecord, I find these scenarios to be best addressed from the perspective of the Rails community at large.  In no way is it an attempt to downplay the concerns, but rather to frame them appropriately due to the large-scale fallout they would almost certainly have.  Far beyond the realm of this PR and PaperTrail.  Here we go:
> 1. rails decides to add some validation to polymorphic-STI associations, starts raising error when we call item.

Amongst the extensive set of Rails production code worldwide it is no doubt fairly common to come across projects with a polymorphic association referencing a model that can be subclassed via STI.  As such, we're in good company, I would expect.  (Anecdotally, several of our current projects where I work fall into this category, polymorphism pointing to something STI.)  As such for any organisation such as ours that has chosen to utilise these standard features of ActiveRecord, it would be incredibly painful if some breaking change were to come along in regards to how subclassed objects are loaded through a belongs_to.  Very specifically, many people would be up-in-arms if `Inheritance#find_sti_class` were not able to properly determine which type to instantiate from.  The backlash would be significant.  In that light, I would never expect that a validator would haphazardly be put into Rails that would break so many projects reliant on how ActiveRecord works.  Being as a validator would also impact performance, it would be highly unlikely to be met with any kind of positive reception for those interested in optimising their code.
> 2. rails decides to build new feature X, based on polymorphic-STI associations, which requires item_type to contain base class. PaperTrail gets bug reports saying "X doesn't work".

Along the same lines of the answer above, if pure ActiveRecord code today can create polymorphic `*_type` entries that refer to the subclass, and ActiveRecord is also able to function perfectly well via a polymorphic belongs_to using that column (indiscriminate of having the base_class or a subclass entered), then we are in good company with others who undoubtedly utilise polymorphic associations pointing to subclassed models.  An enormous range of code (some of my own) is currently in production with no issues surrounding rehydrating objects in this way.

I feel it a bootless errand to spend energy fearing for the sky to fall on us in these ways, for surely if it were to be, we would chime in company with others so affected.

lorint:
Just found that for versions that reference a subclassed STI object that was destroyed, the reifier wasn't able to correctly build the object again.  Have extended an example in pet_spec.rb and implemented a fix.

lorint:
Ha!  Travis had a network error while bundling during one of the runs.  Things really are 100%.

lorint:
Curious ... always native code being fetched apparently, but yesterday it was `byebug` and today it's `nokogiri` -- acting up only with one specific database / ruby / rails combo.

jaredbeck:
> Curious trouble with Travis .. today it's nokogiri

I re-ran that one build and it passed so I wouldn't worry about it.

> Just found that for versions that reference a subclassed STI object that was destroyed .. Have extended an example in pet_spec.rb and implemented a fix.

Nice!

> In no way is it an attempt to downplay the concerns .. the concerns [given would cause] large-scale fallout [in the rails community] ..

First, thank you for your thoughtful consideration of these concerns. 

I am operating under the idea that neither scenario would be seen as a breaking change by the rails team. There would be no fallout. We would be the loners doing something different from the community, something unsupported.

>> 1. rails decides to add some validation to polymorphic-STI associations, starts raising error when we call item

It's my understanding that rails expects `item_type` to contain the base class, and furthermore, that rails expects to totally control that column. If we tamper with that column, doesn't it "void our warranty"? If there is any official documentation to the contrary, please correct me.

> Amongst the extensive set of Rails production code worldwide it is no doubt fairly common to come across projects with a polymorphic association referencing a model that can be subclassed via STI. As such, we're in good company, I would expect.

Yes, there are surely many applications with polymorphic-STI associations, but they depend on rails to control the polymorphic foreign key (eg. `item_type`).  Once we tamper with the `item_type` column, once we store something *different than we're supposed to*, we are no longer in that good company.

That's why I suggested adding another column:

> What if we leave item_type as-is and add another column called e.g. item_reification_type?

What are your concerns about that suggestion? Thanks!

lorint:
> ... rails expects item_type to contain the `base class` ...

The first line of `#find_sti_class` in inheritance.rb seems to support the real class, base class, or anything potentially in-between:

    def find_sti_class(type_name)
      type_name = base_class.type_for_attribute(inheritance_column).cast(type_name)
      subclass = begin
        if store_full_sti_class
          ActiveSupport::Dependencies.constantize(type_name)
        else
          compute_type(type_name)
        end
      rescue NameError
    ............ Message about subclass trouble
      end
      subclass
    end

Reading through test/cases/inheritance_test.rb further seems to confirm this.

> ... rails expects to totally control that column (`item_type`).

Glad we continue to allow Rails to do just that.  Perhaps you're unaware that in the last few commits in this PR `record_trail.rb` does not play with `item_type` or `item_id` any more than it used to.  (Still in `data_for_destroy` those two are messed with, just as before.  I'm eager to get to a state where that's not necessary.)  I guess I'd invite you to approach the thought that in this PR we are not directly setting `item_type` in any new ways, and remain completely compatible with all things ActiveRecord.

We're eager to get to a point that reification through associations works well, and to also accurately track item_type with no new columns.  The current app we're working on is projected to allow searching through tens of millions of audit entries by this time next year.  While adding a new column to store the true class name for items may sound trivial, it has some impact on the overall performance of a larger system.  We greatly respect much of the approach taken with this gem, but just as not everyone needs special InnoDB performance tweaks on MySQL, many folks may just use normal MySQL and be happy.  But for some the "Percona" release is essential.  We are brought to considering creating essentially a "Percona for PaperTrail", supporting all aspects of STI + reify.  Not everyone needs this, but we are dearly interested in having this kind of solution.

westonganger:
By not supporting STI, it does effect the ability to properly do OOP in a larger app. My opinion, STI has been a long standing issue with PT-AT and a solution should be sought after. If its compatible with all currently tested versions of Rails and the `gemspec` is locked down pretty tight then I think this should be considered. Can it get really get that much more broke than it is?

That being said adding an extra column as a safe-measure is not a terrible idea as it still solves the problem and is still "safe". However I want to say avoid the extra column but thats my un-expert opinion.

westonganger:
Is it possible that both approaches could be supported? So by default it would use an extra column. If the extra column is not available in the database then fallback to this PR's approach?

jaredbeck:
In our `Doctor < Person` example, I believe rails would store the string "Person" in `item_type` and this PR would store the string "Doctor". Is this correct?

lorint:
> rails would store the string "Person" in item_type, and this PR would store the string "Doctor"

The code you maintain and the code I propose both are pure Rails, so perhaps it's more accurate to say that the current gem would store "Person", and this PR stores "Doctor".  Please with an open mind look at the two sets of code below from `RecordTrail#record_create` and try to pick which one would somehow break Rails.

Option 1:

    version = versions_assoc.new data_for_create
    version.save

Option 2:

    version = versions_assoc.create! data_for_create

Is there any appreciation for the minor change we're proposing?  Can you imagine Rails not supporting either of these approaches?

Our team would like to indicate that the first approach is more useful for two reasons: (a) -- when searching for specific item_types it's much faster since you don't have to worry about digging up some inheritance_column out of object or object_changes. (b) -- even better for you guys who maintain PaperTrail gems, it makes reification much more straightforward.  In fact the old person_spec.rb examples with BMW 330 and BMX 2.0 pass with no other modifications.  Undo that base_class PR that was put in back in December, and it all works.

> Is it possible that both approaches could be supported? So by default it would use an extra column. If the extra column is not available in the database then fallback to this PR's approach?

It's tougher for me to get behind this solution because benefit (b) from above is not available.  We don't gain the reification benefits.  While it would be possible to make everything work, it would be more goofing around during reification of historic records to dig up the actual class.  That is unless we did directly modify item_type, and had the extra column storing base_class.name.  Truly though I would prefer for ActiveRecord to control that column.

jaredbeck:
> My opinion, STI has been a long standing issue with PT-AT and a solution should be sought after.

We're going to fix this, we just have to do it in a way that will work now, *and in future versions of rails.*

> .. the current gem would store "Person", and this PR stores "Doctor".

This is the change, the **departure from standard rails behavior**, that concerns me. I am concerned that future versions of rails will not support this. You've given a really nice analysis of the benefits of this approach, eg. cleaner code in PT, but those benefits don't matter if it suddenly starts throwing errors in a future version of rails.

Also, it is not sufficient to analyze existing versions of rails. Just because something works in the current version, doesn't mean it will in the future. Documentation, for example, would give an indication of future support. A promise of support from a rails team member would help. Your analysis (accurate and well-written) of existing functions in rails, does not help.

If you can't cite documentation, or the word of rails team member, then let's discuss the two scenarios I've given, in which a future version of rails might not be compatible with the suggested solution. I am still worried about both scenarios. How can I help articulate these scenarios better, or in more detail?

I appreciate your continued participation in this discussion, and your patience with my concerns.

jaredbeck:
I can't stop thinking about this PR, argh :) I should probably take a break from it, but I'm going to research association scopes (eg. passing a lambda to `has_many`) and try to understand better how likely (or not) some of my concerns are. I'm hoping to find some documentation like "this is why associations can take lambdas, here's some official docs on when to use lambdas" etc.

lorint:
Regarding having lambda options in a has_many, which was introduced with Rails 4.0, of particular interest for us is this test from `test/cases/associations/has_many_associations_test.rb` (currently around line 2395 in the Rails 6 code base) which showcases using #unscope in a has_many:

    test "can unscope and where the default scope of the associated model" do
      Car.has_many :other_bulbs, -> { unscope(where: [:name]).where(name: "other") }, class_name: "Bulb"
      car = Car.create!
      bulb1 = Bulb.create! name: "defaulty", car: car
      bulb2 = Bulb.create! name: "other",    car: car
  
      assert_equal [bulb1], car.bulbs
      assert_equal [bulb2], car.other_bulbs
    end

In this PR we are doing exactly this against item_type in order to correctly reify subclassed items.  Based on having an example like this, if any of the #has_many behaviour we rely upon were to be deprecated then I would expect only simple changes to be necessary in our code.

lorint:
Having lambda options in a #has_many was introduced with Rails 4.0, along with #unscope as well.  Of particular interest for us around combining these two is [this test from `test/cases/associations/has_many_associations_test.rb`](https://github.com/rails/rails/blob/9e599a281e3d911eccc683a316a4880294b3bc09/activerecord/test/cases/associations/has_many_associations_test.rb#L2395-L2403) which showcases using #unscope in a has_many:

    test "can unscope and where the default scope of the associated model" do
      Car.has_many :other_bulbs, -> { unscope(where: [:name]).where(name: "other") }, class_name: "Bulb"
      car = Car.create!
      bulb1 = Bulb.create! name: "defaulty", car: car
      bulb2 = Bulb.create! name: "other",    car: car
  
      assert_equal [bulb1], car.bulbs
      assert_equal [bulb2], car.other_bulbs
    end

It was [committed by Rafael Franca in October 2015](https://github.com/rails/rails/commit/1cfcf414a5919ec3fd9e6fb01bfaef62e5ff46e5) to explore related cases around fixing an issue with #rewhere, and remains unchanged since then.

In this PR we are doing exactly what this example showcases to target `item_type` in order to correctly reify subclassed items.  Based on having an example like this remain unchanged for two and a half years, if any of the #has_many behaviour this PR relies upon were to be deprecated then I would expect a graceful timeframe in which we could update PT.

jaredbeck:
>> Documentation, for example, would give an indication of future support.

> Of particular interest for us .. is [this test from `test/cases/associations/has_many_associations_test.rb`](https://github.com/rails/rails/blob/9e599a281e3d911eccc683a316a4880294b3bc09/activerecord/test/cases/associations/has_many_associations_test.rb#L2395-L2403) which showcases using #unscope in a has_many:

Nice find. That goes a long way towards my concerns about the `has_many`. I'll give this some more thought later this week.

jaredbeck:
I think I found another good one:

```
# activerecord/test/cases/associations/belongs_to_associations_test.rb:296
  def test_building_the_belonging_object_with_sti_subclass
    account = Account.new # Account belongs_to :firm, class_name: "Company"
    company = account.build_firm(type: "Firm") # Firm < Company
    assert_kind_of Firm, company, "Expected #{company.class} to be a Firm"
  end
```

It's just STI, not polymorphic-STI, but it seems to demonstrate that rails is OK with overriding the `type` column.

jaredbeck:
I'm feeling positive about rails supporting this in the future.

My only remaining concern is about migrating existing records. Are we going to allow people to have a "mixed history", or are we going to require them to update existing records in `versions`?

**Example 1:** `Doctor < Person`. I record Person, Person, Person, then I update PaperTrail. Now I record Doctor, Doctor. I call `versions`, which will `select .. where item_type = 'Doctor'` and I only find the two most recent versions. Also, `versions.count` will return `2`.  I think people will see that as a bug.

**Example 2:** Apply this patch to your new test and it will fail:

```diff
diff --git a/spec/models/pet_spec.rb b/spec/models/pet_spec.rb
index 4c75fc1..54f7489 100644
--- a/spec/models/pet_spec.rb
+++ b/spec/models/pet_spec.rb
@@ -64,6 +64,7 @@ RSpec.describe Pet, type: :model, versioning: true do
     # in the old way, set one of our versions to be "Animal" instead of "Cat".
     versions = PaperTrail::Version.order(:id)
     versions.second.update(item_type: cat.class.base_class.name)
+    expect(cat.versions.count).to eq(4)
 
     animal = Animal.create                    # Index 4
     animal.update(name: "Muppets Drummer")    # Index 5
```

lorint:
> Are we going to allow people to have a "mixed history", or are we going to require them to update existing records in `versions`?

Good plan to have a generator to build a migration.  Over the past few days I've updated the item_type update script we had used internally and put it into the generator "update_sti".  It's now been made more versatile to cover a range of scenarios.  The "secret sauce" of it all is the ability to define a specific inheritance_column for an existing base class item_type across a range of version IDs.  And you can make as many of these as you like, tucked right into a migration file.  Was kinda fun to write the test for that one!

(Pay no mind to the red X -- all tests actually pass; strange timeout snag on one of the Travis scenarios.)

CHANGELOG.md
```diff
@@ -21,11 +21,14 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
...
   [paper_trail-hashdiff](https://github.com/hashwin/paper_trail-hashdiff)
+- [#1108](https://github.com/paper-trail-gem/paper_trail/pull/1108) -
+  Generator `update_sti` builds a migration to update `item_type` for existing versions.
 
 ### Fixed
 
 - [#1108](https://github.com/paper-trail-gem/paper_trail/pull/1108) -
-  Properly reifying a version of a subclassed STI model requires item_type to represent the actual class name instead of base_class.
+  Properly reifying a version of a subclassed STI model requires `item_type` to represent
+  the specific class name instead of base_class.
```
jaredbeck: Please rebase and move your changelog entries now that 9.2.0 has been released.

README.md
```diff
@@ -939,8 +939,13 @@ class Banana < Fruit
 end
 ```
 
-However, there is a known issue when reifying [associations](#associations),
-see https://github.com/paper-trail-gem/paper_trail/issues/594
+A change in what `item_type` stores for subclassed models was introduced in v9.2, recording
```
jaredbeck: 9.2 has been released already, so please reference the PR instead of a version number.

README.md
```diff
@@ -983,6 +989,58 @@ Runtime options:
...
 ```
 
+##### paper_trail:update_sti
+
+The second generator, `paper_trail:update_sti`, also writes but does not run a migration file
+which scans existing data in the versions table to identify all use of STI. Upon finding each
+entry which refers to an object from a subclassed model, `item_type` is updated to reflect
+the subclass name, providing full compatibility with how the `versions` association works
+in >= v9.2. In order to more quickly update a large volume of version records, updates are
+batched to affect 100 records at a time.
```
jaredbeck: Excellent docs. And, I see this section is referenced from the changelog. Can we make that into a link?

README.md
```diff
@@ -983,6 +989,58 @@ Runtime options:
...
+
+Hints can be used in rare cases when the STI structure is modified over time such as
+when establishing additional intermediary inheritance structure, pointing to a new
+base class, abandoning STI entirely, or changing the name of the inheritance_column. The
+vast majority of users will probably never need to supply hints when using this generator.
```
jaredbeck: 😮 fancy!

lib/generators/paper_trail/install/templates/create_versions.rb.erb
```diff
@@ -3,7 +3,7 @@
 class CreateVersions < ActiveRecord::Migration<%= migration_version %>
 
   # The largest text column available in all supported RDBMS is
-  # 1024^3 - 1 bytes, roughly one gibibyte.  We specify a size
+  # 1024^3 - 1 bytes, roughly one gigabyte.  We specify a size
```
jaredbeck: "gibibyte" is the correct term for 1024^3 bytes. (https://en.wikipedia.org/wiki/Gibibyte) "gigabyte" would be 1000^3 bytes.

lib/generators/paper_trail/update_sti/USAGE
```diff
@@ -0,0 +1,2 @@
+Description:
+  Generates (but does not run) a migration to update item_type for STI entries in an existing versions table.
```
jaredbeck: Nice idea to have a USAGE file here, but please add link to full docs in readme.

lib/generators/paper_trail/update_sti/update_sti_generator.rb
```diff
@@ -0,0 +1,17 @@
...
+
+require_relative "../migration_generator"
+
+module PaperTrail
+  # Updates STI entries for PaperTrail
+  class UpdateStiGenerator < MigrationGenerator
+    source_root File.expand_path("templates", __dir__)
+
+    desc "Generates (but does not run) a migration to update item_type for STI entries in an "\
+      "existing versions table."
```
jaredbeck: Can we link to the full docs in the readme here, also?

spec/models/pet_spec.rb
```diff
@@ -1,13 +1,14 @@
...
 require "spec_helper"
+require "rails/generators"
 
 RSpec.describe Pet, type: :model, versioning: true do
   it "baseline test setup" do
     expect(Pet.new).to be_versioned
   end
 
-  it "can reify successfully" do
+  it "can be reified" do
```
jaredbeck: Oh, this is a pet peeve of mine also, the unnecessary use of the word "successful". It often shows up in flash messages, like "Changes saved successfully". Well .. yeah! 🤓

spec/models/pet_spec.rb
```diff
@@ -54,31 +55,109 @@
...
+    let(:cat) { Cat.create(name: "Garfield") }   # Index 0
+    let(:animal) { Animal.create }               # Index 4
+    let(:migrator) { ::PaperTrailSpecMigrator.new }
+
+    before do
+      # This line runs the `let` for :cat, creating two entries
+      cat.update_attributes(name: "Sylvester")   # Index 1 - second
+      cat.update_attributes(name: "Cheshire")    # Index 2 - third
+      cat.destroy                                # Index 3 - fourth
+      # With PT <= v9.2 a subclassed version's item_type referred to the base_class, but
```
jaredbeck: Again, 9.2 has already been released, so if you have to reference something, please just reference this PR.

spec/spec_helper.rb
```diff
@@ -49,10 +49,12 @@ def params_wrapper(args)
...
 # Migrate
+# First remove any dummy or other stray migrations built from `rails g update_sti`
 require_relative "support/paper_trail_spec_migrator"
-::PaperTrailSpecMigrator.
-  new(::File.expand_path("dummy_app/db/migrate/", __dir__)).
-  migrate
+migrator = ::PaperTrailSpecMigrator.new
+migrator.delete("*_dummy_migration.rb")
+migrator.delete("*_update_versions_for_sti.rb")
+migrator.migrate
```
jaredbeck: Would it be cleaner to move the deletions into the migrator class? Would we ever want to migrate without deleting them?

spec/models/pet_spec.rb
```diff
@@ -54,31 +55,109 @@
...
+      # When using the has_many :versions association, it only finds versions in which
+      # item_type refers directly to the subclass name.
+      expect(cat.versions.count).to eq(3)
+      # To have has_many :versions work properly, you can generate and run a migration
+      # that examines all existing models to identify use of STI, then updates all older
+      # <= v9.2 version entries so they refer to the proper subclass.
+      # (This is the same as running:  rails g paper_trail:update_sti; rails db:migrate)
+      expect do
+        migrator.generate_and_migrate("paper_trail:update_sti")
+      end.to output(/Associated 1 record to Cat/).to_stdout
```
jaredbeck: I appreciate you testing the migration. Currently it is failing CI. I tried re-running the [job](https://travis-ci.org/paper-trail-gem/paper_trail/jobs/405184690) a few times, but it failed 4/5 runs. Don't worry if you can't get the expectation re: output passing, it's the data that matters :)

spec/support/paper_trail_spec_migrator.rb
```diff
@@ -1,19 +1,60 @@
...
+        File.delete(Rails.root.join(file))
+      end
+    end
+    if dummy_version
+      File.delete(@migrations_path.join("#{dummy_version}_dummy_migration.rb"))
+    end
+    # Wait up to 3 seconds for all migrations to complete
+    attempt = 0
+    until ActiveRecord::SchemaMigration.exists?(version: last_version)
+      sleep 0.1
```
jaredbeck: I don't understand, are the migrations running in a background process or something?

spec/models/pet_spec.rb
lorint: Even without this output matcher in place, the association still hasn't properly taken place, so the later expect() will fail.  Have tracked this down to most likely being multiple Travis processes interfering with files in db/migrate (a) as processes are starting up and (b) when these generator tests are run.  Have created an advisory locking approach to deal with this concurrency which waits for up to 5 minutes for other processes to finish their things, and then (somewhat impatiently since perhaps another process has hung) jumps in and just takes an exclusive lock.

spec/support/paper_trail_spec_migrator.rb
lorint: Travis does much to separate concurrently-running processes, but all of them do share a common file system.  So ... I have now put together an advisory lock to keep the peace.

spec/spec_helper.rb
lorint: If my new generator tests fail in somewhat interesting ways then stray files can be left in db/migrate, and I want to make absolutely sure that nothing extra is left around.

jaredbeck: Sorry, I don't understand. Why can't `PaperTrailSpecMigrator#migrate` perform deletion?

spec/models/pet_spec.rb
jaredbeck: >  Have created an advisory locking approach to deal with this concurrency which waits for up to 5 minutes for other processes to finish their things ..

Oh dear .. this complexity may outweigh the benefit of testing the migration.

Perhaps the meat of the migration could be moved into a plain ruby class that is easier to test?

spec/support/paper_trail_spec_migrator.rb
jaredbeck: Which processes are running concurrently?

spec/spec_helper.rb
lorint: Well, `#migrate` certainly does, but in addition I like to make sure of things.  Picture a case where an example for a generator is interrupted or times out.  I want there to be no risks in this scenario.

spec/support/paper_trail_spec_migrator.rb
lorint: The very nature of Travis CI is that from the generated test matrix there are about 5 processes running concurrently.  (All the spinning yellow infinity symbols.)

spec/models/pet_spec.rb
lorint: We create applications in which the integrity and security of customer data is mission-critical.  A proven test methodology is essential for this, and I will pay whatever price it takes in order to confirm a reliable system.

The locking strategy now in place here may allow generated migrations to always be reliably tested.  That's the goal anyway.  And for me it's not just for use here in PaperTrail with Travis, but also with our larger-scale projects that run on other CI tools with MANY more cores concurrently.

README.md
lorint: Thanks much.  Have put in a link simply to 5.c.  Can be made more specific perhaps.

jaredbeck:
> The very nature of Travis CI is that from the generated test matrix there are about 5 processes running concurrently. (All the spinning yellow infinity symbols.)

But, we are using the "container-based" infrastructure. (https://docs.travis-ci.com/user/for-beginners/#Infrastructure-and-environment-notes) So, each "job" runs in a separate container, right? There is no way for a process in one container to talk to a process in another container.

jaredbeck:
There's something funny going on with the diff here. I'm looking at "Files Changed" and seeing changes in files like ` .github/CONTRIBUTING.md`. Can you please rebase on master, squash all of your commits, and review "Files Changed" to make sure it only contains your work? Thanks.

lorint:
> Can you please rebase on master, squash all of your commits, and review "Files Changed"

Surely.  Now rebased onto master, and things look good.

> we are using the "container-based" infrastructure. (https://docs.travis-ci.com/user/for-beginners/#Infrastructure-and-environment-notes) So, each "job" runs in a separate container

Ah ... hmmm.  I can't explain why everything was dandy locally but very intermittent on TravisCI.  Have merged your Events PR, removed the Advisory lock code, and all tests still passing.

jaredbeck:
> .. removed the Advisory lock code ..

Great. I'm looking forward to reviewing this, hopefully this weekend. Thanks.

jaredbeck:
In addition to the changes to `PaperTrailSpecMigrator` I described above, I've added a few tests, made minor changes to the docs, and added a `warn`ing for the people who don't read changelogs.

I'm ready to merge this, but I'd welcome a review of my recent changes. Thanks.

lib/paper_trail/record_trail.rb
```diff
@@ -313,14 +323,17 @@ def update_columns(attributes)
...
+    # indicates that rails should do whatever it normally does.
     def versions_association_item_type
       type_column = @record.class.inheritance_column
-      type_name = (respond_to?(type_column) ? send(type_column) : nil) ||
+      item_type = (respond_to?(type_column) ? send(type_column) : nil) ||
         @record.class.name
-      if type_name == @record.class.base_class.name
-        ""
+      if item_type == @record.class.base_class.name
+        nil
```
lorint: (In connection with the test for .blank? from 184)
Not sure how much performance loss might be experienced, but the blank was an attempt to avoid the unscope / scope when item_type referred to the base_class.  For many STI implementations perhaps it doesn't matter because the majority of things might be subclassed, but in some cases it could provide a benefit.  (Family / CelebrityFamily kind of examples.)

lib/paper_trail/record_trail.rb
```diff
@@ -369,6 +382,23 @@ def whodunnit(value)
...
+
+      # Is this class an STI subclass?
+      record_class = @record.class
+      return if record_class.descends_from_active_record?
+
+      # Have we already issued this warning?
+      ::PaperTrail.config.classes_warned_about_sti_item_types ||= []
+      return if ::PaperTrail.config.classes_warned_about_sti_item_types.include?(record_class)
+
+      ::Kernel.warn(format(E_STI_ITEM_TYPES_NOT_UPDATED, record_class.name))
```
lorint: I love this message!  Thanks for putting this part together.

lorint:
> Your solution for the migration name timestamp conflict was very clever, but I'd like to replace it with ...
(hehe, will not mention our silliness)

Once I better understood that caching issue, you might have seen that I had put in a sleep(2).  For a good day I had toyed around with LOTS of different options, even a temporary monkey patching of Rails just for that test!  Boy was that messy.

After thinking more about it, another option we might consider is to use TimeCop.  But I do like the approach you've arrived upon.

jaredbeck:
> One note is that in RecordTrail#versions_association_item_type it previously returned a blank string if the class type was the same as the base_class. This avoided the overhead of always having to .unscope() / .where(). I wonder about bringing that part back, but must admit that I haven't actually benchmarked how much extra CPU time and memory overhead this may or may not cause.

The `unless item_type.nil?` I have written should be exactly the same as the `if item_type.blank?` you had.

I'll go ahead and merge this, and I hope you can help test it by pointing your apps at `master` and running their test suites.

Thanks for your hard work and patience on this.

seanlinsley:
In my application, I have a `Company` model which via STI can become `Management`. There are no extra attributes that exist for management companies; it's purely a distinction that exists for the sake of ActiveRecord associations (a company can either be an owner, a manager, or both).

While this PR fixes issues for those using more complex STI features, it breaks my basic use-case.

A distinguishing factor is that I don't have a `type` column on the table. Perhaps we could continue to use the base class if the table doesn't have a `type` column?

seanlinsley:
In my application, I have a `Company` model which via STI can become `Management`. There are no extra attributes that exist for management companies; it's purely a distinction that exists for the sake of ActiveRecord associations (a company can either be an owner, a manager, or both).

While this PR fixes issues for those using more complex STI features, it breaks my basic use-case.

A distinguishing factor is that I don't have a `type` column on the table. Perhaps we could continue to use the base class if the table doesn't have a `type` column?

lorint:
> I don't have a `type` column on the table.  Perhaps we could continue to use the base class if the table doesn't have a `type` column

First let me say that I like the minimalist approach that is a hallmark of your suggestions -- we have a large-scale app that will benefit from your involvement around object / object_changes.  In this case about using STI without `type` I catch a sense of, "Why have a `type` column if it's not absolutely necessary?"

So that I can be certain about the goal -- here's my attempt at a failing test.  In the dummy app in the test suite, `Customer` only has :id and :name columns.  Let's say that this STI-ish model were added:

    class Management < Customer
    end

Then perhaps this would this describe your expected behaviour:

    it "utilises base_class for STI classes having no type column" do
      expect(Management.inheritance_column).to eq("type")
      expect(Management.columns.map(&:name)).not_to include("type")

      customer1 = Customer.create(name: "Cust 1")
      customer2 = Management.create(name: "Cust 2")
      cust_versions = PaperTrail::Version.where(item_type: "Customer")
      mgmt_versions = PaperTrail::Version.where(item_type: "Management")
      expect(cust_versions.count).to eq(2)
      expect(mgmt_versions.count).to eq(0)
    end

(And if so then I would certainly like to add in examples for **update** and **destroy** as well.)

To enact a fix then perhaps in `record_trail.rb` there could be this:

    def versions_association_item_type
      type_column = @record.class.inheritance_column
      base_class_name = @record.class.base_class.name
      item_type = (@record.respond_to?(type_column) ? @record.send(type_column) : base_class_name) ||
        @record.class.name
      if item_type == base_class_name
        nil
      else
        item_type
      end
    end

Whadya think?

seanlinsley:
In #1137 I ended up fixing the bug (I think), but haven't added a test for it yet. I'd appreciate if you could review that PR 😄

PR #1112, opened 2018-06-17
Add gem version and SemVer compatibility badges

greysteil:
First up, thanks for PaperTrail!

This PR adds a SemVer compatibility badge that displays the percentage of CI runs that pass when updating PaperTrail between SemVer compatible versions. Data comes from Dependabot (disclosure: I built it). I've also added a gem-version badge, which should make it easier for people adding PaperTrail to their Gemfile (I always end up checking for one / looking on Rubygems).

Here's what the badges look like:

[Gem Version](https://rubygems.org/gems/paper_trail) [SemVer](https://dependabot.com/compatibility-score.html?dependency-name=paper_trail&package-manager=bundler&version-scheme=semver)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # PaperTrail
 
 [![Build Status][4]][5]
+[![Gem Version][53]][54]
+[![SemVer][55]][56]
 
 Track changes to your models, for auditing or versioning. See how a model looked
 at any stage in its lifecycle, revert it to any version, or restore it after it
@@ -1542,3 +1544,7 @@ Released under the MIT licence.
 [50]: https://github.com/izelnakri/paper_trail
 [51]: https://github.com/rikkipitt/rails_admin_history_rollback
 [52]: http://guides.rubyonrails.org/active_record_callbacks.html
+[53]: https://badge.fury.io/rb/paper_trail.svg
+[54]: https://rubygems.org/gems/paper_trail
+[55]: https://api.dependabot.com/badges/compatibility_score?dependency-name=paper_trail&package-manager=bundler&version-scheme=semver
+[56]: https://dependabot.com/compatibility-score.html?dependency-name=paper_trail&package-manager=bundler&version-scheme=semver


jaredbeck:
> .. a gem-version badge, which should make it easier for people adding PaperTrail to their Gemfile (I always end up checking for one / looking on Rubygems).

This could be useful, I suppose.

> .. a SemVer compatibility badge that displays the percentage of CI runs that pass when updating PaperTrail between SemVer compatible versions. Data comes from Dependabot (disclosure: I built it). 

Interesting. This seems like a poor measure of SemVer, though. CI is not capable of detecting, e.g. feature removal.

greysteil:
> This seems like a poor measure of SemVer, though. CI is not capable of detecting, e.g. feature removal. For example, the tests for that feature could also be removed.

CI should be pretty good at detecting that, no? If any of your users were using that feature and had any tests around it then their CI will blow up when they try to upgrade without making change.

jaredbeck:
> If any of your users were using that feature and had any tests around it then their CI will blow up when they try to upgrade without making change.

Oh, you're aggregating data from projects that *depend on* PaperTrail? If so, that's pretty neat. I thought your data was coming from PaperTrail's own CI.

greysteil:
Yep! Sounds like I could make that clearer in the text - will redraft later today.

jaredbeck:
Cool, I'll give it a try, thanks. So the data comes from the update PRs that Dependabot authors? 

One thought: people have intermittently failing test suites. Those suites will mess up the percentage (make it look worse than it is). You could give such suites a lower weight, perhaps.

greysteil:
Awesome! Yep - data is from PRs that Dependabot authors, filtering out any where the tests were originally failing, or where the update is between major versions or involves pre-releases.

Flaky specs are definitely an issue. I think there should be enough data in Dependabot to filter them out, though, pretty much exactly as you describe. I'll take a look at doing that in the next couple of days.

Thanks for giving it a try!

PR #1120, opened 2018-07-22
Extract `Events` module

jaredbeck:
Introduces event classes, like Events::Create, which wrap the
AR instance and have one public method, `#data`, which returns
the attributes of nascent `Version` records. See comments in
Events::Base.

This extraction greatly simplifies RecordTrail. Not only have
we moved a lot of code out of RecordTrail, but we've also
moved the `@in_after_callback` variable. In its new location, this
variable is much easier to understand, and we no longer need the
`ensure` blocks to reset it.

Preserving compatibility with PT-AT caused some minor difficulties,
and we should follow up with them to see if we can simplify the
API they are using. Specifically, we would like to delete the
`data_for_*` methods.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -1,0 +1,273 @@
+# frozen_string_literal: true
+
+module PaperTrail
+  module Events
+    # We refer to times in the lifecycle of a record as "events". There are
+    # three events:
+    #
+    # - create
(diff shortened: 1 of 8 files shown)


jaredbeck:
This PR improves the [flog](https://rubygems.org/gems/flog) score of record_trail.rb from 461 to 270. :tada:

Issue #1099, opened 2018-06-01
Option to not save the current state of the object


volkanunsal:
**Is your feature suggestion related to a problem? Please describe.**

I'd like to use PaperTrail for an event sourcing project. I only want to keep the object_changes and I don't want to store the current state of the object in the events table.

**Describe the solution you'd like to build**

I'd like a configuration option that would tell PaperTrail to skip the `object` column, similar to the option to not record `object_changes`

```ruby
has_paper_trail  save_object: false
```

**Describe alternatives you've considered**

Monkey patching the library, which I'm already doing.

jaredbeck:
@seanlinsley and I discussed this recently in https://github.com/paper-trail-gem/paper_trail/issues/1046. Can you address some of the concerns I gave there? Of course, Sean suggested omitting `object` only for updates, and you suggest omitting `object` entirely, so your suggestion is simpler. 

How would you handle the situation where someone turned on this config option for one week, then turned it off, then tried to `reify` something? How would you handle any `reify`? It's harder to reify from `object_changes` than from `object` but it could be done.

How would you handle compatibility with the experimental association tracking feature? I think it would be OK to say that this option is incompatible, but we have to raise a helpful error if people try to combine them.

Would you have people drop their `object` column? Requiring the column to be dropped would make people stop and think. It would also make it really clear that the data would come from `object_changes` in the case of a `reify`.

volkanunsal:
@jaredbeck Reification seems to be equivalent to [projection.](https://abdullin.com/post/event-sourcing-projections/) The way to derive the current state is to start from the beginning and process the changes of all events in the order in which they are created. This can be an expensive operation when a record has a lot of changes, so derived state is usually stored in a separate table.

> How would you handle the situation where someone turned on this config option for one week, then turned it off, then tried to reify something?

If reify were to apply changes, rather than using materialized `object` for a version, it would still work. For this to completely work, a prerequisite is for user to opt in to store all events.

> Would you have people drop their object column? 

No, this column can be used like a projection table and increase reification performance.

jaredbeck:
> If `reify` were to apply [incremental] changes, rather than using .. `object` ..

It seems to me we are talking about two possible designs for a versioning library:

1. "snapshots" - what PT currently does, vs.
2. "incremental changes"

So, I feel like we're discussing a big change. In fact, it feels like the biggest and most fundamental change. You can see why I am nervous about it :)

[...]

PR #1123, opened 2018-07-24
Track changes on destroy

seanlinsley:
Part 2 of #1099, followup to #1122.

Rails' `changes` method doesn't return any attributes on destroy, so this PR generates them by hand (though create and update still rely on Rails). In order to still apply serialization to the changes, some of the internal API has been refactored.

- [x] get the tests passing
- [x] decide on the versioning / deprecation plan
- [ ] document the upgrade plan for each path. explain why it's a good idea to use the new behavior
- [ ] write SQL & Ruby migrations as examples for those who want to populate `object_changes` on historical data


jaredbeck:
Regarding the config option,

```ruby
PaperTrail.config.record_changes_on_destroy = true
```

You were planning on doing a normal deprecation cycle, which would normally be the right call,


>
> 1. in a minor version: deprecate save_changes, and introduce a setting to enable object_changes on destroy
> 2. in the next major version: remove save_changes, and track changes on destroy by default

but the next release will be a major version, so I'd rather just start always tracking changes on destroy, and not bother with the intermediate step of having a config option.

I don't understand the rationale for removing `save_changes` though. 

> I think #507 which introduced save_changes was misguided: they were trying to reduce the amount of space used by a large text column, but didn't seem to realize that object would duplicate that column on every save.
>

Right, but people might want to keep `save_changes` so they can have it `true` for some models and `false` for others.

seanlinsley:
When trying to save disk space, `save_changes` is the wrong option b/c even with it enabled, every attribute is duplicated in the object column every time a change occurs. Because of that, there's orders of magnitude more space to save by dropping the object column. I think providing both as options would confuse readers who are trying to optimize disk usage.

Do we know if anyone is using `save_changes`? If it weren't for the plan to have the next release be a major version bump, I'd be interested in adding a deprecation warning that'd ask users to reply on GitHub if they're using the feature.

seanlinsley:
I updated this PR with a refactor of the serialization code to work around the fact that Rails doesn't provide `changes` on destroy. It seems to have broken some tests though, and I'm not sure why 🤔

seanlinsley:
Ah, the song model has this code in it: https://github.com/paper-trail-gem/paper_trail/blob/3f9b5be4ebc3f4bbb251e9e1bbbb7ddf8b1fae60/spec/dummy_app/app/models/song.rb#L21-L31

So we should be expecting a length of 240 to be saved in every case. But with Rails 4.2, a length of 4 is incorrectly being stored.

That might've made sense on the destroy event b/c this PR calls `public_send(attr)` instead of `read_attribute(attr)`, but the test failure occurs on the create event, not the destroy event.

seanlinsley:
In that case a deprecation release is probably overkill since there's a clear path if they want the old behavior. Though we'll need to document this in the upgrade guide:
- what are the options, and which should I choose?
- what do I need to do in each case? (with code samples)

We also need to figure out this test failure. AFAICT, it's signalling a bug in Rails 4.2. Should I just update the test to use a different model?

jaredbeck:
I can try to help with the `Song` test, but I'm confused about the `record_changes_on_destroy` config option. You're going to remove that, right?

seanlinsley:
Yes, sorry, that'd be removed.

seanlinsley:
Okay this should be complete, code-wise. I also added a commit to remove `save_changes`.

I'm planning on testing these changes in my application next week. If this ends up getting merged before then, I'll submit another PR with upgrade steps (particularly, a migration to populate data on old destroy events).

jaredbeck:
I like how you've split this into two commits, but can you please split it into two (new) PRs? I have separate comments for each commit and I think they'll be easier to discuss as separate PRs. Thanks.

seanlinsley:
Sure. I broke it out into #1130. Now to update this PR...

lib/paper_trail/events/create.rb
```diff
@@ -20,7 +20,9 @@ def data
           data[:created_at] = @record.updated_at
         end
         if record_object_changes? && changed_notably?
-          data[:object_changes] = recordable_object_changes(changes)
+          changes = serialize_changes(notable_changes)
+          changes = recordable_object_changes(changes)
+          data[:object_changes] = changes
```
jaredbeck: I'm concerned this changes the order of operations. Previously, `object_changes_adapter` happened first, before `AttributeSerializers::ObjectChangesAttribute`. Now it happens after. Was that intentional?

lib/paper_trail/events/destroy.rb
```diff
@@ -21,6 +21,13 @@ def data
         if record_object?
           data[:object] = recordable_object(false)
         end
+        if record_object_changes?
+          # Rails' implementation returns nothing on destroy :/
+          changes = @record.attributes.map { |attr, value| [attr, [value, nil]] }.to_h
+          changes = serialize_changes(changes)
+          changes = recordable_object_changes(changes)
+          data[:object_changes] = changes
+        end
```
jaredbeck: This method is getting a bit long. I think we can extract some part of this into a new method. Maybe something like

```ruby
if record_object_changes?
  data[:object_changes] = object_changes
end
# ..

private

# Rails' implementation returns nothing on destroy, so we build our own hash.
def object_changes
  changes = @record.attributes.map { |attr, value| [attr, [value, nil]] }.to_h
  changes = serialize_changes(changes)
  recordable_object_changes(changes)
end
```

What do you think?

spec/paper_trail/model_spec.rb
```diff
@@ -264,11 +264,18 @@
...
 
-        it "not have changes" do
-          widget = Widget.create(name: "Henry")
-          widget.update_attributes(name: "Harry")
-          widget.destroy
-          expect(widget.versions.last.changeset).to eq({})
+        it "has changes" do
+          book = Book.create! title: "A"
+          changes = YAML.load book.versions.last.attributes["object_changes"]
+          expect(changes).to eq "id" => [nil, book.id], "title" => [nil, "A"]
```
jaredbeck: The predominant code style for this repo uses an additional set of parentheses for the `eq` matcher, and sometimes for the `to` as well.

```ruby
expect(changes).to eq("id" => [nil, book.id], "title" => [nil, "A"])
```

In my experience, using parentheses for neither is hard for many people to follow.

spec/paper_trail/model_spec.rb
```diff
@@ -264,11 +264,18 @@
...
+
+          book.update! title: "B"
+          changes = YAML.load book.versions.last.attributes["object_changes"]
+          expect(changes).to eq "title" => %w[A B]
+
+          book.destroy
+          changes = YAML.load book.versions.last.attributes["object_changes"]
+          expect(changes).to eq "id" => [book.id, nil], "title" => ["B", nil]
```
jaredbeck: This test looks sufficient, but it'd be great to also have unit tests in eg. `spec/paper_trail/events/destroy_spec.rb`. Not a requirement, use your judgement.

lib/paper_trail/events/create.rb
seanlinsley: Hmm, that's not how I'm reading it. Previously we had this code:
```ruby
data[:object_changes] = recordable_object_changes(changes)
```
`changes` used to find the notable changes and serialize them. then `recordable_object_changes`, the last operation, is what called `object_changes_adapter`.

lib/paper_trail/events/destroy.rb
seanlinsley: The previous level of abstraction we had in `changes` clearly wasn't right, since both the update and destroy events would customize the underlying changeset.

Currently every event has these two lines in common w/out any modifications:
```ruby
changes = serialize_changes(changes)
changes = recordable_object_changes(changes)
```

What if we introduced a method to wrap those?

```ruby
# destroy.rb
if record_object_changes?
  changes = @record.attributes.map { |attr, value| [attr, [value, nil]] }.to_h
  data[:object_changes] = prepare_object_changes(changes)
end

# base.rb
def prepare_object_changes(changes)
  recordable_object_changes(serialize_changes(changes))
end

spec/paper_trail/model_spec.rb
seanlinsley: It doesn't look like the unit tests for the various events currently cover `object` or `object_changes`. There's an argument to be made that a lot of the tests in `mode_spec.rb` should be moved out, but I think that's beyond the scope of this PR. This test is probably enough for now.

seanlinsley:
I went ahead and refactored this PR. Let me know if there's anything else that needs changing.

jaredbeck:
"Expect to be a kind of Time" is the name of my memoirs.

jaredbeck:
Thanks for your work on this, Sean.

> I'm planning on testing these changes in my application next week. 

That'd be great, thanks.

> If this ends up getting merged before then, I'll submit another PR with upgrade steps (particularly, a migration to populate data on old destroy events).

How important is such a migration? Should we wait for it before releasing PT 10?

seanlinsley:
🎉 

In the changelog I noted to check #1099 for a backport migration, so a new release doesn't need to wait on the migration.

PR #1131, opened 2018-08-13
Various 2018-08-13

jaredbeck:


jaredbeck:
@seanlinsley what do you think of 7588889 here?

seanlinsley:
¯\\_(ツ)_/¯ it might be useful if we end up having more tests along those lines, but I liked the succinctness of the previous implementation.

One thing to note is that the test no longer verifies that `object_changes` is being set.

jaredbeck:
Another advantage to this approach is that `model_spec.rb` has a chance of staying under a thousand lines.

> One thing to note is that the test no longer verifies that object_changes is being set.

I'll put that back in, thanks.

PR #1137, opened 2018-08-15
Fix `joins`

### LINKED ISSUE #1135: Unable to join versions association
When upgrading Paper Trail from 8.1.2 to 0e94b68c232253b28178e73d129157a2f68e2954, we started running into a Rails error anywhere we performed a join:

```ruby
>> User.joins(:versions).to_a

ArgumentError: The association scope 'versions' is instance dependent (the scope block takes an argument). 
Preloading instance dependent scopes is not supported.

from gems/activerecord-5.1.5/lib/active_record/reflection.rb:481:in `check_preloadable!'
```

What inside Paper Trail caused this error? Is it something we'll have to upgrade Rails to fix?

seanlinsley:
Fixes #1135 (followup to #1108).

Where should I add a test for my use of STI without a type column? https://github.com/paper-trail-gem/paper_trail/pull/1108#issuecomment-412995507


spec/models/pet_spec.rb
```diff
@@ -112,7 +112,7 @@
 
     # After creating a bunch of records above, we change the inheritance_column
     # so that we can demonstrate passing hints to the migration generator.
-    context "simulate a historical change to inheritance_column" do
+    xcontext "simulate a historical change to inheritance_column" do
```
seanlinsley: I don't totally understand these tests... not sure how to fix them. Any ideas? :octocat:

jaredbeck:
I have looked through this, but I didn't understand any of it, probably because I don't understand what problem you're trying to solve. If you could provide a reproduction of the issue using the [bug report template](https://github.com/paper-trail-gem/paper_trail/blob/master/.github/ISSUE_TEMPLATE/bug_report.md) that would help a lot.

@lorint can you help, please?

seanlinsley:
@jaredbeck yeah I understand it's a complex issue. I added failing tests as a separate commit so you can see the failures on CI.

seanlinsley:
Interesting that with the fix, the tests only pass on Rails 5.2. I wonder how the API changed 🤔

seanlinsley:
That's good to hear. In case it wasn't clear, I had to update those tests b/c the events in `record_trail.rb` were calling `versions.build`, and that apparently overrode the `item_type` that was being passed in. As a consequence, the in-memory association wasn't aware of the new record so `count` was needed to trigger a new database request.

lib/paper_trail/model_config.rb
```diff
@@ -141,17 +149,24 @@ def cannot_record_after_destroy?
...
       klass.has_many(
         klass.versions_association_name,
-        lambda do |object|
+        lambda do
           relation = order(model.timestamp_sort_order)
-          item_type = object.paper_trail.versions_association_item_type
-          unless item_type.nil?
-            relation = relation.unscope(where: :item_type).where(item_type: item_type)
-          end
+          item_type = klass.paper_trail.send(:versions_association_item_type)
```
seanlinsley: it's now a public method, so `send` isn't required here

lorint:
Here's the changes I would recommend:

https://github.com/seanlinsley/paper_trail/pull/1/files

lib/paper_trail/model_config.rb
```diff
@@ -141,17 +149,24 @@ def cannot_record_after_destroy?
...
       )
+
+      # We override the assocation when STI models are created from a base class
+      # in order to use the right `item_type`.
+      klass.singleton_class.prepend(Module.new {
+        def inherited(klass)
+          super
+          paper_trail.send(:setup_versions_association, klass)
+        end
+      })
```
lorint: Love this idea!  Worked out well for the fixes for CelebrityFamily tests and migrations to work.

lib/paper_trail/model_config.rb
```diff
@@ -110,6 +110,14 @@ def version_class
       @_version_class ||= @model_class.version_class_name.constantize
     end
 
+    def versions_association_item_type
+      if @model_class.descends_from_active_record?
+        @model_class.base_class.name
+      else
+        @model_class.name
+      end
+    end
```
lorint: After playing around for a bit it ended up seeming that we need a bit more checks to get total consistency with existing tests.  The most intricate ended up [in destroy.rb](https://github.com/seanlinsley/paper_trail/blob/7489c3f73b4bd08b357078a144a518ef85e86e27/lib/paper_trail/events/destroy.rb#L19-L23).

lib/paper_trail/record_trail.rb
```diff
@@ -84,8 +84,7 @@ def record_create
       # `data_for_create` but PT-AT still does.
       data = event.data.merge(data_for_create)
 
-      versions_assoc = @record.send(@record.class.versions_association_name)
-      version = versions_assoc.new(data)
+      version = @record.class.paper_trail.version_class.new(data)
```
lorint: Weirdly I couldn't get these guys to fully work unless I did:
```version = @record.send(@record.class.versions_association_name).new(data)```

File: lib/paper_trail/model_config.rb (hunk 2 above)
Comment (lorint) [reply lib/paper_trail/model_config.rb]: Looks like the `.unscope` / `.where` can simply happen `if klass.descendants.empty?`

**Comment by lorint** [comment]
> ... events in `record_trail.rb` ... overrode the `item_type` that was being passed in. As a consequence, the in-memory association wasn't aware of the new record so `count` was needed to trigger a new database request.

You describe the bane of my existence.  Although hopefully this last update in [PR #1](https://github.com/seanlinsley/paper_trail/pull/1/files) offers a hopefully unobtrusive solution to all of this.

File: lib/paper_trail/model_config.rb (hunk 4 above)
Comment (seanlinsley) [reply lib/paper_trail/model_config.rb]: I don't believe that'll work for the company / management scenario I described here: https://github.com/paper-trail-gem/paper_trail/pull/1108#issuecomment-412995507

Comment (lorint) [reply lib/paper_trail/model_config.rb]: Oh -- I thought you had already written the tests you were after.  Is this perhaps the goal?  It passes BTW:

```
require "spec_helper"

class Management < Customer
end

RSpec.describe Management, type: :model, versioning: true do
  it { is_expected.to be_versioned }

  it "utilises the base_class for STI classes having no type column" do
    # Customer (and thus Management) have no stored `type` column.
    expect(Management.inheritance_column).to eq("type")
    expect(Management.columns.map(&:name)).not_to include("type")

    customer1 = Customer.create(name: "Cust 1")
    customer2 = Management.create(name: "Cust 2")

    # When creating a Management the version refers to Customer:
    cust_pts = PaperTrail::Version.where(item_type: "Customer")
    expect(cust_pts.count).to eq(2)

    # Nothing gets tracked as being Managment:
    mgmt_pts = PaperTrail::Version.where(item_type: "Management")
    expect(mgmt_pts.count).to eq(0)
  end
end
```

Perhaps I'll throw it in the mix for good measure because to me it seems to addresses what sounded like your suggestion.  But if this isn't the goal then please write a failing test that addresses Company / Management.

seanlinsley: It looks like that test covers the scenario I described; thanks for writing it. Are you sure that it passes when checking `descendants.empty?`? It'd seem like it'd only pass when we rely on `descends_from_active_record?`.

lorint: > Are you sure that it passes when checking `descendants.empty?`?

It works fine with `descendants.empty?`, but I've investigated the other possible override on `creation_attributes` instead now.  It ends up a bit more involved since this more generic class method must check to see if it's dealing with a PaperTrail::Version.  (The other override was actually directly creating a method on Reflection objects.)  Pick whichever one you prefer as both approaches work fine.

jaredbeck:
- Should this block the release of PT 10?
- Do you need my help reviewing this? If so, is it ready for review?

Thanks.

lorint:
> Should this block the release of PT 10?

I would *love* to see this be a part of PT 10 because it brings together a couple great fixes, and gets us well-prepared to bring the association_tracking gem further.

The fully working version with all the bells and whistles is [here](https://github.com/seanlinsley/paper_trail/pull/1), and I feel that it's ready for review.

With that one all existing tests get to stay exactly the same -- except for one simple line added in the dummy app's `callback_modifier.rb`.  It also incorporates the Company / Management STI fix Sean had requested.

jaredbeck:
> I would love to see this be a part of PT 10 ..

Fine by me.

> The fully working version with all the bells and whistles is here, and I feel that it's ready for review.

So what's the plan? Sean merges that (https://github.com/seanlinsley/paper_trail/pull/1) and then I review it here? Sorry for jumping into the middle of the conversation here.

> With that one all existing tests get to stay exactly the same ..

That's the dream 👍

lorint:
> the plan -- Sean merges that and then I review it here?

Yeah, that's what I was hoping for.  I had originally set up as that PR to his branch instead of a review here with the hope to make it easy to incorporate all the little changes that bring everything together.  And if you like what you see out there then it's cool also if you'd like to just make a single commit here that brings all of that in.  Whatever is easy.

lib/paper_trail/model_config.rb
seanlinsley: Ah, interesting. I didn't realize that a monkeypatch on `creation_attributes` is effectively global, while the `collection?` monkeypatch is tied specifically to the versions assocation.

seanlinsley:
I was planning to merge https://github.com/seanlinsley/paper_trail/pull/1 into this branch, likely as a squash commit.

@jaredbeck we could use more eyes on the monkeypatch part of the PR: https://github.com/seanlinsley/paper_trail/pull/1#discussion_r211088213

lib/paper_trail/model_config.rb
lorint: I'll go back to the `.collection?()` override then, which is specific to a `Reflection` object.

PR #1138, opened 2018-08-15
Readme fix: caller.find {} rather than caller.first {}

TylerRick:
(`Array#first` doesn't take a block, so it was the same as just calling `caller.first`.)


jaredbeck:
Thanks Tyler

PR #1143, opened 2018-08-22
New optional column: item_subtype

### LINKED ISSUE #594: Reify on associations fails if using Single Table Inheritance

jaredbeck:
Gives PT-AT the data it needs to fix #594. PT-AT will make changes like:

```diff
# paper_trail_association_tracking/reifiers/has_one.rb
# def load_versions(assoc, model, transaction_id, version_at, base_class_name)
            where("#{version_table_name}.item_type = ?", base_class_name).
+           where("#{version_table_name}.item_subtype = ?", assoc.class_name).
            where("created_at >= ? OR transaction_id = ?", version_at, transaction_id).
```

I have tested the above with the "cars and bicycles" example in `person_spec.rb` and it passes.

Replaces (reverts) #1108, which has proved problematic in #1135, leading to further proposed increases in complexity such as https://github.com/seanlinsley/paper_trail/pull/1. @lorint I know you worked hard on #1108, and I did too, but if this is actually the simpler approach, as I propose, I hope we can both approach this objectively. I've never had to revert something as large as #1108 and I don't suggest it lightly.

I haven't written the docs or the migration to populate `item_subtype`. I would like to get your opinions on this approach first.

I made this an optional column because it is my understanding that this issue (#594) only affects PT-AT users with STI, a very small subset of our users.


lorint:
If you're committed to going in this direction then it might be somewhat moot at this point, but I've pushed an update to Sean's [PR#1](https://github.com/seanlinsley/paper_trail/pull/1) based on your review.

jaredbeck:
1. Lorin, it sounds like you agree that this will give PT-AT the data it needs to solve #594. Is that correct?
2. Sean, can you test your app against this branch? I assume that, because I've reverted #1108, that this branch fixes your issue #1135.

seanlinsley:
I'll test this as soon as I can, but in the meantime these tests should verify the behavior if you'd like to add them to this PR:
- https://github.com/paper-trail-gem/paper_trail/pull/1137/commits/3c08782f3699ed1b440d49fee8eadb42aff85e74
- https://github.com/seanlinsley/paper_trail/blob/cd8a5ca87986cab3df353e4a9b749e1333d5fd45/spec/models/management_spec.rb

seanlinsley:
This PR fixes both issues  🙌

jaredbeck:
> .. in the meantime these tests should verify the behavior if you'd like to add them ..

Tests added.

seanlinsley:
We're now using this branch in production & will likely remove our `object` column next week.

Thank you both for all the time you've put into the upcoming release 💛

jaredbeck:
@lorint does this work for you? Does it provide PT-AT with the data it needs to solve #594? Please let me know if you have any objections. If I don't hear from you, I'll probably merge this weekend.

lorint:
It does have the data we would need, and potentially covers Rick Tyler's interest around the **store_base_sti_class** gem.

It does require a fair bit of rework internally, doing checks around `item_type` and `item_subtype`.  And around 16GB of additional storage requirement for this new column.  But this will get offset by the possible removal of the `object` column in the near future.

We are considering both removing `object` and implementing the `item_type` code from PR#1137, in which case we could backport our fixes that use `item_type` for reification to PT-AT using `item_subtype`.

jaredbeck:
> It does have the data we would need, and potentially covers Rick Tyler's interest around the store_base_sti_class gem.

Great, thanks.

> And around 16GB of additional storage requirement for this new column.

If y'all are concerned about disk space, we could choose to store `item_subtype` only when it differs from `item_type`.

jaredbeck:
> We're now using this branch in production.

I usually delete topic branches immediately after merging, so please switch to `master` ASAP.

lorint:
> If y'all are concerned about disk space, we could choose to store `item_subtype` only when it differs from `item_type`.

For us the disk space is of interest only in light of how far-reaching it is to have a half billion records loaded into RAM at once.  We've got 5 covering indexes that leverage `item_type`, and there's some right magic we're able to unfold for our end users based on this, 2 of these 5 b-tree indexes have `item_type` as their primary index.  Hence any even slight update to behaviour around `item_type` can mean significant impact on a fair range of tests, and around the amount of work the database engine goes through overall.  No worries about extra overhead during CUD operations.  Entire focus is on finding related transactions quickly.  For us in these indexes we will continue to store real class names as opposed to base_class so we can have just one column involved, and maintain consistent sub-second searches on `item_type`.

PR #1147, opened 2018-08-29
Warn if versions table pk is int but not bigint

jaredbeck:


seanlinsley:
I'd be hesitant to make this change in my application, again b/c of disk usage. Theoretically it's simple doubling from 4 bytes to 8 bytes, but I'll need to test it on real data.

jaredbeck:
> Theoretically it's a simple doubling from 4 bytes to 8 bytes, but I'll need to test it on real data.

Yeah. It'll also increase the size of any indexes that include that column.

jaredbeck:
Sean, did you get a chance to test this out? Thanks.

jaredbeck:
btw, [`bigint` is the default in rails as of 5.1](https://github.com/rails/rails/pull/26266), so new installs of PT are using `bigint`.

jaredbeck:
@seanlinsley Can you review this weekend? Otherwise I think I'm just going to merge, seeing as this is the default in rails now.

jaredbeck:
Thinking more about this, to do this right, I'd have to provide a config option so advanced users (`int` or `uuid`) can silence this warning, and I just don't want the complexity of another config option.

Issue #1129, opened 2018-08-12
Discussion: timestamps waste a lot of space


seanlinsley:
After #1046 / #1099, I thought it'd be interesting to see how much space `updated_at` ends up using. Turns out it represents 21% of the space used by `object_changes` (41% for updates).

The total amount is even higher, of course, for those that continue to use the now-optional `object` column.

I put "discussion" in the title b/c I'm not sure what Paper Trail the project should do with this information. Are there scenarios where it's important to preserve historical changes to `updated_at`?

The two options would be to change Paper Trail's default behavior to exclude `updated_at`, or to document this issue, making it easy to resolve.

It seems unlikely that we'd change the default behavior if only b/c of momentum. So: what should the documentation look like? I'd worry that the readme is getting a bit hard to follow w/ tips like this scattered about.

```sql
with data as (
  select event,
    count(*) / 1000000 * 1000000 as count,
    sum(pg_column_size(object_changes)) as changes_size,
    sum(
      pg_column_size(jsonb_build_object('updated_at', object_changes->'updated_at'))
        - pg_column_size('{}'::jsonb)
    ) as updated_at_size
  from versions group by event
)
select event, count,
  pg_size_pretty(changes_size) as changes_size,
  pg_size_pretty(updated_at_size) as updated_at_size,
  round(100.0 * updated_at_size / changes_size, 1) as percent
from data
union all
select '(all)' as event,
  sum(count) as count,
  pg_size_pretty(sum(changes_size)) as changes_size,
  pg_size_pretty(sum(updated_at_size)) as updated_at_size,
  round(100.0 * sum(updated_at_size) / sum(changes_size), 1) as percent
from data;

  event  |  count   | changes_size | updated_at_size | percent 
---------+----------+--------------+-----------------+---------
 create  | 23000000 | 11 GB        | 1339 MB         |    11.9
 destroy |  2000000 | 1740 MB      | 126 MB          |     7.2
 update  | 31000000 | 6633 MB      | 2726 MB         |    41.1
 (all)   | 56000000 | 19 GB        | 4191 MB         |    21.4
```

jaredbeck:
> Are there scenarios where it's important to preserve historical changes to updated_at?

Knowing *when* a change was made is very important, but we can get that information from `versions.created_at`. If you want to explore omitting `updated_at`, I'd be interested to hear what you find. Two potential problems:

[...]

PR #1148, opened 2018-08-29
Allow `object_changes_adapter` to use the default behavior

seanlinsley:
#1129

### DIFF
--- a/README.md
+++ b/README.md
@@ -1226,6 +1226,8 @@ A valid adapter is a class that contains the following methods:
 2. load_changeset: Returns the changeset for a given version object
 3. where_object_changes: Returns the records resulting from the given hash of attributes.
 
+To preserve the default behavior for some of these, don't define them in your adapter.
+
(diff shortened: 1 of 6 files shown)


seanlinsley:
This PR should allow an application to e.g. do this, to exclude `updated_at` globally:
```ruby
class SkipUpdatedAt
  def diff(changes)
    changes.except! 'updated_at'
  end
end

PaperTrail.config.object_changes_adapter = SkipUpdatedAt.new
```

I'm not particularly happy with the tests; I was aiming for minimal changes instead of 100% coverage.

jaredbeck:
I wonder if, for this simple case, you could use `has_paper_trail(skip: 'updated_at')`. Of course, `skip` will prevent the creation of a version if `updated_at` is the only thing that changed, but maybe that's OK for your use-case?

seanlinsley:
I'd rather not have it be a per-model config, b/c that means new devs on the team are going to forget to add it.

PR #1152, opened 2018-09-05
Add missing `:string` column type

gaganawhad:
Documentation is missing the `:string` column type for when adding column `sub_type` on `versions`


jaredbeck:
Thanks

PR #1155, opened 2018-09-18
Update references to `PaperTrail.request.controller_info` in README

abrom:
`PaperTrail.controller_info` was removed in v10 so README needs to reference the request method instead.


jaredbeck:
Thanks Andrew

PR #1156, opened 2018-09-25
Provide Gem-Level Option To Ignore timestamp_update_attr Only Changes

MohammadHabbab:
Actions like #touch in active record models, in some cases, only yield an `updated_at` value change (or whatever the timestamp_update_attribute name is). As a developer, it would be useful to provide the option to skip generating/inserting versions records for these updates.

### DIFF
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -21,7 +21,8 @@ module PaperTrail
       :association_reify_error_behaviour,
       :object_changes_adapter,
       :serializer,
-      :version_limit
+      :version_limit,
+      :ignore_update_timestamp_only_changes
     )
 
     def initialize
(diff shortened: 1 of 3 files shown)


MohammadHabbab:
Hi @jaredbeck, Thank you very much for the quick reply. I appreciate your suggestion, however I think I'm trying to address a different issue.

Tracking `updated_at` attribute and including it in a generated version record when other attributes are updated on active record model is useful, however sometimes, a call like `model.touch` would only generate a version record with `updated_at` changing on its own (no other attributes in the model would get changed). 

In your suggestion, we would ignore `updated_at` property for all version records generated, whereas I was looking to introduce an option where, if we only change `updated_at` property on the model, we wouldn't generate a version record for that change of property alone.

Do you think this is a good enough case to differentiate with an option ? Or maybe the option name could be chosen better ?

jaredbeck:
> .. sometimes, a call like model.touch would only generate a version record with updated_at changing on its own (no other attributes in the model would get changed). ..

You can choose not to track `touch`es with:

```ruby
has_paper_trail on: %i[create update destroy] # omits touch
```

Does that help?

MohammadHabbab:
Oh, it definitely does. Thank you very much for taking the time to suggest the solution, I completely missed it in the docs 😅.

Issue #1115, opened 2018-07-11
Background jobs for storing versions


krainboltgreene:
So due to a mistake on my part I brought down my website because paper_trail versions couldn't save. Instead of requiring that versions get saved inside a request, I decided to make them run in the background!

https://gist.github.com/krainboltgreene/7b42135ef2c97977ebea6446f6bf523e

Thoughts?

jaredbeck:
Hi Kurtis. That's an interesting hack. I'm not sure what feedback you're looking for, though. Are you suggesting that PaperTrail should save versions using a background job? Not every application is set up for background jobs. Also, depending on write-frequency, using one job per version insert could lead to a lot of jobs and extra load on the database. Also, background-version-insertion is probably incompatible with [PT-AT](https://github.com/westonganger/paper_trail-association_tracking) because PT-AT expects to insert during the same transaction as the versioned record.

Or, are you suggesting PT should catch more version-creation errors? Some work has been done on that, see 0f90b909. More could be done in that vein, I think.

jaredbeck:
> Well largely I'm looking for a "Holy fuck don't do that you moron" and if not that then I'll gladly write an extension for people to use called paper_trail-background.

It's not clearly a terrible idea :) Worst case scenario, it'll be an interesting experiment.

I think it'd be cool if you used something like the `request_store` gem to keep an in-memory copy of all the version insertions that would have happened and then do those in a batch in the background.

>> Also, background-version-insertion is probably incompatible with PT-AT because PT-AT expects to insert during the same transaction as the versioned record.
> 
> Okay, this is interesting, why does it expect that?

PT-AT uses a `transaction_id` to correlate DML. https://github.com/westonganger/paper_trail-association_tracking/search?q=transaction_id&unscoped_q=transaction_id

jaredbeck:
Closing this, but if you work on this, let's continue to discuss here.

krainboltgreene:
So I found something strange, that I'm betting is accounted for in the sync style:

```
2018-09-24T03:23:11.187Z 92341 TID-ox8b5hwd5 VersionJob JID-cb51093fba1841be253f9df7 INFO: start
2018-09-24T03:23:11.188Z 92341 TID-ox8b5hv9h VersionJob JID-fa18a534857fe2ca569b9343 INFO: start
   (0.7ms)  BEGIN
  ↳ app/jobs/version_job.rb:5
   (0.1ms)  BEGIN
[...]

PR #1157, opened 2018-10-03
Adding async support to papertrail.

eternalfool:


jaredbeck:
In case you are still working on this, I think I'd like to see this done as a separate gem, if that is practical. First, I'd like to gauge interest before adding this feature. Second, I think there are a few different ways to implement async, and I'd like to see a few different implementations tried in practice before committing.

krainboltgreene:
See also #1115

PR #1158, opened 2018-10-10
Add the ability to pass options to the has_many :versions association macro

TylerRick:
Users can already customize some aspects of the `has_many :versions` association, including the association name (via `:versions` option) and the class name (via `:class_name` option).

This changeset allows other [options](https://api.rubyonrails.org/v5.2/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many-label-Options), such as `:extend`, to be passed on to this `has_many`.

### DIFF
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -138,10 +138,7 @@ module PaperTrail
       # @api public
       @model_class.send :attr_accessor, @model_class.version_association_name
 
-      # @api private - `version_class_name` - However, `rails_admin` has been
-      # using it since 2014 (see `rails_admin/extensions/paper_trail/auditing_adapter.rb`,
-      # https://github.com/sferik/rails_admin/commit/959e1bd4e47e0369d264b58bbbe972ff863767cd)
-      # In PR _____ () we ask them to use `paper_trail_options` instead.
+      # @api private - `version_class_name`
       @model_class.class_attribute :version_class_name
       @model_class.version_class_name = options[:class_name] || "PaperTrail::Version"
 
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -18,6 +18,15 @@ module PaperTrail
       `abstract_class`. This is fine, but all application models must be
       configured to use concrete (not abstract) version models.
     STR
+    DPR_PASSING_ASSOC_NAME_DIRECTLY_TO_VERSIONS_OPTION = <<~STR.squish
+      Passing versions association name as `has_paper_trail versions: %{versions_name}`
+      is deprecated. Use `has_paper_trail versions: {name: %{versions_name}}` instead.
+    STR
+    DPR_CLASS_NAME_OPTION = <<~STR.squish
+      Passing Version class name as `has_paper_trail class_name: %{class_name}`
+      is deprecated. Use `has_paper_trail versions: {class_name: %{class_name}}`
+      instead.
+    STR
 
     def initialize(model_class)
       @model_class = model_class
@@ -138,33 +147,19 @@ module PaperTrail
       # @api public
       @model_class.send :attr_accessor, @model_class.version_association_name
 
-      # @api private - `version_class_name`
-      @model_class.class_attribute :version_class_name
-      @model_class.version_class_name = options[:class_name] || "PaperTrail::Version"
-
-      assert_concrete_activerecord_class(@model_class.version_class_name)
-
       # @api public - paper_trail_event
       @model_class.send :attr_accessor, :paper_trail_event
 
-      setup_has_many_versions(options)
+      define_has_many_versions(options)
     end
 
-    def setup_has_many_versions(options)
-      unless options[:versions].is_a?(Hash)
-        options[:versions] = {
-          name: options[:versions]
-        }
-      end
-
-      # @api private - versions_association_name
-      @model_class.class_attribute :versions_association_name
-      @model_class.versions_association_name = options[:versions].delete(:name) || :versions
-
-      scope = options[:versions].delete(:scope) || -> { order(model.timestamp_sort_order) }
+    def define_has_many_versions(options)
+      options = ensure_versions_option_is_hash(options)
+      check_version_class_name(options)
+      check_versions_association_name(options)
+      scope = get_versions_scope(options)
       options[:versions].assert_valid_keys(valid_keys_for_has_many)
 
-      # @api public
       @model_class.has_many(
         @model_class.versions_association_name,
         scope,
@@ -174,6 +169,51 @@ module PaperTrail
       )
     end
 
+    def ensure_versions_option_is_hash(options)
+      unless options[:versions].is_a?(Hash)
+        if options[:versions]
+          ::ActiveSupport::Deprecation.warn(
+            format(
+              DPR_PASSING_ASSOC_NAME_DIRECTLY_TO_VERSIONS_OPTION,
+              versions_name: options[:versions].inspect
+            ),
+            caller(1)
+          )
+        end
+        options[:versions] = {
+          name: options[:versions]
+        }
+      end
+      options
+    end
+
+    def check_version_class_name(options)
+      # @api private - `version_class_name`
+      @model_class.class_attribute :version_class_name
+      if options[:class_name]
+        ::ActiveSupport::Deprecation.warn(
+          format(
+            DPR_CLASS_NAME_OPTION,
+            class_name: options[:class_name].inspect
+          ),
+          caller(1)
+        )
+        options[:versions][:class_name] = options.delete(:class_name)
+      end
+      @model_class.version_class_name = options[:versions][:class_name] || "PaperTrail::Version"
+      assert_concrete_activerecord_class(@model_class.version_class_name)
+    end
+
+    def check_versions_association_name(options)
+      # @api private - versions_association_name
+      @model_class.class_attribute :versions_association_name
+      @model_class.versions_association_name = options[:versions].delete(:name) || :versions
+    end
+
+    def get_versions_scope(options)
+      options[:versions].delete(:scope) || -> { order(model.timestamp_sort_order) }
+    end
+
     def setup_callbacks_from_options(options_on = [])
       options_on.each do |event|
         public_send("on_#{event}")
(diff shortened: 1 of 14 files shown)


spec/dummy_app/app/models/thing.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Thing < ActiveRecord::Base
-  has_paper_trail
+  has_paper_trail extend: VersionsExtensions
```
jaredbeck: Interesting. I've never used `extend` before.

TylerRick: Oh man, you're missing out. I use `extend` all the time to add custom methods to specific association collections. :)

The other `has_many` option — particularly `dependent` and `autosave`, could also be useful to have configurable for the `versions` association, of course — so I figured we may as well make *all* the options (the non-conflicting/not-already-possible options anyway) available to be passed in and passed on... but it was *mostly* for the `extend` that I wanted this...

TylerRick:
> So, instead of `has_paper_trail(x: 'y')`, I think we should nest them, like `has_paper_trail(has_many_versions_options: { x: 'y' }`.

My initial thinking was that these options should be top-level to match the existing `class_name` option, but I'm okay with nesting those options like you suggest, since they will be less commonly used. I hope we can find a less verbose name though. I'll try to think of a few options...

Good point about `has_many :version_associations`. I wasn't aware of that.(https://github.com/westonganger/paper_trail-association_tracking doesn't even mention the `has_many`, only the `version_associations` _table_.) Hmm.

Shouldn't one be able to assume that any options you pass to `has_paper_trail` are for the core PaperTrail association though? In particular, the `class_name` option isn't prefixed like that (maybe because it was added before you became maintainer) and always refers to the `PaperTrail::Version` model (or its replacement), _not_ `PaperTrail::VersionAssociation`. There is no `versions_association_class_name` option (yet), but they are free to add one. My point is that maybe it's okay to let unqualified/unprefixed options refer to the main `versions` association.

**`has_paper_trail versions_options: { option: 'option', ... }`**

Maybe we don't even need the word "_options" in the top-level option, since that's the only thing it could be...

**`has_paper_trail has_many: { option: 'option', ... }`**

Hmm. But we have other options related to the `has_many`. Might be more confusing to add new similar options that are synonyms (`versions` *is* the `has_many` this is referring to). Can we just combine them all?

**`has_paper_trail versions: { option: 'option', ... }`**

This is what I'm leaning towards currently (and is what I just pushed up an implementation for).

There is an existing `versions` option already for association name, of course, but we could continue treat it as a name if it's a non-Hash and as a hash of any options that you want to override the defaults if it's a Hash. 

This would continue to work:
```
  has_paper_trail versions: :paper_trail_versions
```

but this would also now be possible:
```
  has_paper_trail versions: {
                    name: :paper_trail_versions
                  }
```
and this:
```
  has_paper_trail versions: {
                    extend: MyExtensions
                  }
```
(see new tests in `spec/paper_trail/model_config_spec.rb`)

This has the benefit that most of the options for the `versions` association are neatly under a single key.

This is similar to how you can pass options to validations if you need to, or just a boolean if you don't:
```
validates :something, presence: true
validates :something, presence: {message: 'must be checked'}
```

__________

> Alternatively, we could allow people to disable the automatic definition of the association, allowing them to define the association themselves. Perhaps `has_paper_trail(has_many_versions: false)`.

That's not a bad idea, either! I'd be okay with having this as an additional option, though I think the ability to pass in an options hash gets us 90% of the way there and should be more user-friendly for most.

The only case I can think of where the option path wouldn't be configurable enough is if you wanted to pass a block to the `has_many` (but really, you could just throw it in a module and use `extend:`), like in their example:

```
has_many :employees do
  def find_or_create_by_name(name)
    ...
  end
end
```

Anyway, please see what I think of the new version I pushed up.

jaredbeck:
> Shouldn't one be able to assume that any options you pass to `has_paper_trail` are for the core PaperTrail association though?

The PT-AT gem currently uses `skip` and `save_changes`. They may want to add their own options in the future.

> `has_paper_trail versions: { option: 'option', ... }`
>
> There is an existing versions option already for association name, of course, but we could continue treat it as a name if it's a non-Hash and as a hash of any options that you want to override the defaults if it's a Hash.
>

Sounds good. Let's deprecate the old string form though, with a helpful message telling people to use the new form `has_paper_trail versions: { name: :bananas }` instead.

> This has the benefit that most of the options for the versions association are neatly under a single key.

Yeah, we want to be consistent. Which options would we be missing? `class_name`? I'd be OK with moving that under the `versions:` option and deprecating the top-level `class_name:` option.

>> Alternatively, we could allow people to disable the automatic definition of the association, allowing them to define the association themselves. Perhaps has_paper_trail(has_many_versions: false).
>
> That's not a bad idea, either! 

Cool. Let's tackle it separately though. There's already a lot going on in this PR.

TylerRick:
> Yeah, we want to be consistent. Which options would we be missing? `class_name`? I'd be OK with moving that under the `versions:` option and deprecating the top-level `class_name:` option.

Yeah, I believe that's the only other one besides name.

Sounds good. I'll follow through and push up those other changes next week.

> Cool. Let's tackle it separately though. There's already a lot going on in this PR.

Agreed.

TylerRick:
@jaredbeck Just realized we have a `:version` option too, for setting `version_association_name` (usually `:version`).

Would we want to change that to follow the same `version: {name: :version}` convention that we're proposing using for e `versions: {name: :versions}`?

Or should we rationalize that inconsistency with the fact that the `version` "association" isn't really a real association like `versions` is, so it probably won't ever need to have other options besides a name... or _might_ it conceivably have other options down the road?

jaredbeck:
> @jaredbeck Just realized we have a version: option too, for setting version_association_name (usually :version). Would we want to change that to follow the same version: {name: :version} convention that we're proposing ..

Good question. I've never understood the purpose of that feature.

```
version = # ...
record = version.reify
# we have local variable `version`, why do we need `record.version`?
```

So, I'd like to deprecate that feature, actually.

lib/paper_trail/model_config.rb
```diff
@@ -144,22 +144,35 @@ def setup_associations(options)
...
+
+      setup_has_many_versions(options)
+    end
+
+    def setup_has_many_versions(options)
+      unless options[:versions].is_a?(Hash)
+        options[:versions] = {
+          name: options[:versions]
+        }
+      end
```
jaredbeck: Should we add a deprecation warning here?

lib/paper_trail/model_config.rb
```diff
@@ -144,22 +144,35 @@ def setup_associations(options)
...
 
       # @api private - versions_association_name
       @model_class.class_attribute :versions_association_name
-      @model_class.versions_association_name = options[:versions] || :versions
+      @model_class.versions_association_name = options[:versions].delete(:name) || :versions
```
jaredbeck: Why `delete`?

lib/paper_trail/model_config.rb
```diff
@@ -144,22 +144,35 @@ def setup_associations(options)
...
+
+      scope = options[:versions].delete(:scope) || -> { order(model.timestamp_sort_order) }
+      options[:versions].assert_valid_keys(valid_keys_for_has_many)
```
jaredbeck: Neat! I did not know about `assert_valid_keys`.

What would happen if we didn't validate keys? Would `has_many` complain? I think I'd rather let `has_many` validate these, if it does.

spec/models/thing_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      expect(thing.versions.singleton_class).to be < VersionsExtensions
+      expect(thing.versions.inspect).to start_with("1 versions:")
+    end
+
+    it "applies the scope option" do
+      expect(Thing.reflect_on_association(:versions).scope).to be_a Proc
+      expect(thing.versions.to_sql).to end_with "ORDER BY id desc"
+    end
+  end
+end
```
jaredbeck: Good tests.

lib/paper_trail/model_config.rb
TylerRick: Yes, just added. :)

lib/paper_trail/model_config.rb
TylerRick: Works for me! Not sure why I didn't think of that. :)

And yep, `has_many` raises the exact same error (since it uses `assert_valid_keys` too—that's where I learned about it)... so I don't even have to change my test. :)

spec/models/thing_spec.rb
TylerRick: Thanks!

TylerRick:
Thought of another question: If we are going to deprecate the `top-level class_name: option`, do we only need to detect and show a warning if they try to *set* that configuration option, or also detect and show a warning if they try to *read* that configuration option? If we need it for *reading* too, that will be harder so I'm hoping not...

Ironically, that might have been easier to accomplish prior to changing this from a public to private API:
```
@model_class.class_attribute :version_class_name
```
because now I guess folks *have* to go through `paper_trail_options` hash as that is the public API.

(Reading nested hashes is not an extremely user friendly API, but we'll at least initialize `paper_trail_options[:versions]` so it's always a Hash and they don't have to worry about that...)

Looks like RailsAdmin is no longer using that private API, by the way, so we can remove the comment about that (I just did). Now they're [using](https://github.com/sferik/rails_admin/blob/master/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb) `paper_trail_options[:class_name]` instead to get that class name:
```
        def version_class_for(model)
          model.paper_trail_options[:class_name].try(:constantize) || @version_class
        end
```

My concern is that someone like RailsAdmin could upgrade, not see any deprecation warning, and continue to reference `paper_trail_options[:class_name]` forever... which would always be nil now, and not realize they need to change that to `paper_trail_options[:versions][:class_name]`

The code/project that *sets* that option and sees the deprecation warning and updates their code is going to be individual Rails _applications_ that use those engines/libraries,  not necessarily the _libraries_ themselves like RailsAdmin. They could catch it if they had a more comprehensive test suite but the RailsAdmin test suite does not specifically have a test for that option currently ([paper_trail_test.rb](https://github.com/sferik/rails_admin/blob/master/spec/dummy_app/app/active_record/paper_trail_test.rb)).

I guess we just rely on people to read the `Changelog` and notice this change (and if we're nice, we submit a PR to https://github.com/sferik/rails_admin/pulls :))... because I can't think of a nice way to detect that someone is relying on  `paper_trail_options[:class_name]` — without building our own Hash class and overriding `[]` ... but that seems like overkill to me, so I'm inclined to call this good enough. What do you think?

jaredbeck:
> I can't think of a nice way to detect that someone is relying on `paper_trail_options[:class_name]` — without building our own Hash class and overriding `[]`

If someone wants to do this in a separate PR, that's fine with me.

> if we're nice, we submit a PR to https://github.com/sferik/rails_admin/pulls :)

:+1: they should also use their gemspec to constrain PT major versions, if they aren't already

jaredbeck:
Thanks for your work on this, Tyler. I especially appreciate how thorough your tests are.

PR #1164, opened 2018-11-07
Add configurable "default_events" to allow tailoring which events are active on unqualified has_paper_trail declarations

gduprey:

Check the following boxes:

- [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,12 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Added
 
-- None
+- New configurable `default_events` defines which events are enabled by default.  
+  Example:
 
+  ```PaperTrail.config.default_events = %i[ create update destroy ]  # Skip touch events by default```  
+
+  If not set, the default is `%i[ create update destroy touch ]`.  As before, you can override the enabled events on a per-model basis with `has_paper_trail on: [ events.....]`
 ### Fixed
 
 - None
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -21,7 +21,8 @@ module PaperTrail
       :association_reify_error_behaviour,
       :object_changes_adapter,
       :serializer,
-      :version_limit
+      :version_limit,
+      :default_events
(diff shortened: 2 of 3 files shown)


gduprey:
This is related to my post on the issue that added touch events to paper_trail in that it can cause conflicts with other gems (like delay_touching).  It brought up the idea that it may be useful to allow the user of the gem to tailor what events, by default, are enabled (in our case, we'd have a one-line configurable that would disable touch events and still allow per-model overrides if needed).

jaredbeck:
Hi Gerry, thanks for the contribution.

```
# Skip touch events by default
PaperTrail.config.default_events = %i[ create update destroy ] 
```  

I'm generally opposed to adding configuration like this. This can be accomplished instead with:

```
# initializer
::MyPaperTrailDefaults = {
  on: %i[ create update destroy ] # Skip touch events
}.freeze

# banana.rb
class Banana < ApplicationRecord
  has_paper_trail(
    ::MyPaperTrailDefaults.merge(foo: 'bar')
  )
end
```

I'd be happy to review a PR that added instructions like this to the readme.

gduprey:
Except that you then have to update every single model to implement those defaults.  

In the past, this was no big deal, but adding the 'touch' event caused compatibility problems with the delay_touching gem.  They do not work well together.  We have over 115 models and the idea that a update to the gem breaks them all and having to edit every single one seems very difficult and error prone.  

Not sure why config options aren't preferred - seems like a pretty easy way to provide a good solid default behavior with the ability to tailor to a particular applications need without invasive having to add a on: clause (or similar) to every single model implementation.

No one needs to use it, but for those who need to customize which events are active by default (which is absolutely required if using something like delay_touching), this is much cleaner and much less error prone than having to update all existing models and remember to add additional options to every single use of has_paper_trail.

jaredbeck:
I'd be interested in a more general solution, like:

```
# initializer
PaperTrail.config.has_paper_trail_defaults = {
  on: %i[ create update destroy ]
}
```

These defaults would be overwritten (plain `merge`) by anything passed to `has_paper_trail`.

PR #1166, opened 2018-11-08
Add has_paper_trail_defaults configurable

gduprey:

Check the following boxes:

- [ ] Added tests.


lib/paper_trail/config.rb
```diff
@@ -31,6 +32,7 @@ def initialize
 
       # Variables which affect all threads, whose access is *not* synchronized.
       @serializer = PaperTrail::Serializers::YAML
+      @has_paper_trail_defaults = {}
```
jaredbeck: Should we use `HashWithIndifferentAccess`?

lib/paper_trail/config.rb
```diff
@@ -21,7 +21,8 @@ class Config
       :association_reify_error_behaviour,
       :object_changes_adapter,
       :serializer,
-      :version_limit
+      :version_limit,
+      :has_paper_trail_defaults
```
jaredbeck: Instead of an `attr_accessor`, should we synchronize using `@mutex` as we do with `#enabled=`?

lib/paper_trail/has_paper_trail.rb
```diff
@@ -57,7 +57,8 @@ module ClassMethods
       #
       # @api public
       def has_paper_trail(options = {})
-        paper_trail.setup(options)
+        # Merge defaults with per-model specifications
```
jaredbeck: Instead of having this comment here, please write a short paragraph about this in the method documentation above.

CHANGELOG.md
```diff
@@ -11,7 +11,13 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
...
 ### Added
 
-- None
+- Added new configurable `has_paper_trail_defaults` that accepts a hash of default options to be applied to each `has_paper_trail` directive.
+
+  Example:
+
+  ```PaperTrail.config.has_paper_trail_defaults = { on: %i[create update destroy] }```
+  
+  Defines a default for paper_trail events that would not include the touch event. It can be over-ridden on a per-model basis my including an explicit `on: [....]` option to `has_paper_trail`.
```
jaredbeck: I'd prefer to keep the changelog entries short. Please reduce this to a single sentence, and move the more verbose explanation to the readme, under section 1.e. Configuration.

lib/paper_trail/config.rb
gduprey: Sure - makes sense to me.

lib/paper_trail/config.rb
gduprey: Hmm - while I get it, that means mutex overhead for every model class load to reference it.  My gut says this is a config-time thing and an appropriate comment saying this isn't thread safe and intended for initialization time use would avoid adding synchronization overhead.  Given the likely usage pattern (e.g. initializer one-shot stuff), I would avoid it, but if you feel differently, it's not much work.

lib/paper_trail/has_paper_trail.rb
gduprey: Difference in style, but that's ok - will do

CHANGELOG.md
gduprey: Okay - will do.

TylerRick:
Nice. I was thinking it would be good to have something like this for setting default `ignore`/`only` columns. (Also wished for [here](https://stackoverflow.com/questions/41548357/is-it-possible-to-configure-papertrail-gem-to-ignore-attributes-globally)).

jaredbeck:
@TylerRick This upcoming feature is one of the reasons I wanted to avoid modifying the HPT options hash (https://github.com/paper-trail-gem/paper_trail/pull/1168/commits/1c06eed96ba4d30b577bc22522d72cb79174c927)

gduprey:
Changes for documentation (CHANGELOG, README and inside has_paper_trail.rb) added, added with_indifferent_access (and expanded that to the has_paper_trail setup hash as otherwise, it was sort of pointless)

jaredbeck:
Please rebase

gduprey:
Rebased ;-)

jaredbeck:
> Rebased ;-)

Hmmm, but this PR still says there are conflicts.

gduprey:
Okay - I give up - while I've written a metric butt ton of rails/ruby code, this is my first PR and this is going badly.  I seem to be stuck in an infinite loop of rebasing, merging, etc and it all looks good to me.  I've googled the heck out of resolving PR issues and everything seems OK (and my local and the github version both match and have both my changes and the other 10.1.0 changes), so I really don't know how to resolve this.  Sorry.

jaredbeck:
> Okay - I give up .. this is my first PR and this is going badly.

No worries. Rebasing, and resolving conflicts, is not easy, it takes a while to get good at it. I've rebased and squashed this, and made a few changes of my own. Just waiting for CI to run and I'll merge this.

seanlinsley:
this is a great incremental improvement to PaperTrail. thanks for the contribution!

gduprey:
I noticed that you removed the reference/example of default excluded fields in the config section of the changelog.   I added that as there were actually at least a few requests for something like that on SO.

Also, noticed that in next release, the class_name: option is newly deprecated - whats the alternative if you need to use a different table name due to DB table naming conflicts (and hence a custom model/class)?  Redmine, amongst other apps, uses a table called 'versions' for it's own purposes.

jaredbeck:
> I noticed that you removed the reference/example of default excluded fields in the config section of the changelog. I added that as there were actually at least a few requests for something like that on SO.

Yeah, it was too specific for section 1.e. Configuration, but maybe improvements could be made to section 2.c. Choosing Attributes To Monitor?

jgelo:
Great addition, very much appreciated!  I've had the has_paper_trail monkey patched to handle this for so long I had forgotten about it.

I'm running into an issue where a default meta option hash is being overwritten by any loaded class with a meta option.  I think some extra merging may be needed to ensure default meta is preserved, something like the following -

```ruby
def has_paper_trail(options = {})    
  defaults = PaperTrail.config.has_paper_trail_defaults.dup    
  meta     = ((defaults || {}).delete(:meta) || {}).merge(options.delete(:meta) || {})    
  paper_trail.setup(defaults.merge(options).merge(meta: meta))    
end 
```

What do you all think?

jaredbeck:
Could we use ActiveSupport's `Hash#deep_merge`?

jgelo:
Wow, I didn't realize deep_merge was a thing, brilliant!  Just did a quick test and it seems to work ok.

jaredbeck:
>> Could we use ActiveSupport's `Hash#deep_merge`?

> Just did a quick test and it seems to work ok.

Go for it. PRs welcome. New test required. Thanks!

jgelo:
> > > Could we use ActiveSupport's `Hash#deep_merge`?
> 
> > Just did a quick test and it seems to work ok.
> 
> Go for it. PRs welcome. New test required. Thanks!

This issue would affect the versions hash as well, so deep_merge looks like a good solution.  I'll try to work on a pull request this weekend.

TylerRick:
> Also, noticed that in next release, the class_name: option is newly deprecated - whats the alternative if you need to use a different table name due to DB table naming conflicts (and hence a custom model/class)? Redmine, amongst other apps, uses a table called 'versions' for it's own purposes.

Hey @gduprey, just noticed your question here and doesn't look like you got an answer.

The ability to customize the `versions` association like isn't deprecated, just the way that you pass it is changed.

So just change `has_paper_trail class_name: 'class_name'` to `has_paper_trail versions: {class_name: 'class_name'}` and you should be good! We actually made this change so that you have *more* control over the `versions` association, not less. :)

PR #1169, opened 2018-11-12
Update examples of customizing versions association in Readme

TylerRick:

Check the following boxes:

- [ ] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -954,35 +954,51 @@ see https://github.com/paper-trail-gem/paper_trail/issues/594
 
 ### 5.b. Configuring the `versions` Association
 
-You may configure the name of the `versions` association by passing
-a different name to `has_paper_trail`:
+You may configure the name of the `versions` association by passing a different
+name (default is `:versions`) in the `versions:` options hash:
 
 ```ruby
 class Post < ActiveRecord::Base
-  has_paper_trail class_name: 'Version', versions: :drafts
+  has_paper_trail versions: {
+    name: :drafts
+  }
 end
 
 Post.new.versions # => NoMethodError
 ```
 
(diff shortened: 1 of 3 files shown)


spec/dummy_app/app/models/thing.rb
```diff
@@ -2,8 +2,8 @@
 
 class Thing < ActiveRecord::Base
   has_paper_trail versions: {
+    scope: -> { order("id desc") },
```
TylerRick: (scope is conventionally passed first to has_many, before options hash)

README.md
```diff
@@ -954,35 +954,49 @@ see https://github.com/paper-trail-gem/paper_trail/issues/594
...
+  has_paper_trail versions: {
+    name: :drafts
+  }
 end
 
 Post.new.versions # => NoMethodError
 ```
 
-Overriding (instead of configuring) the `versions` method is not supported.
-Overriding associations is not recommended in general.
```
jaredbeck: Why remove this warning?

TylerRick: Added it back. I guess I'll just move it to the bottom of that section.

TylerRick: Still not sure how general advice about overriding superclass methods belongs in the Readme. As with any method override, doing so should be assumed to be at the end-user's own risk.

I did a quick test and was able to override the `versions` method and call `super`. Not sure why one would want to though...

In what way can that even be "supported" or not? :)

Why are we assuming people will be tempted to do that with this particular method (any more than any other method in the API) and why do we care if they try?

jaredbeck:
Merged, thanks!

README.md
TylerRick: Haha, fair enough... :)

PR #1159, opened 2018-10-20
Support rails 6.0.0-alpha

iggant:
We are trying to test our application on rails 6.0.0.alpha but we can't install this because of the version lock
Probably this will be helpful that we can try and report any issues with edge rails

PR #1171, opened 2018-11-30
Remove the upperbound constraint on ActiveRecord:

Edouard-chin:
- This PR aims to test papertrail on ActiveRecord edge in order for
  bundler to be able to resolve dependencies for application
  targetting rails/activerecord edge.
- During every ActiveRecord minor bump, this gemspec needs to be updated,
  I think we are being too conservative and having an upperbound limit
  doesn't help.

  If we wanted to be really conservative we would have to add a
  upperbound constraint on the patch version as well (< 5.x.x) since
  Rails doesn't follow semver and minor version can introduce breaking
  changes.

  What I think is better is to remove the upperbound constraint,
  and test paper-trail directly on ActiveRecord edge.

Check the following boxes:

- [ ] Added tests.


jaredbeck:
Hi Edouard, Thanks for the suggestion. 

```diff
-  s.add_dependency "activerecord", [">= 4.2", "< 5.3"]
+  s.add_dependency "activerecord", ">= 4.2"
```

I'm opposed to removing the upper bound. Rails makes breaking changes constantly. The chance of PaperTrail working perfectly with a new version of rails is very low. I don't want people to use PaperTrail with a version of rails it is incompatible with.

Edouard-chin:
Thanks for the quick reply @jaredbeck . I understand your concern but having a constraint on minor version like we currently do `< 5.3` can also introduce issues on people's app whenever they upgrade Rails to a new `5.2.x` version. Minor patch in AR also introduce breaking changes.

Would you be ok to loose the constraint to `< 6.0` to allow us boot our app on Rails 6.0.0.alpha at least?

jaredbeck:
> Would you be ok to [loosen] the constraint to `< 6.0` ..

```
Gem::Version.new('6.0.0.alpha') < Gem::Version.new('6.0') # => true
```

Nice. I didn't know pre-releases are considered lesser. So:

```
  s.add_dependency "activerecord", ">= 4.2", "< 6.0"
```

That's fine. There will be no rails 5.3, AFAIK.

I wish we could do:

```
  s.add_dependency "activerecord", "(4.2..5.2) or edge"
```

but there is no such syntax.

Appraisals
```diff
@@ -30,3 +30,8 @@ appraise "ar-5.2" do
   gem "activerecord", "~> 5.2.1"
   gem "rails-controller-testing", "~> 1.0.2"
 end
+
+appraise "ar-edge" do
+  gem "activerecord", github: "rails/rails"
+  gem "rails-controller-testing", "~> 1.0.3"
+end
```
jaredbeck: Please add this new appraisal to `.travis.yml`, in the `gemfile:` array

jaredbeck:
This seems like a duplicate of https://github.com/paper-trail-gem/paper_trail/pull/1159. Should we keep both open?

PR #1172, opened 2018-12-03
Support rails 6.0.0

Edouard-chin:
Supersede #1159

Check the following boxes:

- [ ] Added tests.


CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 - [#1158](https://github.com/paper-trail-gem/paper_trail/pull/1158) — Add the
   ability to pass options, such as `scope` or `extend:` to the `has_many
   :versions` association macro.
+- [#1172](https://github.com/paper-trail-gem/paper_trail/pull/1172) -
+  Support rails 6.0.0
```
jaredbeck: Should we say "6.0.0.alpha"?

paper_trail.gemspec
```diff
@@ -28,7 +28,7 @@ has been destroyed.
   s.required_ruby_version = ">= 2.3.0"
 
   # Rails does not follow semver, makes breaking changes in minor versions.
-  s.add_dependency "activerecord", [">= 4.2", "< 5.3"]
+  s.add_dependency "activerecord", [">= 4.2", "< 6.1"]
```
jaredbeck: I'd be more comfortable with `< 6.0` here, as there will probably be more breaking changes before 6.0 is final.

spec/dummy_app/app/models/person.rb
```diff
@@ -54,5 +44,20 @@ def load(value)
...
+      if value.is_a? ActiveSupport::TimeZone
+        super
+      else
+        zone = ::Time.find_zone(value) # nil if can't find time zone
+        super zone
+      end
+    end
+
+    serialize :time_zone, TimeZoneSerializer.new
+  end
```
jaredbeck: Does rails 6 change how `serialize` works? If so, can you point me to some reading on the subject?

lib/paper_trail/type_serializers/postgres_array_serializer.rb
```diff
@@ -29,8 +29,7 @@ def deserialize(array)
       private
 
       def active_record_pre_502?
-        ::ActiveRecord::VERSION::MAJOR < 5 ||
-          (::ActiveRecord::VERSION::MINOR.zero? && ::ActiveRecord::VERSION::TINY < 2)
+        ::ActiveRecord.version < Gem::Version.new("5.0.2")
```
jaredbeck: Is `::ActiveRecord.gem_version` preferred here?

paper_trail.gemspec
iggant: with constrain 6.0 we can't install even alpha rails version

Edouard-chin: @jaredbeck is right, `< 6.0` allow us to support alpha, beta and release candidates without supporting final release.

```ruby
Gem::Version.new('6.0') > Gem::Version.new('6.0.0.alpha')
# true
```

lib/paper_trail/type_serializers/postgres_array_serializer.rb
Edouard-chin: Yes you are right, I confused both `gem_version` wraps the version string into a `Gem::Version` object for free

spec/dummy_app/app/models/person.rb
Edouard-chin: I know the ActiveModel Attribute API was released in 5.1 but I don't know why @iggant made that change. I'll have to dig, or maybe @iggant you can explain

Edouard-chin: Dug into this and the TimeZone Attribute is technically not a requirement to make tests pass on AR 6.0.
Without it, some tests were failing but the reason for the failures weren't directly related do that. I explained the reason in detail in https://github.com/paper-trail-gem/paper_trail/pull/1172/commits/ac12912dcb9f2935938bb48af43979f2a8442b28

That being said, even though adding a new TimeZone Attribute isn't a requirement, it makes the code cleaner. It's not a new 6.0 thing, we could have used it since Rails 5.0 when the Attribute API got released https://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html

Let me know if you prefer to stick with our old serializer.

Edouard-chin:
CI is green, each commit should explain the changes. I will squash everything once you give your ✅ .

Thanks!

spec/dummy_app/app/models/person.rb
jaredbeck: > Dug into this and the TimeZone Attribute is technically not a requirement to make tests pass on AR 6.0.

Thanks for investigating. For now, I have removed this. Let's discuss it in a separate PR.

jaredbeck:
Thanks Anton and Edouard for your work on this. @iggant @Edouard-chin. 

I think I'll release this as PT 10.1.0

Edouard-chin:
Thanks a lot for your quick review and help on this, much appreciated ❤️

PR #1176, opened 2019-01-02
update readme for disabling papertrail within initializer

### LINKED ISSUE #1038: Can't set enabled to false in initializer (rails)
I tried to have logging turned on only for specific endpoints in all environments. If you set enabled to false in the initializer it will not work.

In `config/initializers/paper_trail.rb`
```PaperTrail.config.enabled = false```

Then `rails console`
```
[1] pry(main)> PaperTrail.enabled?
=> true
```

After some digging, I found that engine overrides it to true when environment config is missing.

In `PaperTrail::Rails::Engine`
```
initializer "paper_trail.initialisation" do |app|
  PaperTrail.enabled = app.config.paper_trail.fetch(:enabled, true)
end
```

It probably should use `PaperTrail.enabled?` instead of `true`, or even not set `enabled` when config value is missing.

### PR DESCRIPTION
**Pull request by oehlschl**
Fixes #1038 

Updating the readme to clarify that `PaperTrail.config.enabled` cannot be set within an initializer since it's eventually overwritten by the environment-specific config (or by the internal default of `true`).

I tried to be clear without giving unnecessary details. Also tried to use style consistent with the rest of the readme, but please let me know if you have suggestions.

- [ ] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,6 @@ Syntax example: (options described in detail later)
 
 ```ruby
 # config/initializers/paper_trail.rb
-PaperTrail.config.enabled = true
 PaperTrail.config.has_paper_trail_defaults = {
   on: %i[create update destroy]
 }
@@ -292,9 +291,14 @@ PaperTrail.config.version_limit = 3
 ````
 
 These options are intended to be set only once, during app initialization (eg.
-in `config/initializers`). It is unsafe to change them while the app is running.
+in `config/initializers` or in an environment-specific configuration file such as
+`config/environments/test.rb`). It is unsafe to change them while the app is running.
 In contrast, `PaperTrail.request` has various options that only apply to a
 single HTTP request and thus are safe to use while the app is running.
+
+Note that `PaperTrails.config.enabled` will be overwritten if it is set within
+`config/initializers`. For more information, see
+[Turning PaperTrail Off](#2d-turning-papertrail-off) below.
 
 ## 2. Limiting What is Versioned, and When
 
@@ -502,6 +506,11 @@ PaperTrail.enabled = false
 
 **Do not use this in production** unless you have a good understanding of
 threads vs. processes.
+
+Also note that PaperTrail can only be disabled globally during or after
+environment-specific initialization. In other words, changes to
+`PaperTrail.enabled` (or its alias `PaperTrail.config.enabled`) in
+`config/initializers` will be overwritten.
 
 A legitimate use case is to speed up tests. See [Testing](#7-testing) below.
 

### CODE REVIEW

**Comment by oehlschl** [comment]
Also note that I did mention both `PaperTrail.enabled` and `PaperTrail.config.enabled` for clarity, since that they seemed to be used interchangeably (and without preference) in the readme.

**Comment by jaredbeck** [comment]
Thanks for taking a stab at this. I've worked hard on the documentation for PT's various on/off switches, and I'm always open to suggestions on how to improve the docs.

> .. `PaperTrail.enabled` (or its alias `PaperTrail.config.enabled`) ..

To say that they are aliases is effectively correct, and a useful addition to the docs. They are not technically aliases, but the former does delegate to the later. We can fine-tune the wording and placement.

> .. PaperTrail.config.enabled cannot be set within an initializer since it's eventually overwritten by the environment-specific config ..

Hmmm. According to my testing, environment files (`config/environments`) happen *before* initializers (`config/initializers`). See also [The Rails Initialization Process][1] guide.

[1]: https://guides.rubyonrails.org/initialization.html

**Comment by oehlschl** [comment]
Thanks for the response, @jaredbeck. It could just be poor wording on my part; it's actually that any custom initialization of `enabled` gets overwritten by PaperTrail's own internal initialization hook as described here: https://github.com/paper-trail-gem/paper_trail/issues/1038

You're correct on the order, but from my experience (and from the other report), what happens is:
1) environment-specific initialization
2) config/initializers/paper_trail.rb
3) /paper_trail/lib/paper_trail/frameworks/rails/engine.rb

Due to the `PaperTrail.enabled = app.config.paper_trail.fetch(:enabled, true)` line, any `.enabled` value in a custom initializer from (2) will be overwritten by the environment-specific value defined in (1) or by the `true` default defined in (3).

Obviously that's a lot of detail to put in the README, but the effect is still the same as what I described; custom initialization gets overwritten by the environment config (even though the environment config is _evaluated_ first).

**Comment by jaredbeck** [comment]
Thanks for the clarification. 

> 1. environment-specific initialization
> 2. config/initializers/paper_trail.rb
> 3. [the initializer defined by] /paper_trail/lib/paper_trail/frameworks/rails/engine.rb

I'm seeing the same order. 

Maybe we should we deprecate `config.paper_trail.enabled`. I don't see what value it provides, and having two ways to do the same thing is confusing. Let's compare the two:

| `PaperTrail.enabled=` | `config.paper_trail.enabled=` |
| -------------------- | ---------------------------- |
| happens immediately  | happens later, when the initializer defined in `engine.rb` runs. The order in which this happens, relative to other initializers is not known. In our testing it is after custom initializers, but is this guaranteed? |
| about 5 LoC          | hundreds of LoC, including arcane "rails engine" stuff that few people on Earth really understand, myself excluded |
| can be used in custom initializer or standard environment file | can only be used in standard environment file? |
| can be used after startup | cannot be used after startup |
| works in all frameworks | only works in rails |

I think we should keep `PaperTrail::Rails::Engine` (it allows us to register our `paths["app/models"]` so that's nice) but maybe we should deprecate `config.paper_trail.enabled=`. Thoughts?

**Comment by oehlschl** [comment]
I hadn't thought too much about it before now, but it seems like there are some good arguments for deprecating `config.paper_trail`.

For counter arguments, I could imagine something related to Rails magic and  specifics with config objects and threads under the hood, but that's just thinking hypothetically; I can't actually think of anything concrete. That said, it also looks from the source like PaperTrail already accounts for threads well enough on its own.

For other pro arguments, the `config.paper_trail` option would also keep all the library configuration on the module itself and collapse two ways of doing something into one source of truth.

If you do want to go forward with this, I'd suggest first raising/logging a deprecation messaging in the engine if `config.paper_trail.enabled` is set.

**Comment by jaredbeck** [comment]
@oehlschl please review https://github.com/paper-trail-gem/paper_trail/pull/1183, thanks!

**Comment by jaredbeck** [comment]
Closed by https://github.com/paper-trail-gem/paper_trail/pull/1183

**Comment by HoneyryderChuck** [comment]
Hi. I'm on 10.3.2, and I'm going to keep this option, mostly because using the recommend `PaperTrail.enabled = ` does not work, i.e. my tests are storing versions.

**Comment by iberianpig** [comment]
@oehlschl @jaredbeck 

This issue has been unresolved yet.

At first, "PaperTrail.enabled=" is loaded in the following order.

1. config/environments/*.rb
1. config/initializers/*.rb
1. lib/paper_trail/frameworks/rails/engine.rb:35 (in PaperTrail Gem)
1. config.after_initialize { } in config/environments/*.rb 

Even if you set `PaperTrail.enabled=false` to `config/environments/development.rb`,

`lib/paper_trail/frameworks/rails/engine.rb` will override `PaperTrail.enabled=true`.

```lib/paper_trail/frameworks/rails/engine.rb
      initializer "paper_trail.initialisation" do |app|
        enable = app.config.paper_trail[:enabled] # => nil
        if enable.nil?
          unless PaperTrail.enabled? # =>false
            ::ActiveSupport::Deprecation.warn(DPR_RUDELY_ENABLING)
            PaperTrail.enabled = true # !!!override here!!!
          end
        else
          ::ActiveSupport::Deprecation.warn(DPR_CONFIG_ENABLED)
          PaperTrail.enabled = enable
        end
      end
```

Therefore, it will never become `PaperTrail.enabled=false`.

Is this behavior correct? 

If this behavior is correct, the documentation should explain where should we set PaperTrail.config.enabled.

PR #1177, opened 2019-01-02
fix: do not store ignored and skipped attributes in `object_changes` on destroy

michaelkhabarov:
I think skipped and ignored attributes should be ignored in `object_changes` on `destroy` (same way as on `create` and `update`). This PR fixes that.

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -106,16 +106,9 @@ module PaperTrail
         (changed_in_latest_version - ignore) - skip
       end
 
-      # Rails 5.1 changed the API of `ActiveRecord::Dirty`. See
-      # https://github.com/paper-trail-gem/paper_trail/pull/899
-      #
       # @api private
       def changed_in_latest_version
-        if @in_after_callback && RAILS_GTE_5_1
-          @record.saved_changes.keys
-        else
-          @record.changed
-        end
+        changes_in_latest_version.keys
       end
 
       # @api private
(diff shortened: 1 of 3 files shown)


lib/paper_trail/events/base.rb
```diff
@@ -106,16 +106,9 @@ def changed_and_not_ignored
...
-      # https://github.com/paper-trail-gem/paper_trail/pull/899
-      #
       # @api private
       def changed_in_latest_version
-        if @in_after_callback && RAILS_GTE_5_1
-          @record.saved_changes.keys
-        else
-          @record.changed
-        end
+        changes_in_latest_version.keys
```
jaredbeck: This avoids duplicating the conditional? Nice!

lib/paper_trail/events/destroy.rb
```diff
@@ -22,13 +22,18 @@ def data
...
-          data[:object_changes] = prepare_object_changes(changes)
+          data[:object_changes] = prepare_object_changes(notable_changes)
         end
         merge_item_subtype_into(data)
         merge_metadata_into(data)
       end
+
+      private
+
+      # Rails' implementation returns nothing on destroy :/
```
jaredbeck: Nice use of polymorphism!

When overriding methods in the future, please add a `# @override` comment, like this:

```
      # Rails' implementation returns nothing on destroy :/
      #
      # @override
```

PR #1183, opened 2019-01-27
Deprecate config.paper_trail.enabled

jaredbeck:
- [ ] Added tests.

### DIFF
--- a/lib/paper_trail/frameworks/rails/engine.rb
+++ b/lib/paper_trail/frameworks/rails/engine.rb
@@ -4,11 +4,41 @@ module PaperTrail
   module Rails
     # See http://guides.rubyonrails.org/engines.html
     class Engine < ::Rails::Engine
+      DPR_CONFIG_ENABLED = <<~EOS.squish.freeze
+        The rails configuration option config.paper_trail.enabled is deprecated.
+        Please use PaperTrail.enabled= instead. We are *considering* removing
+        this config option because people were getting confused that PT has
+        both, specifically regarding *when* each was happening. See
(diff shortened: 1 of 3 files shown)


lib/paper_trail/frameworks/rails/engine.rb
```diff
@@ -4,11 +4,41 @@ module PaperTrail
   module Rails
     # See http://guides.rubyonrails.org/engines.html
     class Engine < ::Rails::Engine
+      DPR_CONFIG_ENABLED = <<~EOS.squish.freeze
```
oehlschl: FWIW, you could make these constants private if desired.

lib/paper_trail/frameworks/rails/engine.rb
```diff
@@ -4,11 +4,41 @@ module PaperTrail
...
       config.paper_trail = ActiveSupport::OrderedOptions.new
       initializer "paper_trail.initialisation" do |app|
-        PaperTrail.enabled = app.config.paper_trail.fetch(:enabled, true)
+        enable = app.config.paper_trail[:enabled]
+        if enable.nil?
+          unless PaperTrail.enabled?
+            ::ActiveSupport::Deprecation.warn(DPR_RUDELY_ENABLING)
+            PaperTrail.enabled = true
+          end
+        else
```
oehlschl: Just to sanity check, it looks like this covers all the cases:

- app.config nil, PaperTrail enabled = no-op
- app.config nil, PaperTrail disabled = "rudely enabled" warning
- app.config false, * = deprecation warning
- app.config true, * = deprecation warning

👍

PR #1184, opened 2019-01-29
No need to calculate previous values of skipped attributes

khiav223577:
I'm going to remove a column of one model. I want to make sure that no one calls it anymore before removing it, so I overwritted it as below:
```rb
def my_column
  fail 'should not be called'
end
```

But `paper_trail` will call it even the attribute is skipped.
It calculates all attributes and then filters out skipped attributes.
https://github.com/paper-trail-gem/paper_trail/blob/85884ba06c1eefbc76114d70c4f8f05791f5659a/lib/paper_trail/events/base.rb#L219-L221

The flow can be improved to calculate only non-skipped attributes.

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -59,8 +59,10 @@ module PaperTrail
       end
 
       # @api private
-      def attributes_before_change(is_touch)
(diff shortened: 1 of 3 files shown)


jaredbeck:
Thanks, Rumble.

PR #1188, opened 2019-03-10
Lower the memory allocations

stokarenko:
It prevents the repeating hashes creation during the calculation of `object` and `object_changes`.

The real-life case, discovered with a help of `memory_profiler` gem:
1. It saves `~35Mb` per bulk of 100 typical objects like `User`;
2. It saves `~875Mb` per Sidekiq process full of bulk jobs like P1, under default concurrency 25.

As a part of RAM optimization we do:
*  memoization of repeatedly called `PaperTrail::Events::Base` methods;
* cache the AR `changed_attributes` for AR < 5.1;
* prefer to build the Version objects by pure `Class.new` instead of `association.build`.

The record's association cache will be invalidated on the way by `association.reset`, so no any explicit `versions.reload` calls are needed now.

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -60,15 +60,28 @@ module PaperTrail
 
       # @api private
       def nonskipped_attributes_before_change(is_touch)
-        record_attributes = @record.attributes.except(*@record.paper_trail_options[:skip])
+        cache_changed_attributes do
+          record_attributes = @record.attributes.except(*@record.paper_trail_options[:skip])
 
-        Hash[record_attributes.map do |k, v|
-          if @record.class.column_names.include?(k)
-            [k, attribute_in_previous_version(k, is_touch)]
-          else
-            [k, v]
+          record_attributes.each_key do |k|
+            if @record.class.column_names.include?(k)
+              record_attributes[k] = attribute_in_previous_version(k, is_touch)
+            end
           end
-        end]
+        end
+      end
+
+      # Rails 5.1 changed the API of `ActiveRecord::Dirty`.
+      # @api private
+      def cache_changed_attributes
+        if RAILS_GTE_5_1
+          # Everything works fine as it is
+          yield
+        else
+          # Any particular call to `changed_attributes` produces the huge memory allocation.
+          # Lets use the generic AR workaround for that.
+          @record.send(:cache_changed_attributes) { yield }
+        end
       end
 
       # Rails 5.1 changed the API of `ActiveRecord::Dirty`. See
@@ -110,7 +123,8 @@ module PaperTrail
 
       # @api private
       def changed_in_latest_version
-        changes_in_latest_version.keys
+        # Memoized to reduce memory usage
+        @changed_in_latest_version ||= changes_in_latest_version.keys
       end
 
       # Rails 5.1 changed the API of `ActiveRecord::Dirty`. See
@@ -118,10 +132,13 @@ module PaperTrail
       #
       # @api private
       def changes_in_latest_version
-        if @in_after_callback && RAILS_GTE_5_1
-          @record.saved_changes
-        else
-          @record.changes
+        # Memoized to reduce memory usage
+        @changes_in_latest_version ||= begin
+          if @in_after_callback && RAILS_GTE_5_1
+            @record.saved_changes
+          else
+            @record.changes
+          end
         end
       end
 
@@ -202,16 +219,19 @@ module PaperTrail
 
       # @api private
       def notably_changed
-        only = @record.paper_trail_options[:only].dup
-        # Remove Hash arguments and then evaluate whether the attributes (the
-        # keys of the hash) should also get pushed into the collection.
-        only.delete_if do |obj|
-          obj.is_a?(Hash) &&
-            obj.each { |attr, condition|
-              only << attr if condition.respond_to?(:call) && condition.call(@record)
-            }
+        # Memoized to reduce memory usage
+        @notably_changed ||= begin
+          only = @record.paper_trail_options[:only].dup
+          # Remove Hash arguments and then evaluate whether the attributes (the
+          # keys of the hash) should also get pushed into the collection.
+          only.delete_if do |obj|
+            obj.is_a?(Hash) &&
+              obj.each { |attr, condition|
+                only << attr if condition.respond_to?(:call) && condition.call(@record)
+              }
+          end
+          only.empty? ? changed_and_not_ignored : (changed_and_not_ignored & only)
         end
-        only.empty? ? changed_and_not_ignored : (changed_and_not_ignored & only)
       end
 
       # Returns hash of attributes (with appropriate attributes serialized),
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -67,14 +67,26 @@ module PaperTrail
 
     def record_create
       return unless enabled?
-      event = Events::Create.new(@record, true)
+
+      build_version_on_create(in_after_callback: true).tap do |version|
+        version.save!
+        # One more Version object has been created. Need to invalidate
+        # the `versions` association cache, otherwise the new version
+        # can stay invisible within that association
+        versions.reset
+      end
+    end
+
+    # @api private
+    def build_version_on_create(in_after_callback:)
+      event = Events::Create.new(@record, in_after_callback)
 
       # Merge data from `Event` with data from PT-AT. We no longer use
       # `data_for_create` but PT-AT still does.
-      data = event.data.merge(data_for_create)
+      data = event.data.merge!(data_for_create)
 
-      versions_assoc = @record.send(@record.class.versions_association_name)
-      versions_assoc.create!(data)
+      # Pure `version_class.new` reduces memory usage compared to `versions_assoc.build`
+      @record.class.paper_trail.version_class.new(data)
     end
 
     # PT-AT extends this method to add its transaction id.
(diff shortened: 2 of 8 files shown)


spec/paper_trail/cleaner_spec.rb
```diff
@@ -58,7 +58,7 @@ module PaperTrail
           expect(animal.paper_trail.versions_between(date, (date + 1.day)).size).to(eq(3))
           PaperTrail.clean_versions!(date: date)
           expect(PaperTrail::Version.count).to(eq(8))
-          expect(animal.versions.reload.size).to(eq(2))
+          expect(animal.versions.size).to(eq(2))
```
jaredbeck: Why remove this `reload`?

I'd expect a memory optimization to *not* change existing tests.

jaredbeck:
> It saves ~35Mb per bulk of 100 typical objects like User;

Awesome! Thanks for working on this.

jaredbeck:
> I'd expect a memory optimization to not change existing tests.

Said another way, I'd be a lot more comfortable reviewing this if existing tests can be kept exactly the same.

spec/paper_trail/cleaner_spec.rb
stokarenko: @jaredbeck The existing tests are able to pass without any changes. I've removed that because `reload` call is rudimental now:

1. We going to change the way how to build the Version objects, in order to optimize the memory usage. Now we use pure `Class.new` instead of `association.build`;
2. That will fail some tests since `Class.new` has no side effects around association cache;
3. in order to fix that we invalidate the association cache implicitly by `association.reset` right after any Version object saving;
4. at this point the `versions.reload` becomes to be rudimental. I.e. `versions` association works as expected even after the PT callbacks without any additional efforts.

I can move `reloads` back if you like. In additional, I can add some tests which will prove that `reload` calls are no more necessary :)

stokarenko:
Reverted the changes on tests, @jaredbeck please revisit.

lib/paper_trail/record_trail.rb
```diff
@@ -67,14 +67,25 @@ def previous_version
 
     def record_create
       return unless enabled?
-      event = Events::Create.new(@record, true)
+
+      build_version_on_create(in_after_callback: true).tap do |version|
+        version.save!
+        versions.reset
```
jaredbeck: Please add a comment explaining this `reset`.

lib/paper_trail/record_trail.rb
```diff
@@ -67,14 +67,25 @@ def previous_version
...
+      event = Events::Create.new(@record, in_after_callback)
 
       # Merge data from `Event` with data from PT-AT. We no longer use
       # `data_for_create` but PT-AT still does.
-      data = event.data.merge(data_for_create)
-
-      versions_assoc = @record.send(@record.class.versions_association_name)
-      versions_assoc.create!(data)
+      data = event.data.merge!(data_for_create)
+      @record.class.paper_trail.version_class.new(data)
```
jaredbeck: Please add a comment explaining that `version_class.new` reduces memory usage compared to `versions_assoc.create!`.

lib/paper_trail/record_trail.rb
```diff
@@ -117,24 +128,36 @@ def data_for_destroy
     # @api private
     # @return - The created version object, so that plugins can use it, e.g.
     # paper_trail-association_tracking
-    def record_update(force:, in_after_callback:, is_touch:)
+    def record_update(build_version_params)
```
jaredbeck: I'd prefer not to change the signature of this method, if that is practical.

lib/paper_trail/record_trail.rb
```diff
@@ -67,14 +67,25 @@ def previous_version
...
+
+      build_version_on_create(in_after_callback: true).tap do |version|
+        version.save!
+        versions.reset
+      end
+    end
+
+    # Returns the ready-for-save Version object built for a new record.
+    # Can be useful for manual Version saving, for instance as a part of
+    # bulk processing (https://github.com/zdennis/activerecord-import).
```
jaredbeck: I don't want to make any changes to our public API as part of an optimization PR. Please either:

1. inline this method, or
2. delete this comment and mark as `@api private`

Whichever you prefer.

I'd be happy to consider changes to our public API as a separate PR.

lib/paper_trail/record_trail.rb
```diff
@@ -117,24 +128,36 @@ def data_for_destroy
...
+      version = build_version_on_update(build_version_params)
+      return unless version
 
-      versions_assoc = @record.send(@record.class.versions_association_name)
-      version = versions_assoc.create(data)
+      version.save
       if version.errors.any?
         log_version_errors(version, :update)
       else
+        versions.reset
```
jaredbeck: Please add a comment explaining this `reset`

spec/paper_trail/association_reify_error_behaviour/error_spec.rb
```diff
@@ -29,7 +29,7 @@
       expect {
         person.reload.versions.second.reify(has_one: true)
       }.to(
-        raise_error(::PaperTrail::Reifiers::HasOne::FoundMoreThanOne) do |err|
+        raise_error(::PaperTrailAssociationTracking::Reifiers::HasOne::FoundMoreThanOne) do |err|
```
jaredbeck: If you feel like it, please move this to a separate PR so it can be merged immediately.

spec/paper_trail/model_spec.rb
```diff
@@ -868,4 +869,74 @@
...
+      end
+
+      it "is frugal enough" do
+        # Some time ago there was 144kbs..
+        # At the time of commit the test passes with assertion on 27kbs.
+        # Lets assert 35kbs then, to avoid flaky fails.
+        expect(&version_building).to allocate_less_than(35).kilobytes
+      end
+    end
+  end
```
jaredbeck: It's very cool that you were able to write these benchmarks as tests. I've never worked on a project with such tests, but I'm willing to try it. My only concern would be if, in the future, memory usage inside rails increases (in a way we cannot control) causing these tests to fail. If it becomes painful in the future, we may need to separate these benchmarks from the test suite.

spec/support/performance_helpers.rb
```diff
@@ -0,0 +1,38 @@
...
+    @allocated = benchmark.total_allocated_memsize
+    @allocated /= 1024 if @scale == :kbs
+    @allocated <= expected
+  end
+
+  failure_message do
+    "expected that example will allocate less than #{expected}#{@scale},"\
+    " but allocated #{@allocated}#{@scale}"
+  end
+end
```
jaredbeck: I did not know the RSpec matcher API was so flexible. This is pretty great.

lib/paper_trail/events/base.rb
```diff
@@ -60,15 +60,28 @@ def attribute_changed_in_latest_version?(attr_name)
...
+      # @api private
+      def cache_changed_attributes
+        if RAILS_GTE_5_1
+          # Everything works fine as it is
+          yield
+        else
+          # Any particular call to `changed_attributes` produces the huge memory allocation.
+          # Lets use the generic AR workaround for that.
+          @record.send(:cache_changed_attributes) { yield }
+        end
```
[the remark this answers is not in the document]
stokarenko: I just seen the strong PT support even for `AR v4.2`, fell that it important to keep it..

As for that particular `if-else` switch around `RAILS_GTE_5_1` - it looks not much tricky than all other alike switches within `Events::Base` class. Even more, if we will remove this switch from `Events::Base`, we will be forced to add it to performance tests, since the memory allocation will be noticeably larger then.

Probably the collapsing of all the `RAILS_GTE_5_1` switches can be a target for `PT v11`.

PS. And, honestly, I need that switch personally, please don't ask me why :)

lib/paper_trail/events/base.rb
```diff
@@ -285,7 +302,7 @@ def serialize_object_changes(changes)
         AttributeSerializers::ObjectChangesAttribute.
           new(@record.class).
           serialize(changes)
-        changes.to_hash
+        changes
```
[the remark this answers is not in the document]
stokarenko: Here we go - https://github.com/paper-trail-gem/paper_trail/pull/1189

spec/paper_trail/association_reify_error_behaviour/error_spec.rb
stokarenko: Here we go - https://github.com/paper-trail-gem/paper_trail/pull/1190

spec/paper_trail/model_spec.rb
stokarenko: Any optimization efforts become to be mindless without such kind of tests :)
Any next PR can easily ruin all the winnings.

As for the third-party changes, which are really out of our control - at least we will know about the problems with them. We can mark such particular gems and versions as `ineffective` both in readme and tests for example.

In general you're right, the life will be harder with these tests :)

stokarenko:
@jaredbeck please take a look at https://travis-ci.org/paper-trail-gem/paper_trail/builds/505655659

As you can see, the removal of https://github.com/paper-trail-gem/paper_trail/pull/1189 is strongly affect to performance. I'm forced to make more pessimistic assertions for this PR.

stokarenko:
@jaredbeck it is ready for the next review )

lib/paper_trail/events/base.rb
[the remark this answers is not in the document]
jaredbeck: > As for that particular if-else switch around RAILS_GTE_5_1 - it looks not much tricky than all other alike switches ..

Yeah, it's not too painful. We can add it, but ..

> I just seen the strong PT support even for AR v4.2, fell that it important to keep it..

PT follows the rails EoL policy. As soon as rails 6 is released, PT will drop support for rails < 5.2.

> Probably the collapsing of all the RAILS_GTE_5_1 switches can be a target for PT v11.

Yes. When we drop support for old rails versions, SemVer does not require us to make a major version bump, but yes, we often do anyway.

lib/paper_trail/record_trail.rb
jaredbeck: So, this `reset` is necessary because we are using `version_class.new` instead of `versions_assoc.build`? If so, can you mention this specifically in the comment please?

lib/paper_trail/events/base.rb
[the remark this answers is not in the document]
stokarenko: > Yeah, it's not too painful. We can add it, but ..
> PT follows the rails EoL policy. As soon as rails 6 is released, PT will drop support for rails < 5.2.

Thanks, I'm ok with that )

lib/paper_trail/record_trail.rb
stokarenko: Already done, looks like that:

```ruby
      build_version_on_create(in_after_callback: true).tap do |version|
        version.save!
        # One more Version object has been created. Need to invalidate
        # the `versions` association cache, otherwise the new version
        # can stay invisible within that association
        versions.reset
      end
```

stokarenko:
> please rebase and restore the memory numbers in the test.

done, please revisit )

Thank you for assistance during this journey )

jaredbeck:
> I don't want to make any changes to our public API as part of an optimization PR.

I'm going to make `build_version_on_create` and `build_version_on_update` private, if you can live with that. I want it to be clear that they are not part of our public API (yet).

If you'd like to discuss some ideas for the public API, please email me (jared@jaredbeck.com) as we have temporarily disabled GH issues.

jaredbeck:
Released as 10.2.1

PR #1189, opened 2019-03-13
Skip excess cloning of JSONized `object_changes`

stokarenko:
Every `clone`-like operation over the `object_changes` column will allocate the `O(n)` of RAM, where `n` is a kind of complexity of `object` model in general and applied changes in particular. 

Lets skip excess `object_changes.to_hash` call when the `object_changes` is a kind of `json` column.

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -238,9 +238,14 @@ module PaperTrail
       # serialization here, using `PaperTrail.serializer`.
       #
       # @api private
+      # @param changes HashWithIndifferentAccess
       def recordable_object_changes(changes)
         if PaperTrail.config.object_changes_adapter&.respond_to?(:diff)
-          changes = PaperTrail.config.object_changes_adapter.diff(changes)
+          # We'd like to avoid the `to_hash` here, because it increases memory
+          # usage, but that would be a breaking change because
+          # `object_changes_adapter` expects a plain `Hash`, not a
+          # `HashWithIndifferentAccess`.
+          changes = PaperTrail.config.object_changes_adapter.diff(changes.to_hash)
         end
 
         if @record.class.paper_trail.version_class.object_changes_col_is_json?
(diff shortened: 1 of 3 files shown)


jaredbeck:
Somehow, this makes more sense to me when I see it as a separate PR. Looks good, I'm just going to add some comments.

jaredbeck:
Hmm this may be a breaking change, I'm not sure. 

Until now, we have passed a plain `Hash` to `object_changes_adapter.diff`. This PR would change that part of our public API, passing a `HashWithIndifferentAccess` instead.

However, I think this change would not affect the [reference implementation][1] of `object_changes_adapter`.

What do you think? Should we treat this as a breaking change?

[1]: https://github.com/hashwin/paper_trail-hashdiff/blob/master/lib/paper_trail_hashdiff.rb

jaredbeck:
I've added some comments, and fixed the `object_changes_adapter` API. Please review.

stokarenko:
I feel that change from `HashWithIndifferentAccess` to pure `Hash` can be a breaking one, but not vice versa..

On the other hand I'm able to think up some dummy `object_changes_adapter` implementation which will be sensitive to PR.

Looks like it's better to mark the change as rarely breaking,  to be honest with consumers..

jaredbeck:
> Looks like it's better to mark the change as rarely breaking, to be honest with consumers..

Please see https://github.com/paper-trail-gem/paper_trail/pull/1189/commits/309a8c120ec95af1c1506cbd94869cb52658e2fb, in which I preserve the public API by using `.to_hash`.

```diff
-         changes = PaperTrail.config.object_changes_adapter.diff(changes)
+          changes = PaperTrail.config.object_changes_adapter.diff(changes.to_hash)
```

What do you think?

stokarenko:
> I've added some comments, and fixed the `object_changes_adapter` API. Please review.

LGTM! Now there is no any chance to break the things for consumers :)

PR #1190, opened 2019-03-13
Locate missing test files for the suite

stokarenko:
Test suite will run only the test files with a `_spec` postfix.
But there are some test files which play against that convention.
As a result we did not even noticed that some tests are broken.

- [ ] Added tests.


jaredbeck:
Thank you for breaking this into smaller pieces, it's very helpful

PR #1194, opened 2019-03-18
add 'limit' option to has_paper_trail to override global PaperTrail.config.version_limit setting

enrico:
…PaperTrail.config.version_limit value on a per-model basis.

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -349,7 +349,7 @@ module PaperTrail
     #
     # TODO: Duplication: similar `constantize` in Reifier#version_reification_class
     def version_limit
-      if PaperTrail::Version.column_names.include?("item_subtype")
+      if self.class.item_subtype_column_present?
         klass = (item_subtype || item_type).constantize
         if klass&.paper_trail_options&.key?(:limit)
           return klass.paper_trail_options[:limit]
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -329,13 +329,25 @@ module PaperTrail
     # Enforces the `version_limit`, if set. Default: no limit.
     # @api private
     def enforce_version_limit!
-      limit = PaperTrail.config.version_limit
+      limit = version_limit_for_item
       return unless limit.is_a? Numeric
       previous_versions = sibling_versions.not_creates.
         order(self.class.timestamp_sort_order("asc"))
       return unless previous_versions.size > limit
       excess_versions = previous_versions - previous_versions.last(limit)
       excess_versions.map(&:destroy)
+    end
+
+    # @api private
+    def version_limit_for_item
+      if PaperTrail::Version.column_names.include?("item_subtype")
+        # we can look for and use :limit specified in class
+        klass = (item_subtype || item_type).constantize
+        if klass.respond_to?(:paper_trail_options) && klass.paper_trail_options.key?(:limit)
+          return klass.paper_trail_options[:limit]
+        end
+      end
+      PaperTrail.config.version_limit
     end
   end
 end
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -25,6 +25,10 @@ module PaperTrail
 
     # :nodoc:
     module ClassMethods
+      def item_subtype_column_present?
+        column_names.include?("item_subtype")
+      end
+
       def with_item_keys(item_type, item_id)
         where item_type: item_type, item_id: item_id
       end
@@ -329,7 +333,7 @@ module PaperTrail
     # Enforces the `version_limit`, if set. Default: no limit.
     # @api private
     def enforce_version_limit!
-      limit = version_limit_for_item
+      limit = version_limit
       return unless limit.is_a? Numeric
       previous_versions = sibling_versions.not_creates.
         order(self.class.timestamp_sort_order("asc"))
@@ -338,12 +342,16 @@ module PaperTrail
       excess_versions.map(&:destroy)
     end
 
+    # See docs section 2.e. Limiting the Number of Versions Created.
+    # The version limit can be global or per-model.
+    #
     # @api private
-    def version_limit_for_item
+    #
+    # TODO: Duplication: similar `constantize` in Reifier#version_reification_class
+    def version_limit
       if PaperTrail::Version.column_names.include?("item_subtype")
-        # we can look for and use :limit specified in class
         klass = (item_subtype || item_type).constantize
-        if klass.respond_to?(:paper_trail_options) && klass.paper_trail_options.key?(:limit)
+        if klass&.paper_trail_options&.key?(:limit)
           return klass.paper_trail_options[:limit]
         end
       end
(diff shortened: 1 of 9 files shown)


enrico:
I cannot seem to find out why this test fails (based on my changes).

 It's also interesting that one of the platforms works (Ruby: 2.3.8 Gemfile: gemfiles/ar_4.2.gemfile DB=sqlite)

So, the line failing is 
``
./spec/models/widget_spec.rb:165
``
And this is the source:
```
    it "can have multiple destruction records" do
      versions = lambda { |widget|
        # Workaround for AR 3. When we drop AR 3 support, we can simply use
        # the `widget.versions` association, instead of `with_item_keys`.
        PaperTrail::Version.with_item_keys("Widget", widget.id)
      }
      widget = Widget.create(name: 'A Widget')
      assert_equal 1, widget.versions.length
      widget.destroy
      assert_equal 2, versions.call(widget).length
      widget = widget.version.reify
      widget.save
      assert_equal 3, versions.call(widget).length
      widget.destroy
      assert_equal 4, versions.call(widget).length
      assert_equal 2, versions.call(widget).where(event: "destroy").length
    end
  end

```

after the creation and the first destruction we correctly have 2 versions of the object.

but after the next save we still have 2:
```
      widget = widget.version.reify
      widget.save
      assert_equal 3, versions.call(widget).length
```
I added a debug statement and I can see that after 'widget.version.reify', the object is not changed (call to widget.changed? returns false), so I'm assuming the save doesn't get performed? 

Any hints?

enrico:
Here's more data...

Without changing the line
```
widget = widget.version.reify
```
results in
```
widget.changed? # this is false
widget.persisted? # this is true
Widget.find(widget.id) # this is RecordNotFound
widget.save # result is true
Widget.find(widget.id) # this is STILL RecordNotFound
```

But when changing the reify line to:
```
      widget = widget.version.reify(dup: true)
```
everything works as expected and the test passes.

jaredbeck:
```ruby
has_paper_trail limit: 2
```

Hi Enrico, I like this proposed API. I feel like this idea has been discussed in the past, but don't have time to go looking for it right now.

enrico:
Jared, didn't know about the closed pr #915 .
I looked at the conversation and found out this comment by you:

> Somehow calling item in that after_create callback affects how reify works in that deleted versions test ..
> 
> Yeah. It might be because enforce_version_limit! is an after_create callback and calling item goes to the db and loads the record, changing the "persistence status" of that record somehow.
> 
> I can tell you for sure that item_type.constantize is no good, it is not compatible with STI. Consider the case of the Vehicle and the Car in our test suite. The car is the STI child, and is versioned (calls has_paper_trail). The Vechicle is the STI parent and is not versioned. Because AR stores the string "Vehicle" in versions.item_type, we cannot use item_type.constantize (because Vehicle.paper_trail_options would raise a NoMethodError).
> 
> This is a tricky one. It seems you can't call item during an after_create, but without calling item how are you going to get at the options? I don't know.

Looks like the solution could be to use item_subtype, but I read in the documents that it's supposed to be optional in the migrations... (yet the tests rely on it, since it's part of the dummy app).

I made some changes to make it work using the item_subtype attribute (by using constantize and and avoiding instantiating item). All tests work (including the one I had to change the code with .reify(dup: true)

```ruby
    def version_limit_for_item
      klass = (respond_to?(:item_subtype) && item_subtype ? item_subtype : item_type).constantize
      if klass.respond_to?(:paper_trail_options) && klass.paper_trail_options.key?(:limit)
        return klass.paper_trail_options[:limit]
      end
      PaperTrail.config.version_limit
    end
```

enrico:
Jared,

so my conclusion is that this feature (per-class limit in the has_paper_trail block) works correctly for non-STI classes but fails with STI classes when the `item_subtype` column is missing.

The code as of now uses respond_to?(:item_subtype) to cover the optionality of the column, but could be simplified if the column is actually always there. 

Questions: 

1. Is there any particular reason why the column should not be part of the default migration for paper trail 10? ( of course, this would be my favorite option :)

2. If we still want to leave `item_subtype` optional, could we still ship this and document the requirement of `item_subtype`? We could also display an error when we detect the `limit` key in the `paper_trail_options` hash and the `item_subtype` column is missing...

Thanks!
Enrico.

jaredbeck:
Hi Enrico, I haven't had time to review this PR (I haven't even read all the comments) but I'll try to quickly answer your latest questions.

> Is there any particular reason why the column [`item_subtype`] should not be part of the default migration for paper trail 10?

[It was determined](https://github.com/paper-trail-gem/paper_trail/blob/master/CHANGELOG.md#breaking-changes-most-people-wont-care-about) that most people do not need `item_subtype`. Also, by making it optional, we made it easier for people to upgrade to PT 10.

> If we still want to leave item_subtype optional, could we still ship this and document the requirement of item_subtype?

Yes, in [section 4.b.1](https://github.com/paper-trail-gem/paper_trail#4b1-the-optional-item_subtype-column) please.

> We could also display an error when we detect the limit key in the paper_trail_options hash and the item_subtype column is missing...

In theory, errors like this can be very helpful (better than docs) and I'd encourage you to add one. In practice they can be hard to implement because it can't be checked until we're certain that ActiveRecord will have loaded the schema. Generally speaking, the later you perform the check, the better.

enrico:
I think I'm done. please check when you get a chance.

jaredbeck:
Hi Enrico, I've done my first review. I've pushed [my suggestions](https://github.com/paper-trail-gem/paper_trail/pull/1194/commits/07699c04b99542fe2f5aba518a8ca7a8a605a8a8). Please review them. Summary:

- when item_subtype is absent, raise instead of warning
- copywriting
  - changelog entries should be concise
  - focus new docs in section 2.e.1
- faster tests (create fewer records)

lib/paper_trail/version_concern.rb
```diff
@@ -329,13 +333,29 @@ def object_changes_deserialized
...
     end
+
+    # See docs section 2.e. Limiting the Number of Versions Created.
+    # The version limit can be global or per-model.
+    #
+    # @api private
+    #
+    # TODO: Duplication: similar `constantize` in Reifier#version_reification_class
+    def version_limit
+      if PaperTrail::Version.column_names.include?("item_subtype")
```
enrico:
```suggestion
      if self.class.item_subtype_column_present?
```
I think this is a must, now that you've added  the class method to check for the item_subtype...

enrico:
Jared, your proposed changes look a lot cleaner than what I came up with (and I have to embarrassedly admit that in this process I  learned about `&.`)!. 

Please let me know about my suggestion above. If you agree, I can make the change and squish all commits.

jaredbeck:
> Jared, your proposed changes look a lot cleaner ..

Thanks! 😊 

> .. I learned about &. ..

Did you know that Matz calls it the "lonely operator" because it looks like a person sitting and looking at a stone? 😀 

> Please let me know about my suggestion above.

Good catch, patched by 81f760bc32f9aec179641b01cb21af54a7d0337d

> If you agree, I can make the change and squish all commits.

You don't have to, I can squash when I merge this PR. You'll still be the author of the resulting commit. I'm just waiting for CI to pass.

PR #1196, opened 2019-04-09
Fixes item_id For Versions Migration

jswright61:
Added 'limit: 8' as an additional parameter to the item_id field for the versions table. Bigint is the current default datatype for the id column of Rails tables and a reasonable choice for non Rails projects. This change allows id values to be properly stored when they would exceed the range of a 4 byte integer datatype.

Check the following boxes:

- [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,12 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1196](https://github.com/paper-trail-gem/paper_trail/pull/1196) -
+  Added 'limit: 8' as an additional parameter to the item_id field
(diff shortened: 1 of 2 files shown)


jswright61:
This is my very first PR, so please be gentle. Honestly they probably don't get much simpler than this.

The item_id being a 4 byte integer cause an out of range error when I made a change to a record on a table tracked by paper_trail:
`ActiveModel::RangeError: 918464985025609734 is out of range for ActiveModel::Type::Integer with limit 4 bytes`

I'm storing some Twitter data and decided to use the id from the Twitter User instead of an incremental id for each record. It probably would have been quite a while until I discovered this issue had I been using sequential ids starting from 1. I can only guess that people are not using this gem on super huge tables which is why nobody noticed or fixed it before.

PR #1199, opened 2019-04-23
update migration to jsonb with object_changes

localhostdotdev:
tried it on my rails 6.0.0beta3 project and worked fine, trying to make it future-proof by commenting out the `reset` which as far as I remember isn't needed on rails 5 either.

- Added tests. (tested it on my project, could add a test for it if needed)
- Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new 
  code introduces user-observable changes. (no need)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1212,14 +1212,27 @@ loop over every record and parse it in Ruby, then write to a temporary column:
 
 ```ruby
 add_column :versions, :new_object, :jsonb # or :json
+# add_column :versions, :new_object_changes, :jsonb # or :json
 
-PaperTrail::Version.reset_column_information
+# PaperTrail::Version.reset_column_information # needed for rails < 6
+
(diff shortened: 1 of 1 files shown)


jaredbeck:
I don't understand these changes. Maybe it's the commented out code that's confusing me? Can you explain what your goal is here? Thanks.

localhostdotdev:
the readme didn't cover the case when I have `object_changes` as well as `object` so I almost forgot to migrate that column too, that's why I thought of adding it to the readme.

jaredbeck:
> the readme didn't cover the case when I have `object_changes` as well as `object` so I almost forgot to migrate that column too, that's why I thought of adding it to the readme.

I see, that makes sense. And you commented it out because not everyone has `object_changes`. Sounds good.

PR #1200, opened 2019-05-10
Faster way to update from YAML to JSON.

schwern:
Rather than fetch every version and then check if it has an object, select only those with objects. For my database this cuts it from 13 million fetches to 1.7 million.


jaredbeck:
Thanks Michael

PR #1206, opened 2019-06-04
Update README: Add missing option to generator documentation

martinbjeldbak:
I ran the `--help` command on my development machine and noticed the command outputs a bit more than what's in the README. This PR updates the README to reflect this new feature.

Check the following boxes:

- N/A Added tests.
- N/A Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new 

### DIFF
--- a/README.md
+++ b/README.md
@@ -1065,7 +1065,8 @@ Usage:
   rails generate paper_trail:install [options]
 
 Options:
-  [--with-changes], [--no-with-changes]            # Store changeset (diff) with each version
+  [--skip-namespace], [--no-skip-namespace]  # Skip namespace (affects only isolated applications)
+  [--with-changes], [--no-with-changes]      # Store changeset (diff) with each version
 
 Runtime options:
   -f, [--force]                    # Overwrite files that already exist


jaredbeck:
Hi Martin. 

Looks like the `--skip-namespace` option is coming from rails, not PT. It is a change between rails 5.1 and 5.2. (https://github.com/rails/rails/commit/fee24f2189b53c19082770b5b8a62d4a5a0b830c)

What effect does this option have on PT? If none, I'd prefer to disable it, to avoid confusion.

```
# .../railties-5.2.1/lib/rails/generators/base.rb:21
class_option :skip_namespace, type: :boolean, default: false, ...
```

Maybe there's a `remove_class_option` method?

martinbjeldbak:
Interesting! Apologies, I didn't know this option came from Rails - should've checked the PT codebase for this option before opening a PR.

I didn't run the command, but I can try on the project I'm setting PT up on and report back within the next week or so. Will close this issue for now to prevent issue pollution.

Agreed that if it doesn't anything it would be great if it could be removed from the `--help` command as there's already quite a lot to take in when first setting up this awesome gem 😄

PR #1207, opened 2019-06-04
Updating new links

tusharmaroo:
Updated the new gem links for reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1603,7 +1603,7 @@ See our [contribution guidelines][43]
 
 ## Inspirations
 
-* [Simply Versioned](https://github.com/mmower/simply_versioned)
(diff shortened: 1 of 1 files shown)


jaredbeck:
Hi Tushar, 

I'm able to confirm that the acts_as_audited change is correct, but I'm not able to confirm simply_versioned. 

https://rubygems.org/gems/simply_versioned points to http://github.com/jerome/simply_versioned, not to https://github.com/mmower/simply_versioned.

tusharmaroo:
@jaredbeck Yes. You are right. Both the repos were checked out from the same trunk long back and mmower's one was older. Though the rubygems points to jerome one's so i have updated the PR as per rubygems.org.

jaredbeck:
Thanks Tushar

PR #1214, opened 2019-07-19
refactor: remove unused generator

oicitrapdraz:
At least in my case all generator's operations falls in `generators/paper_trail/install/install_generator.rb`.

I don't see the use of this definition of `InstallGenerator` class.


jaredbeck:
Hi Patricio. That's strange. Good find! Maybe it should have been deleted when `lib/generators/paper_trail/install/install_generator.rb` was created. Thanks.

jaredbeck:
What should we do with `spec/generators/paper_trail/install_generator_spec.rb`? Move it to `spec/generators/install/install_generator_spec.rb`?

oicitrapdraz:
If you are asking me, I would just delete it, but let's hear more opinions

jaredbeck:
>> What should we do with spec/generators/paper_trail/install_generator_spec.rb?

> If you are asking me, I would just delete it ...

Why would you delete the spec?

oicitrapdraz:
> Why would you delete the spec?

lol nevermind, I missread you.

oicitrapdraz:
Then I think it is reasonable to move it, when I get home I'll make the commit.

oicitrapdraz:
On second thought I think it makes more sense to move the generators inside the `lib/generators/paper_trail` folder (outside `install`), but anyway, I always ran into a problem (RSpec failed) after moving the files so I'll leave this branch clean (only with the first commit) and let you make the decision to accept the PR or not.

jaredbeck:
> .. I always ran into a problem (RSpec failed) after moving the files ..

No worries, thanks for trying, and thanks for the contribution.

PR #1216, opened 2019-07-28
Allow incompatible versions of ActiveRecord

jaredbeck:
Re: https://github.com/paper-trail-gem/paper_trail/pull/1213

For advanced users only. A warning is produced. See discussion
in paper_trail/compatibility.rb

- [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 - None
 
+### Dependencies
+
+- Allow contributors to install incompatible versions of ActiveRecord.
+  See discussion in paper_trail/compatibility.rb 
+
 ## 10.3.0 (2019-04-09)
 
 ### Breaking Changes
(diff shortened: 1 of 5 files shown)


jaredbeck:
@rafaelfranca does this work for you?

Edouard-chin:
Thanks for looking into this @jaredbeck , much appreciated 🙇 .

I think outputting a warning each time the gem is loaded is not very user friendly (even if we can disable it with a special env). It would be better if the warning was output if papertrail get installed for the first time or whenever activerecord gets updated through a `bundler update`.

We can achieve that by using a small rubygem plugin and the `Gem.post_install` hook https://github.com/Edouard-chin/paper_trail/pull/new/ec-allow-incompatible-ar

jaredbeck:
Thanks for the reviews.

> It would be better if the warning was output if papertrail get installed for the first time or whenever activerecord gets updated through a bundler update.

Interesting idea, Edouard, but I'm not comfortable with it, sorry. Once we give up the safety net of an upper-bound, normal users *will* accidentally install incompatible versions. It will happen frequently. I want to be certain that the warning is seen. We have to take care of beginners and normal users.

PR #1226, opened 2019-10-09
Support custom id key per model

santib:
## Motivation / Use case
In a project I'm currently working on we use `Heroku Connect` to sync tables from Salesforce into our Postgres DB. `Heroku Connect` will sync the existing rows in Salesforce and the primary key `id` from Rails will be automatically generated. The problem is that we can't count on that `id` because Heroku Connect could reload or resync the tables in some situations. So the `id` isn't actually identifying the row and what is actually identifying them is the column `sfid` which is handled by Salesforce (a 18 character length string). So we need to use that column as the `item_id` in PaperTrail. I thought on changing the `primary_key` of that models but then I realised it would break other things. For that reason I'm adding a per-model configuration where you can change the id to use for `item_id`.

## Notes
I feel like the tests could not be enough, but I was following the testing for "custom primary key" model. I'm open to add more tests if you can suggest me which ones could be worth it.
I'll also make the changes to the CHANGELOG if you approve moving forward with this feature.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -46,6 +46,11 @@ module PaperTrail
 
       private
 
+      def record_id
+        id_key = @record.paper_trail_options[:id_key]
+        id_key.present? ? @record[id_key] : @record.id
+      end
+
       # Rails 5.1 changed the API of `ActiveRecord::Dirty`. See
       # https://github.com/paper-trail-gem/paper_trail/pull/899
       #
--- a/lib/paper_trail/events/create.rb
+++ b/lib/paper_trail/events/create.rb
@@ -14,6 +14,7 @@ module PaperTrail
       def data
         data = {
           item: @record,
+          item_id: record_id,
           event: @record.paper_trail_event || "create",
           whodunnit: PaperTrail.request.whodunnit
         }
--- a/lib/paper_trail/events/destroy.rb
+++ b/lib/paper_trail/events/destroy.rb
@@ -13,7 +13,7 @@ module PaperTrail
       # @api private
       def data
         data = {
-          item_id: @record.id,
+          item_id: record_id,
           item_type: @record.class.base_class.name,
           event: @record.paper_trail_event || "destroy",
           whodunnit: PaperTrail.request.whodunnit
--- a/lib/paper_trail/events/update.rb
+++ b/lib/paper_trail/events/update.rb
@@ -26,6 +26,7 @@ module PaperTrail
       def data
         data = {
           item: @record,
+          item_id: record_id,
           event: @record.paper_trail_event || "update",
           whodunnit: PaperTrail.request.whodunnit
         }
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -187,6 +187,7 @@ module PaperTrail
         @model_class.versions_association_name,
         scope,
         class_name: @model_class.version_class_name,
+        primary_key: options[:id_key] || @model_class.primary_key,
         as: :item,
         **options[:versions].except(:name, :scope)
       )
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -46,7 +46,7 @@ module PaperTrail
     #  "live" item, we return nil.  Perhaps we should return self instead?
     def next_version
       subsequent_version = source_version.next
-      subsequent_version ? subsequent_version.reify : @record.class.find(@record.id)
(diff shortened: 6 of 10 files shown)


santib:
Hi @jaredbeck I'd really appreciate if you can give me some feedback on this PR. Thanks 😄

jaredbeck:
Hi Santiago, thanks for the contribution.

> So the id [can change without warning] .. what is actually [permanently] identifying them is the column sfid which is handled by Salesforce (a 18 character length string). So we need to use that column as the item_id in PaperTrail

This situation seems rare among rails apps. I don't expect many people to use this feature. I'm not opposed to adding it, but I would like to discuss some alternatives first.

My first instinct would be to add an `sfid` column to my `versions` table. PT calls such columns [metadata](https://github.com/paper-trail-gem/paper_trail#4c-storing-metadata). This would be a low-impact solution (no changes to PT) but you would be unable to use `reify`.

If you need to `reify`, we could add a new `reify` option called `basis`. For example:

```
# assuming sfid is a "metadata column" on versions table ..
basis = MyModel.find_by!(sfid: version.sfid)
version.reify(basis: basis)
```

If a `basis` were provided, then `reify` would use it instead of calling `init_model`. I believe this meets your needs, is less disruptive to other parts of PT, and might also be useful in other situations.

Any other solutions might we consider?

What concerns should we have about compatibility with the `paper_trail-association_tracking` gem?

santib:
Hey @jaredbeck, thanks for taking the time to respond 😄 

I completely agree it's a rare situation and I understand your doubts around adding these changes to the gem. I also agree that the solution is a bit "aggressive" and it makes the codebase more complex to maintain/extend.

Regarding the `metadata` solution, I thought about it but I discarded it because of two reasons: 
1. I could potentially need to ability to reify models (but you already proposed a solution)
2. Even if the `sfid` will be stored and we can match the correct row to the versions, there will still be a wrong `versions` association which can be misleading and very error prone to have in a real world project. For example imagine that I have `record id 1 and sfid A` and `record id 2 and sfid B`. For `record id 1` I have one version which has `item_id 1 and sfid A`. Suddenly `HerokuConnect` reloads the table and the ids are switched, letting us with `record id 1 sfid B` and `record id 2 sfid A`. Now, even if we can know that the version should be associated through the `sfid`, if I do `Model.find(1).versions`, I'll be getting versions from another record. And I could even reify my current object with a version of another one. So I think this is very dangerous.

Maybe another solution would be to have a gem as an extension of `paper-trail`, it's not ideal to me since it won't have much visibility and I'll be left alone with its maintenance but well.. I understand this could be considered a border case.

Regarding the `paper_trail-association_tracking` gem I don't know what impact it can have yet since I don't know how it works (I even tried to use it without success 😕). But as soon as I have some time I'll investigate this, thanks for the heads up 👍

santib:
@jaredbeck If we go with the gem, would something similar to the solution on this PR be acceptable? I'm talking about setting the `item_id` with the value the user prefers. Always assuming it does work correctly in combination with the associations gem. 

If it is acceptable then maybe if we centralize in paper-trail a `record_id` method that returns the `primary_key` and it's used in every single place where it's needed, then I can build a gem that overrides that method with the configuration provided by a user. I think that solution wouldn't be disruptive for this gem, it's mostly a refactor DRYing up how the id is fetched and maybe the need to explicitly use the method in some places where currently the primary_key is implicitly used.

What do you think? Do you have any other solution in mind?

jaredbeck:
> If we go with the gem, would something similar to the solution on this PR be acceptable?  I'm talking about setting the `item_id` with the value the user prefers.

Yes.

> If it is acceptable then maybe if we centralize in paper-trail a `record_id` method - I think it's just what you said with `primary_key_for_has_many_versions` - that returns the `primary_key` and it's used in every single place where it's needed, then I can build a gem that overrides that method with the configuration provided by a user.

Exactly. Can you write a PR for that please? No new tests or documentation will be necessary.
 
> Do you have any other solution in mind?

The separate gem is my favorite plan so far. 

Per rubygems naming conventions, the gem name should have a hypenated "paper_trail-" prefix. For example: `paper_trail-configurable_item_id`. If you want to host its repo in the `paper-trail-gem` [organization](https://github.com/paper-trail-gem) on github, you're welcome to.

santib:
@jaredbeck Sure I'll take care of this as soon as I get some time. I like the idea to host the gem in the `paper-trail-gem` organization, so it's part of the official "ecosystem". Should I create it on my own github and then transfer it over or join the org and create it there?

jaredbeck:
Closing, succeeded by https://github.com/paper-trail-gem/paper_trail/pull/1228

PR #1228, opened 2019-10-21
Introduce API for the model primary key

santib:
Relates to https://github.com/paper-trail-gem/paper_trail/pull/1226

This PR centralizes in one method the id that should be used for the `item_id`. It can be overridden by another gem to extend PT with the possibility to have configurable items ids.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,8 @@ has been destroyed.
   - [6.a. Custom Version Classes](#6a-custom-version-classes)
   - [6.b. Custom Serializer](#6b-custom-serializer)
   - [6.c. Custom Object Changes](#6c-custom-object-changes)
+  - [6.d. Excluding the Object Column](#6d-excluding-the-object-column)
+  - [6.e. Custom Primary Key for the versions Association](#6e-custom-primary-key-for-the-versions-association)
 - [7. Testing](#7-testing)
   - [7.a. Minitest](#7a-minitest)
   - [7.b. RSpec](#7b-rspec)
@@ -1339,6 +1341,13 @@ For an example of a complete and useful adapter, see
 The `object` column ends up storing a lot of duplicate data if you have models that have many columns,
 and that are updated many times. You can save ~50% of storage space by removing the column from the
 versions table. It's important to note that this will disable `reify` and `where_object`.
+
+
+### 6.e. Custom Primary Key for the versions Association
+
+The versions table by default uses the class `primary_key` for the association. If you want another column
+to be used for the association, you can override the `primary_key_for_has_many_versions` method of the
+`PaperTrail::ModelConfig` class or use the `paper_trail-configurable_item_id` gem.
 
 ## 7. Testing
 
--- a/lib/paper_trail/events/create.rb
+++ b/lib/paper_trail/events/create.rb
@@ -13,7 +13,8 @@ module PaperTrail
       # @api private
       def data
         data = {
-          item: @record,
+          item_id: record_id,
+          item_type: record_type,
           event: @record.paper_trail_event || "create",
           whodunnit: PaperTrail.request.whodunnit
         }
--- a/lib/paper_trail/events/update.rb
+++ b/lib/paper_trail/events/update.rb
@@ -25,7 +25,8 @@ module PaperTrail
       # @api private
       def data
         data = {
-          item: @record,
+          item_id: record_id,
+          item_type: record_type,
           event: @record.paper_trail_event || "update",
           whodunnit: PaperTrail.request.whodunnit
         }
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -122,6 +122,16 @@ module PaperTrail
       setup_callbacks_from_options options[:on]
     end
 
+    # Single source of truth of the primary_key used for the versions association
+    # Can be overriden by a gem to allow custom primary keys for the has_many
+    # versions association
+    # https://github.com/paper-trail-gem/paper_trail/pull/1226
+    # https://github.com/paper-trail-gem/paper_trail/pull/1228
+    # @api public
+    def primary_key_for_has_many_versions
+      @model_class.primary_key
+    end
+
     def version_class
       @_version_class ||= @model_class.version_class_name.constantize
     end
@@ -187,6 +197,7 @@ module PaperTrail
         @model_class.versions_association_name,
         scope,
         class_name: @model_class.version_class_name,
+        primary_key: primary_key_for_has_many_versions,
         as: :item,
         **options[:versions].except(:name, :scope)
       )
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -46,7 +46,7 @@ module PaperTrail
     #  "live" item, we return nil.  Perhaps we should return self instead?
     def next_version
       subsequent_version = source_version.next
-      subsequent_version ? subsequent_version.reify : @record.class.find(@record.id)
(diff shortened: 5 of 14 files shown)


santib:
@jaredbeck Is this PR fine or did I miss something?

I haven't created the gem yet but I was expecting it to be just prepending a module to the `PaperTrail::ModelConfig` which overrides `primary_key_for_has_many_versions` with a column name provided in the `has_paper_trail` method invocation.

lib/paper_trail/events/create.rb
```diff
@@ -14,6 +14,7 @@ class Create < Base
       def data
         data = {
           item: @record,
+          item_id: record_id,
```
jaredbeck: Should we be specifying both `:item` and `:item_id`?

lib/paper_trail/model_config.rb
```diff
@@ -122,6 +122,12 @@ def setup(options = {})
       setup_callbacks_from_options options[:on]
     end
 
+    # Single source of truth of the primary_key used for the versions association
```
jaredbeck: Please add comment about how this method will be used by other gem, referencing all issue and PR numbers.

Also, `@api public` would be appropriate here.

lib/paper_trail/record_trail.rb
```diff
@@ -285,10 +285,15 @@ def build_version_on_update(force:, in_after_callback:, is_touch:)
     def log_version_errors(version, action)
       version.logger&.warn(
         "Unable to create version for #{action} of #{@record.class.name}" \
-          "##{@record.id}: " + version.errors.full_messages.join(", ")
+          "##{record_id}: " + version.errors.full_messages.join(", ")
       )
     end
 
+    # @api private
```
jaredbeck: `@api private` is correct here, I think.

jaredbeck:
Also, please rebase, as I've made some fixes to the test suite.

README.md
```diff
@@ -58,6 +58,8 @@ has been destroyed.
   - [6.a. Custom Version Classes](#6a-custom-version-classes)
   - [6.b. Custom Serializer](#6b-custom-serializer)
   - [6.c. Custom Object Changes](#6c-custom-object-changes)
+  - [6.d. Excluding the Object Column](#6d-excluding-the-object-column)
```
santib: this one was missing

spec/dummy_app/app/models/custom_id_key_record.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "securerandom"
+
+module PaperTrailCustomId
+  module ModelConfig
+    def primary_key_for_has_many_versions
```
santib: My only concern with this test is that it's actually overriding the code for all of the following tests, but I couldn't find a better way to do it. The only other option I can think of would be mocking the `primary_key_for_has_many_versions ` in the `actual test file`. Thoughts?

santib:
Hey @jaredbeck I just pushed the changes. I'm open to rephrase anything I wrote. Also, I think it could be a good idea to delay merging this PR until the `paper_trail-configurable_item_id` gem is created so we can also add links to it in the `README`.

lib/paper_trail/model_config.rb
santib: I thought this was `private` because users of the `paper_trail` weren't going to override it. Anyways, I changed it

lib/paper_trail/events/create.rb
```diff
@@ -13,7 +13,8 @@ class Create < Base
       # @api private
       def data
         data = {
-          item: @record,
+          item_id: record_id,
+          item_type: @record.class.base_class.name,
```
jaredbeck: If we are going to specify `item_type`, we need to be careful about STI. Do our tests cover STI, ie. do we have assertions about the persisted `item_type`?

lib/paper_trail/events/update.rb
```diff
@@ -25,7 +25,8 @@ def initialize(record, in_after_callback, is_touch, force_changes)
       # @api private
       def data
         data = {
-          item: @record,
+          item_id: record_id,
+          item_type: @record.class.base_class.name,
```
jaredbeck: Calculation of `item_type` is duplicated here and in `create.rb`. Should we fix this duplication by extracting a method into the parent class?

lib/paper_trail/model_config.rb
jaredbeck: > I thought this was private because users of the paper_trail weren't going to use it. Anyways, I changed it

I prefer public because I see your new gem as a user of PT, and I want to treat changes to this method as breaking changes.

spec/dummy_app/app/models/custom_id_key_record.rb
jaredbeck: Re-opening `ModelConfig` is definitely a concern for me too. I'll think about it.

jaredbeck:
> Regarding the cop that failed [AbcSize, init_model] would you rather exclude it from the linter or extract klass.paper_trail.primary_key_for_has_many_versions into a private method?

Sure, please try either that extraction or maybe extract a method for `find_cond`.

lib/paper_trail/events/create.rb
santib: I took this implementation from https://github.com/paper-trail-gem/paper_trail/blob/master/lib/paper_trail/events/destroy.rb#L17 and I think it works fine with STI. I can add some tests just in case

lib/paper_trail/events/update.rb
santib: sounds good

santib:
@jaredbeck I tested a bit how do these changes play together with the associations gem and it would require some changes on that gem too. For example in the queries that use plain strings instead of hashes (e.g. https://github.com/westonganger/paper_trail-association_tracking/blob/master/lib/paper_trail_association_tracking/reifiers/has_many.rb#L102) . Do you think it's a must to get both gems working together? Being honest, the associations gem is so buggy that I'm not sure if it's worth the effort.. (btw I feel it could be a good idea to start a new gem for association tracking from scratch)

spec/dummy_app/app/models/custom_id_key_record.rb
santib: I have been thinking about this for a while. I think this test just doesn't belong to this gem, instead it belongs to the `paper_trail-configurable_item_id` gem. I feel that's the reason why it looks so bad here. Are there any other tests we can have instead?

jaredbeck: > I think this test just doesn't belong to this gem, instead it belongs to the `paper_trail-configurable_item_id` gem.

Makes sense to me. 

> Are there any other tests we can have instead?

How about if we delete this file (`models/custom_id_key_record.rb `), and stub `#primary_key_for_has_many_versions`?

```ruby
allow(SomeModel).to receive(:paper_trail).and_return(
  instance_double(
    RecordTrail, 
    primary_key_for_has_many_versions: 'uuid'
  )
)
```

Then, can we test `RecordTrail#record_id` and `Reifier#find_item`?

jaredbeck:
Hi Santiago, Coming back to this after a while, it looks good, but we just need to resolve the conversation in `custom_id_key_record.rb`? Is this something you're still working on?

santib:
Hi @jaredbeck I haven't had time to continue it yet. I can make the changes you proposed in `custom_id_key_record.rb` but I also need work on the gem & make a PR to the associations gem to make it be compatible.

santib:
Hey Jared sorry for the trouble, I think I'm gonna close this at least for now. Thanks for your time

PR #1233, opened 2020-01-02
Remove duplicate gem dependencies

aried3r:
Bundler was printing warnings because some gems where declared twice, I
believe this is because `appraisal generate` wasn't run when these were
moved to the `.gemspec` file.

For some reason I was unable to move `rails-controller-testing` to the `.gemspec` file.

- [ ] Added tests.


jaredbeck:
> .. I believe this is because appraisal generate wasn't run when these were
moved to the .gemspec file.

Sounds like a thing I would do :grin:, thanks.

PR #1235, opened 2020-02-24
Prepare for v11

jaredbeck:
Preparing for major release 11:

- Drop support for rails <= 5.1 (reached EOL when 6.0 was released,
  per https://guides.rubyonrails.org/maintenance_policy.html)
- Drop support for ruby 2.3 (reached EOL on 2019-04-01)

SemVer does not require a major version bump when dependency versions change, but our next release is going to be major anyway because of #1221 (see note in changelog also)


jaredbeck:
d6e6dc9e1e1210af429c4a0613d827ae6c5239b5 makes me happy :smile: ✂️. It's also the part that would benefit most from a second pair of eyes.

lib/paper_trail/frameworks/rails/controller.rb
```diff
@@ -25,7 +25,7 @@ def self.included(controller)
       # @api public
       def user_for_paper_trail
         return unless defined?(current_user)
-        ActiveSupport::VERSION::MAJOR >= 4 ? current_user.try!(:id) : current_user.try(:id)
+        current_user.try!(:id)
```
tlynam:
```suggestion
        current_user.try(:id) || current_user
```
What do you think of removing the rescuing?

lib/paper_trail/compatibility.rb
```diff
@@ -17,8 +17,8 @@ module PaperTrail
   # newer rails versions. Most PT users should avoid incompatible rails
   # versions.
   module Compatibility
-    ACTIVERECORD_GTE = ">= 4.2"
-    ACTIVERECORD_LT = "< 6.1"
+    ACTIVERECORD_GTE = ">= 5.2" # enforced in gemspec
```
tlynam: Would it make sense to freeze these?

paper_trail.gemspec
```diff
@@ -26,7 +26,10 @@ has been destroyed.
   s.require_paths = ["lib"]
 
   s.required_rubygems_version = ">= 1.3.6"
-  s.required_ruby_version = ">= 2.3.0"
+
+  # Ruby 2.4 reaches EoL at the end of March of 2020
```
tlynam: Would we bump this to 2.5 at the end of March?

lib/paper_trail/compatibility.rb
jaredbeck: Yes, but they're already frozen by the `frozen_string_literal` setting, right?

lib/paper_trail/frameworks/rails/controller.rb
jaredbeck: Remove the `rescue NoMethodError`? I like it. The "Commit Suggestion" feature is only showing a one-line change though, so can you please suggest the full multi-line change, or just push a commit to this branch?

paper_trail.gemspec
jaredbeck: We could, or we could wait a while. Supporting 2.4 does make our lives a little harder because rails 6 doesn't (see eg. `travis.yml`). But, it's not a big deal.

lib/paper_trail/compatibility.rb
tlynam: Oh right, thanks I missed that

lib/paper_trail/frameworks/rails/controller.rb
```diff
@@ -25,7 +25,7 @@ def self.included(controller)
...
       rescue NoMethodError
         current_user
```
tlynam:
```suggestion
        current_user.try(:id) || current_user
```

PR #1238, opened 2020-03-18
Reduce the number of requests when reifying items.

Nike0:
Hello,

During development, I've found that when we are trying to reify the destroyed record from the versions table, it tries to load this record from the original table twice.
I've added an additional condition for the `destroy` event and done some code refactoring (Rubocop didn't pass).

I don't know which type is this fix (bug or feature), so I didn't add information to the changelog file.

### DIFF
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -52,23 +52,23 @@ module PaperTrail
       # not the actual subclass. If `type` is present but empty, the class is
       # the base class.
       def init_model(attrs, options, version)
-        if options[:dup] != true && version.item
-          model = version.item
-          if options[:unversioned_attributes] == :nil
-            init_unversioned_attrs(attrs, model)
-          end
-        else
-          klass = version_reification_class(version, attrs)
-          # The `dup` option always returns a new object, otherwise we should
-          # attempt to look for the item outside of default scope(s).
-          find_cond = { klass.primary_key => version.item_id }
-          if options[:dup] || (item_found = klass.unscoped.where(find_cond).first).nil?
-            model = klass.new
-          elsif options[:unversioned_attributes] == :nil
-            model = item_found
-            init_unversioned_attrs(attrs, model)
-          end
+        klass = version_reification_class(version, attrs)
+
+        # The `dup` option and destroyed version always returns a new object,
+        # otherwise we should attempt to load item or to look for the item
+        # outside of default scope(s).
+        model = if options[:dup] == true || destroy_event?(version)
+                  klass.new
+                else
+                  find_cond = { klass.primary_key => version.item_id }
+
+                  version.item || klass.unscoped.where(find_cond).first || klass.new
+                end
+
+        if options[:unversioned_attributes] == :nil && !model.new_record?
+          init_unversioned_attrs(attrs, model)
         end
+
         model
       end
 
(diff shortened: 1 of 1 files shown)


lib/paper_trail/reifier.rb
```diff
@@ -52,23 +52,23 @@ def apply_defaults_to(options, version)
...
-            model = item_found
-            init_unversioned_attrs(attrs, model)
-          end
+        klass = version_reification_class(version, attrs)
+
+        # The `dup` option and destroyed version always returns a new object,
+        # otherwise we should attempt to load item or to look for the item
+        # outside of default scope(s).
+        model = if options[:dup] == true || destroy_event?(version)
+                  klass.new
```
gurgelrenan, line 61: The previous verification to return `klass.new` was:

```ruby
if options[:dup] || (item_found = klass.unscoped.where(find_cond).first).nil?
```

I'm not sure that *destroy_event?(version)* is the same thing here. Could you explain?

Nike0: Hi @gurgelrenan. I try to explain my changes.
First of all, the count of queries for reifying destroyed objects without option `:dup` was reduced from 2 to 0.
Here is the first place, when we try to load not existed records from the original table:
```ruby
if options[:dup] != true && version.item
```
Second place was here:
```ruby
if options[:dup] || (item_found = klass.unscoped.where(find_cond).first).nil?
```
Again, we try to find not existed record from the original table.

> The previous verification to return `klass.new` was:
> 
> ```ruby
> if options[:dup] || (item_found = klass.unscoped.where(find_cond).first).nil?
> ```
> 
> I'm not sure that _destroy_event?(version)_ is the same thing here. Could you explain?

Yes, _destroy_event?(version)_ is not the same thing. It was added as a new condition. We don't need to load not existed records. The old condition was moved below into `else` block.

In the `esle` block we trying to load item directly, then without scope and if nothing was found, return a new object.

If you want I can create a table to demonstrate all possible test cases for my fix and old solution. My fix affects only items with `destroy` action. For the `create` and `update` actions logic still the same.

gurgelrenan:
@Nike0 thanks for your explanation. This LGTM, but I would like @jaredbeck take a look.

jaredbeck:
Hi Ilya, thanks for contributing this optimization. Before I review the code changes, what is a good test I can run that demonstrates the unnecessary queries? For example, `animal_spec.rb:63` seems to `select from animals` twice.

```
  Animal Load (0.2ms)  SELECT "animals".* FROM "animals" WHERE "animals"."id" = $1 LIMIT $2  
[["id", 27], ["LIMIT", 1]]
  Cat Load (0.2ms)  SELECT "animals".* FROM "animals" WHERE "animals"."species" = $1 AND "animals"."id" = $2 ORDER BY "animals"."id" ASC LIMIT $3  
[["species", "Cat"], ["id", 27], ["LIMIT", 1]]
```

Is this a good example? Is there a better example? Thanks.

Nike0:
> Hi Ilya, thanks for contributing this optimization. Before I review the code changes, what is a good test I can run that demonstrates the unnecessary queries? For example, `animal_spec.rb:63` seems to `select from animals` twice.
> 
> ```
>   Animal Load (0.2ms)  SELECT "animals".* FROM "animals" WHERE "animals"."id" = $1 LIMIT $2  
> [["id", 27], ["LIMIT", 1]]
>   Cat Load (0.2ms)  SELECT "animals".* FROM "animals" WHERE "animals"."species" = $1 AND "animals"."id" = $2 ORDER BY "animals"."id" ASC LIMIT $3  
> [["species", "Cat"], ["id", 27], ["LIMIT", 1]]
> ```
> 
> Is this a good example? Is there a better example? Thanks.

Hi Jared.
Yes, this is a good example. There will no longer be unnecessary requests from the `animals` table with my changes.

Issue #1236, opened 2020-03-10
object_changes is stored for destroy event


bli:
The current readme page says:

> The object_changes are only stored for creation and updates, not when an object is destroyed.

But this seems outdated. For some reason, paper_trail has been changed to store object_changes on destroy event as well.

tlynam:
Thank you for catching this, would you mind opening a PR to fix this?  It seems like we can remove this line.

FanaHOVA:
See #1239

CodeMonkeySteve:
Is there a way to prevent `object_changes` from being stored on destroy events?  It seems like a waste of disk, at least for my use case.

PR #1239, opened 2020-03-26
Update README.md to remove old documentation

FanaHOVA:
Removing "The object_changes are only stored for creation and updates, not when an object is destroyed." per #1236.

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -748,9 +748,6 @@ widget.versions.last.changeset
 # {}
 ```
 
-The `object_changes` are only stored for creation and updates, not when an
-object is destroyed.
-
 Please be aware that PaperTrail doesn't use diffs internally.  When I designed
 PaperTrail I wanted simplicity and robustness so I decided to make each version
 of an object self-contained.  A version stores all of its object's data, not a


README.md
```diff
@@ -748,9 +748,6 @@ widget.versions.last.changeset
 # {}
 ```
 
-The `object_changes` are only stored for creation and updates, not when an
```
tlynam: Hi @jaredbeck, does this look good to you?  I verified on my local machine.

jaredbeck: Yeah, deleting this sentence would be correct. Looks like this change was introduced by #1123 and released in 10.0.0.

Alternatively, this sentence could say: "Prior to 10.0.0, the `object_changes` were only stored for create and update events. As of 10.0.0, they are stored for all three events." Given that 10.0.0 is somewhat recent (1.5 yr), I don't mind having this much detail in the readme.

FanaHOVA: @tlynam @jaredbeck 👍 I'll update to mention that, and link to the 10.0.0 changelog entry as well. One more question, the example shows that post-destroy, the changeset is an empty hash:

```
widget.destroy
widget.versions.last.changeset
# {}
```

but when I do it locally on my machine (running 10.3.0), I get `nil` instead. Should that be updated as well? See below:

```
2.6.4 :005 > s.versions.last
 => #<PaperTrail::Version id: 5401922, item_type: "Startup", item_id: 15766, event: "destroy", whodunnit: nil, object: "---\nid: 15766\nname: XXX\nwebsite: https://www.xx...", created_at: "2020-03-30 15:16:17"> 
2.6.4 :006 > s.versions.last.changeset
 => nil 
```

tlynam: That's interesting it returns `nil`.  Do you have the `object_changes` column?  When I run it without the `object_changes` column it returns `nil` and returns an empty hash for when I do have the column.  I don't see it in this example so I'm suspecting this is why it's behaving this way.  `=> #<PaperTrail::Version id: 5401922, item_type: "Startup", item_id: 15766, event: "destroy", whodunnit: nil, object: "---\nid: 15766\nname: XXX\nwebsite: https://www.xx...", created_at: "2020-03-30 15:16:17">`

tlynam: Hi @FanaHOVA, would you like to update the text?  Were you able to see if you had the `object_changes` column?  If that causes `changeset` to return `nil`?

PR #1244, opened 2020-05-14
Log error if create fails

KirkDiggler:


KirkDiggler:
woops, wrong repo. Sorry about that

Issue #1242, opened 2020-04-24
Generator make wrong migration for Oracle database


ukolovda:
Thank you for your contribution!

Generated migration for table `versions` have type `integer` for column `item_id`.
Latest Rails uses `bigint` for object ids.

However it works fine for PostgreSQL and SQLite3. 

But my test crash for Oracle database, because `t.integer  :item_id,   null: false, limit: 8` translated to type NUMBER(8), and have no space for 9-digits IDs from fixtures. (I use gem `activerecord-oracle_enhanced-adapter`)

**Describe the solution you'd like to build**

Please, consider changing in migration generator for table `versions` from
`t.integer  :item_id,   null: false, limit: 8`
to
`t.bigint  :item_id,   null: false`

gurgelrenan:
Hi @ukolovda, thanks for your time to report this issue. Bug reports must use the issue template https://github.com/paper-trail-gem/paper_trail/blob/master/.github/ISSUE_TEMPLATE/bug-report.md, please update your issue in order to help us.

marceloperini:
@ukolovda what version of `activerecord-oracle_enhanced-adapter` you are using? The last version use [`NUMBER(38)` as `integer`](https://github.com/rsim/oracle-enhanced/blob/master/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L393).

ukolovda:
@marceloperini , I use `activerecord-oracle_enhanced-adapter` 6.0.2
Migration generator have
`t.integer  :item_id,   null: false, limit: 8`
and for Oracle it create `NUMBER(8)` column type.

gurgelrenan:
Fixes on #1245

marceloperini:
@ukolovda can you test this issue using the `master` branch of this respository?

PR #1245, opened 2020-05-26
Convert `item_id` field to bigint

marceloperini:
Converting `item_id` field to bigint when generate the `versions` table.

When using paper_trail with PostgreSQL, MySQL and SQLite this patch will not change any field settings, but if using OracleDB it will create a `NUMBER(19)` field instead of `NUMBER(8).

* This PR are related to issue #1242


westonganger:
@jaredbeck No problems for PT-AT

marceloperini:
thanks for your review @jaredbeck and @westonganger

PR #1247, opened 2020-06-14
Updates Readme sections for `only` and `ignore`

### LINKED ISSUE #1240: Ignore or Skip attributes dynamically is not working

```ruby
# frozen_string_literal: true

require "bundler/inline"

gem 'paper_trail', '10.3.1'
gem 'rails', '5.2.3'

    t.timestamps null: false
  end

  create_table :versions do |t|
    t.string :event, null: false
    t.string :whodunnit
    t.datetime :created_at
  end
end
require "paper_trail"

  has_paper_trail on: %i[update destroy], ignore: { first_name: Proc.new { |obj|  true } }
end

  def test_1
    assert_difference(-> { PaperTrail::Version.count },  0) {
      User.update(first_name: "Jane 2")
    }
  end
end

```

tlynam:

Check the following boxes:

- [ ] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,6 +428,14 @@ a.versions.length                         # 2
 a.paper_trail.previous_version.title      # 'My Title'
 ```
 
+The `:ignore` option can also accept `Hash` arguments that we are considering deprecating.
+
+```ruby
+class Article < ActiveRecord::Base
+  has_paper_trail ignore: [:title, { color: proc { |obj| obj.color == "Yellow" } }]
+end
+```
+
 #### Only
 
 Or, you can specify a list of the `only` attributes you care about:
@@ -450,11 +458,11 @@ a.versions.length                         # 2
 a.paper_trail.previous_version.content    # nil
 ```
 
-The `:ignore` and `:only` options can also accept `Hash` arguments.
+The `:only` option can also accept `Hash` arguments that we are considering deprecating.
 
 ```ruby
 class Article < ActiveRecord::Base
-  has_paper_trail only: { title: Proc.new { |obj| !obj.title.blank? } }
+  has_paper_trail only: [{ title: Proc.new { |obj| !obj.title.blank? } }]
 end
 ```
 


tlynam:
Hi @jaredbeck, hope you're well.  It looks like `ignore` was only working for arrays.  Ignore was working for hashes when calculating `changed_and_not_ignored`.  Can we reuse this code for both cases?  Once it's flushed out, I can add tests and update the changelog.

jaredbeck:
I couldn't find documentation re: hashes in the `:ignore` array. Thankfully, we do have tests in `spec/models/article_spec.rb`. Based on my review of those tests, and `events/base.rb`, it seems like https://github.com/paper-trail-gem/paper_trail/issues/1240 is not a bug report, but rather, is proposing a new feature.

Currently, we allow the `:ignore` array to *contain* hashes. By comparison, https://github.com/paper-trail-gem/paper_trail/issues/1240 proposes that `:ignore` *be* a hash. Therefore, https://github.com/paper-trail-gem/paper_trail/issues/1240 proposes a new feature.

I'm hesitant to increase the complexity of our public API in this way.

jaredbeck:
Frankly, the current API, which allows the `:ignore` array to *contain* hashes, seems oddly complicated to me. I'd be open to simplifying it in our next major version. Perhaps `:ignore` should only accept an array of symbols, and a new option `:ignore_if` could take a `Proc`.

tlynam:
Thank you @jaredbeck!  Can you please help me understand this part of the documentation?  To me, it reads like `ignore` can be a hash?

```
The :ignore and :only options can also accept Hash arguments.

class Article < ActiveRecord::Base
  has_paper_trail only: { title: Proc.new { |obj| !obj.title.blank? } }
end
```

jaredbeck:
Ah, good find. But, looking at `Events::Base#notably_changed`, I think that documentation is incorrect?

https://github.com/paper-trail-gem/paper_trail/blob/21ead888296f022ad63d907b4b15ea598596d1e6/lib/paper_trail/events/base.rb#L221-L235

As with `:ignore`, I think that `:only` can *contain* hashes, but cannot *be* a hash.

So, now I'm not sure whether the docs are wrong or the implementation is wrong 😅 but I think they're *different*. 😓

tlynam:
Ah, thank you!  I see that now.  Okay, would the simplest route be to update the documentation for `ignore` describing that it only works with an array of symbols?  I like your idea of separating `ignore` and `ignore_if`.  Would `ignore_if` be a hash of procs?

jaredbeck:
> .. would the simplest route be to update the documentation ..

Yeah, let's just update the docs for now.

> .. [Should the docs] for `ignore` describing that it only works with an array of symbols?

Yes, but we should also give an example of an array of symbols that also *contains* a hash. See `spec/dummy_app/app/models/article.rb`. We may want to say that we are considering deprecating the later.

> I like your idea of separating `ignore` and `ignore_if`. Would `ignore_if` be a hash of procs?

I was thinking it'd be a single proc. WDYT?

tlynam:
Ah, thank you, I was confused by what you meant by `ignore_if`.

Okay, so the bug report is mentioning that a hash that's inside of an array isn't working as expected for `ignore`.  It looks like it responds to a hash within an array for `changed_and_not_ignored` and was trying to extract and use this same logic for `ignored_attr_has_changed?` as well.  It looks like `ignored_attr_has_changed?` may just work with an array without hashes.  Do you see the same thing?  Is it possible there isn't an existing test for `ignore` with a hash within an array?

jaredbeck:
>  [is there] an existing test for ignore with a hash within an array?

Yes, `spec/models/article_spec.rb`.

I'd suggest that we focus, for now, on adding examples of current behavior to the docs (section 2.c.) and hold off on discussing any changes to the code until that's been done.

README.md
```diff
@@ -428,6 +428,16 @@ a.versions.length                         # 2
...
 
+The `:ignore` option can also accept `Hash` arguments that we are considering deprecating.
+
+```ruby
+class Article < ActiveRecord::Base
+  has_paper_trail ignore: [:title, { color: proc { |obj| obj.color == "Yellow" } }]
+end
+```
+
+If an attribute is ignored and only that attribute was updated, then a version won't be created for `updated_at`.
```
tlynam, line 439: I didn't see that we document this.  Is it located somewhere else?

tlynam:
Hi @jaredbeck, I hope you're well.

1) Okay, what do you think of these Readme updates saying we support ignore with an array of symbols and hashes with an example?
2) Do we need to document that we don't save a version row for timestamp attributes if the only updated column is ignored?
3) I believe I was able to reproduce the reported bug as a failing test in `spec/models/gadget_spec.rb`.  It's saving a version record for `updated_at` when the updated column is ignored via Hash.  This doesn't occur when the column is ignored via symbol.  I believe the changes in `lib/paper_trail/events/base.rb` addresses it.  I think `changed_notably?` would evaluate `ignored_attr_has_changed?` as false when it should be true and so it would create a version row for `updated_at`.
4) If you see this as a bug, should we address it?

README.md
```diff
@@ -450,7 +460,7 @@ a.versions.length                         # 2
 a.paper_trail.previous_version.content    # nil
 ```
 
-The `:ignore` and `:only` options can also accept `Hash` arguments.
+The `:only` option can also accept `Hash` arguments that we are considering deprecating.
```
jaredbeck, line 463: I think the example that follows this line is still incorrect. It shows `only: {}`, but should be `only: []`, right?

README.md
```diff
@@ -428,6 +428,16 @@ a.versions.length                         # 2
...
 
+The `:ignore` option can also accept `Hash` arguments that we are considering deprecating.
+
+```ruby
+class Article < ActiveRecord::Base
+  has_paper_trail ignore: [:title, { color: proc { |obj| obj.color == "Yellow" } }]
+end
+```
+
+If an attribute is ignored and only that attribute was updated, then a version won't be created for timestamp attributes.
```
jaredbeck, line 439: I may be misunderstanding, but I think this has already been said, above.

jaredbeck:
>  Do we need to document that we don't save a version row for timestamp attributes if the only updated column is ignored?

Sure, but can you make it a separate PR please?

>    I believe I was able to reproduce the reported bug as a failing test in spec/models/gadget_spec.rb. It's saving a version record for updated_at when the updated column is ignored via Hash. This doesn't occur when the column is ignored via symbol. I believe the changes in lib/paper_trail/events/base.rb address it. I think changed_notably? would evaluate ignored_attr_has_changed? as false when it should be true and so it would create a version row for updated_at.
>    If you see this as a bug, should we address it?

I may be missing something, but I don't see the need to make any code changes. For now, I'm satisfied with correcting the documentation.

README.md
tlynam: Thank you, I missed that 👍

README.md
tlynam: Thank you!  Updated the example 👍

tlynam:
Thank you @jaredbeck, this PR now just updates the Readme.  I created a new PR with a failing test https://github.com/paper-trail-gem/paper_trail/pull/1256

I also created a PR to set the Rubocop gem to a fixed minor version so CI won't install a newer version that starts erroring in CI.  It also updates the Rubocop gems.  https://github.com/paper-trail-gem/paper_trail/pull/1254

PR #1253, opened 2020-07-15
CI: Drop mention of removed directive sudo: false

olleolleolle:
This is an attempt to simplify the comment in the CI configuration.

This PR removes mention of the no-longer-used Travis setting `sudo: false`. See [more at the Travis blog](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

---

Check the following boxes:

- [ ] Added tests.
- [x] The PR relates to *only* one subject with a clear title  and description in grammatically correct, complete sentences.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,13 +14,7 @@ env:
     - DB=postgres
     - DB=sqlite
 
-# We want to use `sudo: false` because the container infrastructure is supposed
-# to be faster, but Travis is having issues with containers lately ..
-#
-# > No output has been received in the last 10m0s
-#
-# .. and they recommend we use the VM infrastructure (`sudo: required`) in
-# the meantime.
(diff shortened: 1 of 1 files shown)


aried3r:
Hey! Looks like the builds are failing because of rubocop. It seems we should be locking the rubocop extensions to an exact version as well. We define `~> 1.4` for `rubocop-performance` so `1.7` is installed on Travis which is causing this, I believe.

Anyway, as for this PR, is `sudo: required` still needed at all?

batter:
Since this is just a comment you can likely do `[ci skip]` on the message to skip the build for this, though perhaps if this is no longer recommended we would be better off looking at removing this key entirely

aried3r:
Sorry for the long wait, I believe this PR is a good idea, since `sudo: false` has been deprecated and the comment no longer makes sense with current Travis.

PR #1254, opened 2020-07-27
Updates Rubocop Gems

tlynam:
- Sets rubocop-performance minor version to prevent unexpected offenses from appearing in CI.
- Updates rubocop gems.

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -247,8 +247,7 @@ module PaperTrail
       # @api private
       def prepare_object_changes(changes)
         changes = serialize_object_changes(changes)
-        changes = recordable_object_changes(changes)
-        changes
+        recordable_object_changes(changes)
       end
 
       # Returns an object which can be assigned to the `object_changes`
--- a/spec/dummy_app/config.ru
+++ b/spec/dummy_app/config.ru
@@ -2,5 +2,5 @@
 
 # This file is used by Rack-based servers to start the application.
 
-require ::File.expand_path("../config/environment", __FILE__)
+require ::File.expand_path("config/environment", __dir__)
 run Dummy::Application
--- a/spec/models/version_spec.rb
+++ b/spec/models/version_spec.rb
@@ -12,7 +12,7 @@ module PaperTrail
         specify { expect(PaperTrail.serializer).to be PaperTrail::Serializers::YAML }
 
         it "store out as a plain hash" do
-          expect(value =~ /HashWithIndifferentAccess/).to be_nil
+          expect(value).not_to include("HashWithIndifferentAccess")
         end
       end
 
@@ -46,7 +46,7 @@ module PaperTrail
         end
 
         it "store out as a plain hash" do
-          expect(value =~ /HashWithIndifferentAccess/).to be_nil
+          expect(value).not_to include("HashWithIndifferentAccess")
         end
 
         after do
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -10,7 +10,7 @@ Metrics/AbcSize:
   Max: 20 # Goal: 15
 
 Metrics/CyclomaticComplexity:
-  Max: 7 # Goal: 6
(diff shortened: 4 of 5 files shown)


tlynam:
Hi @aried3r, I created a PR to set the RuboCop Performance version so we stop introducing unexpected offenses in CI.  What do you think?

aried3r:
I like it! It's there a reason you used 1.6.1 instead of 1.7.1?

aried3r:
Maybe we should to the same for `rubocop-rspec` as well.

lib/paper_trail/events/base.rb
```diff
@@ -218,7 +218,7 @@ def notable_changes
       end
 
       # @api private
-      def notably_changed
+      def notably_changed # rubocop:disable Metrics/CyclomaticComplexity
```
tlynam, line 221: Rubocop reported this method but it appears too complex to refactor for this PR.  What do you think of disabling for this method?

spec/dummy_app/config.ru
```diff
@@ -2,5 +2,5 @@
 
 # This file is used by Rack-based servers to start the application.
 
-require ::File.expand_path("../config/environment", __FILE__)
+require ::File.expand_path("config/environment", __dir__)
```
tlynam, line 5: Rubocop reported this.  I'm not familiar with this but I'm assuming it's a safe change?

spec/models/version_spec.rb
```diff
@@ -12,7 +12,7 @@ module PaperTrail
         specify { expect(PaperTrail.serializer).to be PaperTrail::Serializers::YAML }
 
         it "store out as a plain hash" do
-          expect(value =~ /HashWithIndifferentAccess/).to be_nil
+          expect(value).not_to include("HashWithIndifferentAccess")
```
tlynam, line 15: Rubocop reported and it's easy to fix

tlynam:
Hi @aried3r, that sounds good.  Initially was trying to making the smallest change that resolved it but how about we update all the rubocop gems?  What do you think?

spec/dummy_app/config.ru
aried3r: Yes, that's alright.

lib/paper_trail/events/base.rb
aried3r: I assume this is because of recent changes in rubocop, see here: https://github.com/rubocop-hq/rubocop/pull/8149.

They also bumped the default from 6 to 7.

`paper_trail` does set a lower goal. Do you by any chance remember how high complexity was? But I agree this is too much for this PR. But rather than disabling it, I'd increase it in `.rubocop_todo.yml`.

https://github.com/paper-trail-gem/paper_trail/blob/fa8b1b270de434b882571b85ad00483817b923cb/.rubocop_todo.yml#L12-L13

I'm a bit hesitant to change it, since @jaredbeck has set these rules and goals. WDYT, @jaredbeck?

jaredbeck:
> I .. set the [development gem] version so we stop introducing unexpected offenses in CI. What do you think?

I'm 100% in favor of locking down development gems to make CI more reliable. Nice work!

> [RuboCop] bumped the default [CyclomaticComplexity] from 6 to 7.

Since we haven't achieved 6 yet, this is convenient for us 😆 . Seriously, I don't see a big difference between 6 and 7 and I'm happy either way, thanks!

aried3r:
Alright, then I think we should set the internal goal to 7 (RuboCop's new default) and the `Max` value in `.rubocop_todo.yml` to whatever is the current maximum according to rubocop instead of selectively disabling individual blocks of code.

Since by now rubocop 0.89.0 has been released, I'd like to get this shipped soon. My apologies for the slow replies.

@tlynam, feel free to leave it at 0.88.0 or update to 0.89.0 if you see fit. Thanks!

tlynam:
Hi @aried3r, what do you think of keeping it at `0.88.0` for this PR because `0.89.0` introduces another 7 rubocop offenses?

PR #1256, opened 2020-08-03
Skip creating version for timestamp when ignoring attribute with a Hash

### LINKED ISSUE #1240: Ignore or Skip attributes dynamically is not working

tlynam:
- Resolves #1240
- When a column is ignored via Hash and only that column is updated, it shouldn't create a version for `updated_at`.

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -107,7 +107,7 @@ module PaperTrail
       end
 
       # @api private
-      def changed_and_not_ignored
+      def calculated_ignored_array
         ignore = @record.paper_trail_options[:ignore].dup
         # Remove Hash arguments and then evaluate whether the attributes (the
         # keys of the hash) should also get pushed into the collection.
@@ -117,8 +117,12 @@ module PaperTrail
               ignore << attr if condition.respond_to?(:call) && condition.call(@record)
             }
         end
+      end
+
+      # @api private
+      def changed_and_not_ignored
         skip = @record.paper_trail_options[:skip]
-        (changed_in_latest_version - ignore) - skip
+        (changed_in_latest_version - calculated_ignored_array) - skip
       end
 
       # @api private
@@ -148,7 +152,7 @@ module PaperTrail
       #
       # @api private
       def ignored_attr_has_changed?
-        ignored = @record.paper_trail_options[:ignore] + @record.paper_trail_options[:skip]
+        ignored = calculated_ignored_array + @record.paper_trail_options[:skip]
         ignored.any? && (changed_in_latest_version & ignored).any?
       end
 
--- a/spec/models/gadget_spec.rb
+++ b/spec/models/gadget_spec.rb
@@ -8,12 +8,25 @@ RSpec.describe Gadget, type: :model do
   it { is_expected.to be_versioned }
 
   describe "updates", versioning: true do
-    it "generates a version for updates to `name` attribute" do
+    it "generates a version for updates" do
       expect { gadget.update_attribute(:name, "Hammer") }.to(change { gadget.versions.size }.by(1))
     end
 
-    it "ignores for updates to `brand` attribute" do
-      expect { gadget.update_attribute(:brand, "Stanley") }.not_to(change { gadget.versions.size })
+    context "ignored via symbol" do
+      it "doesn't generate a version" do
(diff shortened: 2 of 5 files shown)


spec/models/gadget_spec.rb
```diff
@@ -8,12 +8,25 @@
...
+      end
+    end
+
+    context "ignored via Hash" do
+      it "generates a version when the ignored attribute isn't true" do
+        expect { gadget.update_attribute(:color, "Blue") }.to(change { gadget.versions.size }.by(1))
+        expect(gadget.versions.last.changeset.keys).to eq %w[color updated_at]
+      end
+
+      it "doesn't generate a version when the ignored attribute is true" do
```
tlynam, line 27: This test fails.  This should succeed right?

spec/models/gadget_spec.rb
```diff
@@ -8,12 +8,25 @@
...
-    it "generates a version for updates to `name` attribute" do
+    it "generates a version for updates" do
       expect { gadget.update_attribute(:name, "Hammer") }.to(change { gadget.versions.size }.by(1))
     end
 
-    it "ignores for updates to `brand` attribute" do
-      expect { gadget.update_attribute(:brand, "Stanley") }.not_to(change { gadget.versions.size })
+    context "ignored via symbol" do
+      it "doesn't generate a version" do
+        expect { gadget.update_attribute(:brand, "Picard") }.not_to(change { gadget.versions.size })
```
tlynam, line 17: Updated name to `Picard` so the line doesn't exceed the maximum for Rubocop.

spec/models/gadget_spec.rb
tlynam: Hi @jaredbeck, so when using `ignored` with a Hash, it still creates a version for the timestamp.

spec/models/gadget_spec.rb
jaredbeck: That's fine. You can also break long lines like this:

```suggestion
        expect { 
          gadget.update_attribute(:brand, "Stanley") 
        }.not_to(change { gadget.versions.size })
```

spec/models/gadget_spec.rb
jaredbeck: > This test fails. This should succeed right?

Yeah, I think so. Nice test. 👍

lib/paper_trail/events/base.rb
```diff
@@ -148,7 +152,7 @@ def changes_in_latest_version
       #
       # @api private
       def ignored_attr_has_changed?
-        ignored = @record.paper_trail_options[:ignore] + @record.paper_trail_options[:skip]
```
tlynam, line 151: Here it was only considering ignored attributes defined in the array as symbols and not Hashes.

lib/paper_trail/events/base.rb
```diff
@@ -148,7 +152,7 @@ def changes_in_latest_version
...
+        ignored = calculated_ignored_array + @record.paper_trail_options[:skip]
```
tlynam, line 155: This ignores when attributes are either defined as symbols or hashes.

lib/paper_trail/events/base.rb
```diff
@@ -107,7 +107,7 @@ def attribute_in_previous_version(attr_name, is_touch)
       end
 
       # @api private
-      def changed_and_not_ignored
+      def calculated_ignored_array
```
tlynam, line 110: Extracts this calculation to be used by `changed_and_not_ignored` as well as `ignored_attr_has_changed?`

spec/models/gadget_spec.rb
tlynam: Sounds good!

tlynam:
Hi @jaredbeck, hope you're well.  Please let me know if I should go into more detail on this change.

PR #1258, opened 2020-08-18
Various 2020-08-18

jaredbeck:
- Preparing for 11.0.0
- Updating some dev. dependencies
- General linting TODOs

I'm not really sure what else we might want to address before releasing 11, but it's been a while, might as well do it soon unless y'all have some objection.

Some of the linting here produced some big diffs, sorry.

### DIFF
--- a/spec/paper_trail/model_spec.rb
+++ b/spec/paper_trail/model_spec.rb
@@ -279,24 +279,26 @@ RSpec.describe(::PaperTrail, versioning: true) do
     end
   end
 
-  # rubocop:disable RSpec/InstanceVariable
   context "a record's papertrail" do
-    before do
-      @date_time = Time.now
-      @time = Time.now
-      @date = Date.new(2009, 5, 29)
-      @widget = Widget.create(
+    let!(:d0) { Date.new(2009, 5, 29) }
+    let!(:t0) { Time.now }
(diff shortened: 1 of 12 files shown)


spec/paper_trail/model_spec.rb
```diff
@@ -486,249 +468,251 @@
   end
 
   context "A subclass" do
+    let(:foo) { FooWidget.create }
```
tlynam, line 471: Nit: What do you think of using `let!` everywhere or at least when we know it'll be called?

jaredbeck: I don't have strong opinions. I'd rather have any test, however written (within reason) than no test. 

Using `let!` when not needed makes tests slower. Using either form of `let`, compared to local variables, makes tests harder to understand, IMO.

My personal goal is to see the whole test on one page, so that I can understand it quickly. So, I tend to prefer local variables over either form of `let`. This leads to duplication between tests, but I prefer some duplication if it makes tests easier to understand.

jaredbeck:
Thanks for the review

spec/paper_trail/model_spec.rb
tlynam: Sounds good!  I agree, it can be difficult to scroll up a large file to see where variables are defined 👍

PR #1266, opened 2020-10-19
No need to send a public method "Object#extend"

amatsuda:
`Object#extend` has been defined as a public method since the last century (I don't exactly know when, but it already was a public method even on version 1.0.0).
So we don't need to call it via `send`.

- [ ] ~~Added tests.~~
- [ ] ~~Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes.~~


gurgelrenan:
Thanks @amatsuda! I think it's a simple and clean PR ready to merge. Any thoughts @jaredbeck ?

jaredbeck:
LGTM

PR #1268, opened 2020-11-20
Correct minor issue in code-sample on README.

tgaff:
Encountered an error when following one of the code-samples in the README.  Initially thought this may be something that was a change in an upcoming version, but that doesn't appear to be the case.

`model.live?` should be `model.paper_trail.live?`

Check the following boxes:

- [x] Added tests. 🚫 - I suspect this is unnecessary for a readme change
  code introduces user-observable changes. 🚫 does not apply

### DIFF
--- a/README.md
+++ b/README.md
@@ -703,9 +703,9 @@ it came instead from a previous version -- with `live?`:
 
 ```ruby
 widget = Widget.find 42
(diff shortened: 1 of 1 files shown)


tlynam (approved):
Thank you for updating this!

PR #1274, opened 2020-12-14
Tests: Various updates to `dummy_app`, in preparation for rails 6.1

jaredbeck:
See commit comments

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -8,13 +8,14 @@
 # > appraisal. If something is specified in both the Gemfile and an appraisal,
 # > the version from the appraisal takes precedence.
 # > https://github.com/thoughtbot/appraisal
-
+#
+#
 appraise "ar-5.2" do
-  gem "activerecord", "~> 5.2.4"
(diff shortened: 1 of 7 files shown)


Appraisals
```diff
@@ -8,13 +8,14 @@
 # > appraisal. If something is specified in both the Gemfile and an appraisal,
 # > the version from the appraisal takes precedence.
 # > https://github.com/thoughtbot/appraisal
-
+#
+#
 appraise "ar-5.2" do
-  gem "activerecord", "~> 5.2.4"
+  gem "rails", "~> 5.2.4"
```
aried3r, line 14: Should we maybe call the appraisals now `rails-5.2` etc, since it's not just `activerecord` (`ar`) anymore?

aried3r (commented):
Looking good! Should give us a more realistic dummy app for our tests :)

Appraisals
jaredbeck: Good idea

PR #1273, opened 2020-12-10
Add compatibility with Rails 6.1

sedubois:
Fixes #1272 

---

- [x] Wrote [good commit messages][1].
- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [x] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new 
  code introduces user-observable changes.
- [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.

[1]: http://chris.beams.io/posts/git-commit/

PR #1275, opened 2020-12-15
Rails 6.1

jaredbeck:
Supersedes #1273, preserving @sedubois authorship.


aried3r (approved):
LGTM!

aried3r:
Seeing how much waiting time there is with Travis these days for FLOSS projects, maybe we should consider switching to GitHub Actions at some point.

See these discussions from a different project for some reasons to make the switch and how a setup could look like for us (except we'd have a bit more setup and a bigger matrix due to different DBs):

https://github.com/shrinerb/content_disposition/pull/4#issuecomment-732498784
https://github.com/shrinerb/content_disposition/pull/6

aried3r:
Hah, just saw you're already working on that: https://github.com/paper-trail-gem/paper_trail/compare/switch_ci_to_gha

Might I recommend `ruby/setup-ruby` over `actions/setup-ruby`? It does a lot of the heavy lifting and their generation of cache-keys is a bit more stable than what's (easily) possible with the official version. One example being, that it includes the OS version of the runner, like `ubuntu-18.04`, busting the cache once you switch to `ubuntu-20.04`. We've been bitten by this at work since native extensions of gem dependencies weren't being regenerated.

But whatever you choose, the switch is the important part and the fine tuning can be done incrementally.

https://github.com/ruby/setup-ruby

jaredbeck:
> Seeing how much waiting time there is with Travis these days for FLOSS projects, maybe we should consider switching to GitHub Actions at some point.

I actually just emailed the other contributors about this, but didn't have your email address. Please email me so I can include you in that conversation.

Short story: travisci dot org is shutting down, there will be no free plan, and (pending other suggestions) I'm researching GH Actions on the `switch_ci_to_gha` (PRs into that branch welcome!)

PR #1276, opened 2020-12-16
Tests: replace TravisCI with GitHub Actions

jaredbeck:
After years of providing an awesome service for free, which we are very
grateful for, TravisCI will be dropping their free plan on Dec 31. So,
we are switching to GHA.

Drops multi-db (foo/bar) tests. Managing three databases per RDBMS
was turning into a huge hassle, and they needed to be rewritten anyway for
rails 6, per Eileen's talk.


jaredbeck:
@gurgelrenan and @aried3r I will greatly appreciate your feedback as this was my first time using GH Actions, and I'm sure there are many things I could have done better.

> Drops multi-db (foo/bar) tests. Managing three databases per RDBMS was turning into a huge hassle, and they needed to be rewritten anyway for rails 6, per Eileen's talk.

This is my biggest concern. I'd obviously rather not drop these tests.

In one sense, we have until Dec. 31 to make this switch, as that is when travisci dot com shuts down. In another sense, it'll be nice to have 5-minute builds again (Lately, TravisCI builds have taken 5 hours)

Once this is merged, please remind me to revoke the TravisCI app under Org -> OAuth settings.

aried3r:
> After years of providing an awesome service for free, which we are very  grateful for

👍 

LGTM! Personally, I'd not test the DB connection, since I'd just trust that it works and is maintained by GitHub themselves. MySQL even does health checks itself, [see logs](https://github.com/paper-trail-gem/paper_trail/pull/1276/checks?check_run_id=1561244604#step:2:77). Only postgres has the `health` commands directly since I believe the container doesn't do it itself (at least that's what it [says here](https://github.com/actions/example-services/blob/989ef69ed164330bee413f11ce9332d76f943af7/.github/workflows/postgres-service.yml#L29)).

As you can see, the longest task is bundler. And while there's `actions/cache`, I recommend [`ruby/setup-ruby`](https://github.com/ruby/setup-ruby/) which does the heavy lifting for caching, installing Rubies (even on Ubuntu 20.04) etc, using the proper bundler version, using `--no-document`, installs gems in parallel etc. It's a bit closer in usage to what the setup with Travis used to be like.

But all this is optional and we can improve upon the setup if necessary. For now I'd like test runs that are fast again. :)

.github/workflows/test.yml
```diff
@@ -0,0 +1,158 @@
+name: gha-workflow-pt-test
+on: [push, pull_request]
```
aried3r, line 2: We might to limit this to our main branch to avoid duplicate runs, but let's see. https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration

jaredbeck:
>  Personally, I'd not test the DB connection, since I'd just trust that it works and is maintained by GitHub themselves. MySQL even does health checks itself, see logs.

Good point. Those connection-tests are a just remnant of me learning how to do "GHA docker port-mapping". I'll remove them.

> .. I recommend ruby/setup-ruby which does the heavy lifting for caching [etc] ..

That sounds great. Let's do that next.

gurgelrenan:
Awesome job @jaredbeck. I have notice that the current CI time is less than 4 minutes. Great job!

My only sugestion here is use `Cache - Actions` https://github.com/marketplace/actions/cache

I'm not sure if we should change this now, but I currently use this action like this example:

```yml
      - name: Cache dependencies
        uses: actions/cache@v2
        with:
          path: vendor/bundle
          key: gem-cache-${{ hashFiles('Gemfile.lock') }}
```

jaredbeck:
Thanks Renan. It sounds like Anton recommends `ruby/setup-ruby` over `actions/cache`, and I have no idea, so I'll let you two figure that out 😁 It would be nice to cache the bundler step.

jjb:
Hi folks - I believe the tests that were removed here are the only thing testing the ability to specify a different database for the versions table (some of the discussion above seems to suggest that it's testing something else). Introduced here: https://github.com/paper-trail-gem/paper_trail/pull/289

Seems like the versions-on-different-db case is still supported in the code, but maybe would be good to have some tests for it, and documentation (had to do a lot of googling to get here).

Thanks for an awesome project!

PR #1279, opened 2020-12-17
gemspec: Exclude Gemfile and gemspec from gem

olleolleolle:
This change was inspired by the rubocop-packaging efforts, which in turn
are Debian efforts. They wrote a kind rationale for their work at https://packaging.rubystyle.guide/#using-git-in-gemspec

### DIFF
--- a/paper_trail.gemspec
+++ b/paper_trail.gemspec
@@ -23,6 +23,11 @@ has been destroyed.
   # > list, not directories, etc.
   # > https://guides.rubygems.org/specification-reference/#files
   #
+  # > Avoid using `git ls-files` to produce lists of files. Downstreams (OS
+  # > packagers) often need to build your package in an environment that does
+  # > not have git (on purpose).
+  # > https://packaging.rubystyle.guide/#using-git-in-gemspec
+  #
   # By convention, the `.gemspec` is omitted. Tests and related files (like
   # `Gemfile`) are omitted. Documentation is omitted because it would double
   # gem size. See discussion:
--- a/paper_trail.gemspec
+++ b/paper_trail.gemspec
@@ -19,9 +19,7 @@ has been destroyed.
   s.email = "jared@jaredbeck.com"
   s.license = "MIT"
 
-  s.files = `git ls-files -z`.split("\x0").select { |f|
-    f.match(%r{^(Gemfile|LICENSE|lib/|paper_trail.gemspec)})
-  }
+  s.files = Dir["lib/**/*", "LICENSE"]
   s.executables = []
   s.require_paths = ["lib"]
 
--- a/paper_trail.gemspec
+++ b/paper_trail.gemspec
@@ -19,7 +19,16 @@ has been destroyed.
   s.email = "jared@jaredbeck.com"
   s.license = "MIT"
 
-  s.files = Dir["lib/**/*", "LICENSE"]
+  # > Files included in this gem. .. Only add files you can require to this
+  # > list, not directories, etc.
+  # > https://guides.rubygems.org/specification-reference/#files
+  #
+  # By convention, the `.gemspec` is omitted. Tests and related files (like
+  # `Gemfile`) are omitted. Documentation is omitted because it would double
+  # gem size. See discussion:
+  # https://github.com/paper-trail-gem/paper_trail/pull/1279#pullrequestreview-558840513
+  s.files = Dir["lib/**/*", "LICENSE"].reject { |f| File.directory?(f) }
+
   s.executables = []
   s.require_paths = ["lib"]
 


olleolleolle:
cc @utkarsh2102

utkarsh2102:
Aw, thank you, @olleolleolle! ❤️ 
Also, this not only helps Debian, but also Ubuntu, OpenSUSE, Fedora, et al! \o/

paper_trail.gemspec
```diff
@@ -19,9 +19,7 @@ has been destroyed.
   s.email = "jared@jaredbeck.com"
   s.license = "MIT"
 
-  s.files = `git ls-files -z`.split("\x0").select { |f|
-    f.match(%r{^(Gemfile|LICENSE|lib/|paper_trail.gemspec)})
-  }
+  s.files = Dir.glob(["lib/**/*", "LICENSE"])
```
tlynam, line 22: Thank you for this change!  Can we please include `Gemfile` and `paper_trail.gemspec`?

tlynam: Judging by the title, this looks like it was on purpose.  I understand removing the git dependency but why remove `Gemfile` and the gemspec?

utkarsh2102: Well, honestly, shipping Gemfile and gemspec doesn't make much sense. There's, of course, no harm is including them but it is redundant, really! 😄

tlynam: Oh okay, it sounds like I'm missing something.  Don't we need the Gemfile to load the gemspec?

utkarsh2102: Yes, but only to build the package. What you mention in the `s.files` field are the things that will be shipped to the end user. For an end user, having Gemfile and gemspec is rather redundant.

utkarsh2102: OTOH, docs and README are something that could benefit the end users, so I'd probably ship that.

But since it wasn't shipped earlier, I'll just leave it as a slight suggestion 😄

tlynam: I see, thanks for explaining!

Hi @jaredbeck, what do you think of:
- Updating this to not depend on `git`
- Remove `Gemfile` and the gemspec
- Include the Readme and Changelog

utkarsh2102, line 22:
```suggestion
  s.files = Dir.glob(["CHANGELOG.md", "doc/*", "lib/**/*", "LICENSE", "README.md"])
```

olleolleolle: Hm, there are development-specific doc/ files. Perhaps only include the one with the warning?

paper_trail.gemspec
```diff
@@ -19,9 +19,7 @@ has been destroyed.
   s.email = "jared@jaredbeck.com"
   s.license = "MIT"
 
-  s.files = `git ls-files -z`.split("\x0").select { |f|
-    f.match(%r{^(Gemfile|LICENSE|lib/|paper_trail.gemspec)})
-  }
+  s.files = Dir.glob(["CHANGELOG.md", "doc/warning_about_not_setting_whodunnit.md", "lib/**/*", "LICENSE", "README.md"])
```
utkarsh2102, line 22: Aw, this now breaks the LineLength issue. I believe updating RuboCop simply should be enough.

OTOH, as I see it, the `README.md` is absolutely well written and I don't see the need to include `doc/`, do you think we can get rid of it altogether?

Either way, I don't mind, really and also, thanks a bunch for your work on this! ❤️

tlynam: If it helps, we can put array entries on their own lines.  I think it could more flexible to include the doc directory instead of a single file within it in case something changes.
```
s.files = Dir.glob([
  "CHANGELOG.md",
  "doc/*",
  "lib/**/*",
  "LICENSE",
  "README.md"
])
```

utkarsh2102: Sure, looks good! \o/

jaredbeck (changes requested):
Hi Olle, thanks for the contribution. Thanks Todd for reviewing.

I think the main point here is to stop using `git` to generate our file list. That's fine with me.

To keep gem size down, I'd prefer not to begin shipping documentation (changelog, readme, /doc). Adding documentation would double the size of the gem. It's my personal experience that people will read the documentation online. If I'm wrong, let's discuss this change (the introduction of docs) in a separate PR please, and focus here on the Gemfile and gemspec.

`Dir.glob` would include directories. The RubyGems documentation is contradictory on this topic. First, it says:

> Only add files you can require to this list, not directories, etc.
> https://guides.rubygems.org/specification-reference/#files

That seems clear! But then it says,

> Directories are automatically stripped from this list when building a gem, other non-files cause an error.

Without a clear answer from the specification, I'd prefer to simply not change this. That is, let's not include directories.

Finally, regarding the removal of the `.gemspec` file, I couldn't find documenation, but it seems conventional. Neither ActiveRecord nor Devise  (simply the first two gems I looked at) ship their `.gemspec`.

paper_trail.gemspec
```diff
@@ -19,9 +19,13 @@ has been destroyed.
...
-  s.files = `git ls-files -z`.split("\x0").select { |f|
-    f.match(%r{^(Gemfile|LICENSE|lib/|paper_trail.gemspec)})
-  }
+  s.files = Dir.glob([
+    "CHANGELOG.md",
+    "doc/warning_about_not_setting_whodunnit.md",
+    "lib/**/*",
+    "LICENSE",
+    "README.md"
+  ])
```
utkarsh2102, line 28: Hi @jaredbeck 👋🏻 

> I think the main point here is to stop using git to generate our file list. That's fine with me.
>
> To keep gem size down, I'd prefer not to begin shipping documentation (changelog, readme, /doc). Adding documentation would double the size of the gem. It's my personal experience that people will read the documentation online. If I'm wrong, let's discuss this change (the introduction of docs) in a separate PR please, and focus here on the Gemfile and gemspec.
>
> Without a clear answer from the specification, I'd prefer to simply not change this. That is, let's not include directories.
>
> Finally, regarding the removal of the .gemspec file, I couldn't find documenation, but it seems conventional. Neither ActiveRecord nor Devise (simply the first two gems I looked at) ship their .gemspec. Less surprising to me, neither ships a Gemfile. So, I think the removal of both files is fine.

Okay, that's fair enough! 👍🏻 
Here's something that you'd probably like to go with then:


```suggestion
  s.files = Dir["lib/**/*", "LICENSE"].reject { |f| File.directory?(f) }
```

This ensures no directories are included, git is dropped, and gemspec and Gemfile are dropped as well!
I hope we're on the same page with this^?

jaredbeck:
Out of curiosity, why does eg. Debian re-build gems? Do some companies have a rule against using rubygems.org as a gem source?

utkarsh2102:
Hello 👋🏻 

> Out of curiosity, why does eg. Debian re-build gems? Do some companies have a rule against using rubygems.org as a gem source?

Oh no, certainly not! Debian ❤️ RubyGems! 
Debian ships a lot of gems as packages. For example, I maintain the Ruby interpreter, Rails, RubyGems, and a bunch of other Ruby gems in Debian. I also maintain the paper_trail gem :)

Whilst maintaining all the gems in Debian as packages, we build the gem from source and run the tests to see if there's no regression or anything and try to fix and forward the patch to the upstream (that is here!). Debian has a policy to not ship any generated binaries but ship them after building it from source, which is then signed by the trusted key (mine in this case). And then you can install these packages from apt as usual (`apt install <package_name>`).

Please let me know if you have any more questions, I'll be happy to answer them! 😄

utkarsh2102:
> ... and forward the patch to the upstream (that is here!).

FWIW, I've created #1282 to help with the upstream and downstream collaboration and maintenance! :)

jaredbeck:
> .. I maintain the Ruby interpreter, Rails, RubyGems, and a bunch of other Ruby gems in Debian. I also maintain the paper_trail gem :)

Neat. :) Using APT to install the interpreter is something I'm familiar with, but what are the advantages of using APT to install gems (instead of eg. bundler)? Sorry if this is a basic question.

utkarsh2102:
> Using APT to install the interpreter is something I'm familiar with, but what are the advantages of using APT to install gems (instead of eg. bundler)? Sorry if this is a basic question.

Not a basic question at all, it took me some time to understand it as well! :)
So it all started when we tried to package Ruby applications for Debian, for example, gitlab (yes, `apt install gitlab` is a thing! 😄), and other rails-based applications. It is the Debian policy to not pull anything from the internet whilst doing the build, so we had to package the dependencies of the main package. Dependencies having further dependencies had to be packaged as well! 😞 

As a result, we have packaged a lot of gems that are needed by applications to build and run. So that's why we have them there. These packages exist *not* for the developers but for the users. Basically, when you install any rails application via APT, it's other Ruby dependencies will be installed, too. These packages will be used by the main package that was installed and isn't meant to be used by developers, really.

So we all use `bundler` and `rubygems` work development related stuff but have to package the same gems for APT in case there's a need for it by the main package or something.

Hope this somewhat gives you an insight into the reasons as to why we need to package the gems for the APT archive? If you didn't understand anything or have something more, please let me know! 😄

jaredbeck:
Thanks for taking the time to explain that, Utkarsh. I'd always been curious.

utkarsh2102:
No problem and thanks for all your work on paper_trail! ❤️

PR #1281, opened 2020-12-27
Fix lazy-loading of ActiveRecord

jaredbeck:
Superesedes https://github.com/paper-trail-gem/paper_trail/pull/1237, preserving Eric Proulx' authorship.

I've fixed some of my own concerns from #1237, but two remain:

1. Most people have some PaperTrail configuration in an initializer of their own, `config/initializers/paper_trail.rb`. The documentation [encourages this](https://github.com/paper-trail-gem/paper_trail/tree/v10.3.1#1e-configuration). I assume we want the end-user's initializer to happen last? If so, is there a way to make that explicit in code?
2. What is the best way to test these changes?


jaredbeck:
> .. I assume we want the end-user's initializer to happen last?

I was able to make this explicit by adding the following:

```diff
- initializer "paper_trail" do
+ initializer "paper_trail", before: "load_config_initializers" do
```

Incidentally, to my delight, it seems that Rails initializers are not a simple queue, but rather a directed, acyclic graph. See the `tsort_each` in `Rails::Initializable#run_initializers` if you're curious.

> What is the best way to test these changes?

It's difficult to test these changes with our `spec/dummy_app` because its boot process does not perfectly match a conventional Rails app. To gain the necessary confidence, we can test this branch in our real Rails apps, with the usual technique:

```ruby
# Gemfile
gem 'paper_trail', github: 'paper-trail-gem/paper_trail', branch: 'jb_lazy_load_rails'
```

I have tested this branch in this way, and it'd be great if at least one other person could too.

So, I think this is ready for final review. @ericproulx of course, and who else? @seanlinsley ? @batter ?

jaredbeck:
Our next release will most likely be a major version, 12, because `master` already contains breaking changes. This is sort of a breaking change also, so I'd like to include it in 12. It's the last thing that I'd like to include in 12. If any committers have other requests, please email me.

lib/paper_trail.rb
```diff
@@ -126,27 +109,18 @@ def version
...
+# our Railtie makes sure not to load the AR-dependent parts of PT until AR is
+# ready. A typical Rails `application.rb` has:
+#
+# ```
+# require 'rails/all' # Defines `Rails`
+# Bundler.require(*Rails.groups) # require 'paper_trail' (this file)
+# ```
+#
+# Non-rails applications should take similar care to load AR before PT.
+if defined?(Rails)
```
tlynam, line 122: Do you know if there was a reason it was previously `::Rails` instead of `Rails`?

tlynam:
I can help test this locally

lib/paper_trail.rb
jaredbeck: I think an absolute (`::`) constant name is unnecessary, because this is the top-level scope.

jaredbeck:
> I can help test this locally

Thanks Todd, how'd it go?

tlynam:
Sorry @jaredbeck, I should have mentioned I was planning on testing this weekend

tlynam:
Hi @jaredbeck, this looks good to me.  Tested:

- `PaperTrail.enabled = false` in initializer
- whodunnit in controller
- some model configs
- reify

tlynam (approved):
LGTM!

jaredbeck:
Thank you @ericproulx for writing the heart of this, and @tlynam for testing.

I'll release this shortly as 12.0.0.

PR #1285, opened 2021-01-04
Touch callback does not always create version

quainjn:
Currently `touch` with an ignored or skipped attribute will still create a new version. This changes the behavior to make it consistent with `update` callback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 - Expand kwargs passed to `save_with_version` using double splat operator - Rails 6.1 compatibility
 
+- [#1285](https://github.com/paper-trail-gem/paper_trail/pull/1285) -
+  Touch callback does not create a new version for skipped or ignored attributes
+  for ActiveRecord 6.0 and higher
+
 ### Dependencies
 
 - Drop support for ruby 2.4 (reached EoL on 2020-03-31)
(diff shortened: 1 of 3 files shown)


tlynam:
Hi @jaredbeck, I hope you're well.  This looks good to me.  Does this look good to you?

tlynam:
Do you know why the tests are failing?  Looks like the tests need to be updated?

jaredbeck:
> Currently `touch` with an ignored or skipped attribute will still create a new version.

I wonder why. Touch-tracking was implemented by @westonganger in 02b6de2d. Weston, do you remember why `on_touch` uses `force: true`? Is it OK to change that?

> .. is it reasonable to have the behavior of `touch` depend on the version of Rails between 5 and 6?

Yes, if the differences are thoroughly documented.

westonganger:
Well I see that in 02b6de2 there was no option for `is_touch` so I assume that `force: true` was required to cause the update event for `touch` when no attributes had changed. With the `is_touch` option its possible we no longer need to use the `force: true`to achieve this.

https://github.com/paper-trail-gem/paper_trail/commit/02b6de2dac30e8d3e08c2c4d4eb801a8b18a9994#diff-1f8592ba4c3060dd4ee3248e4104efc57d55d350f3234cf8562a55825b949dc0R123

jaredbeck (approved):

jaredbeck:
Released in 12.1.0

PR #1286, opened 2021-01-07
Query Versions Where Object Changed From Attributes

kevin-j-m:
This proposes an extension to the public API to provide more targeted
querying of object changes. `where_object_changes` will look for either
side of the change of the attributes provided - either versions where
the attribute changed __from__ the provided value, or changed __to__ the
provided value.

The proposed `where_object_changes_from` addition focuses only on one
side of that equation. If you want to find versions where the
attribute(s) explicitly changed *from* some known value, this will only
show those changes, as opposed to both *from* and *to*.

The first commit was required to have the test suite pass locally for me.

If there's appetite to support this additional API, I'll follow up with
`where_object_changes_to` for the other side of the equation.

------
Check the following boxes:

- [x] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1338,6 +1338,7 @@ An adapter can implement any or all of the following methods:
 1. diff: Returns the changeset in the desired format given the changeset in the original format
 2. load_changeset: Returns the changeset for a given version object
 3. where_object_changes: Returns the records resulting from the given hash of attributes.
+4. where_object_changes_from: Returns the records resulting from the given hash of attributes where the attributes changed *from* the provided value(s).
 
 Depending on what your adapter does, you may have to implement all three.
 
--- a/lib/paper_trail/queries/versions/where_object_changes_from.rb
+++ b/lib/paper_trail/queries/versions/where_object_changes_from.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+module PaperTrail
+  module Queries
+    module Versions
+      # For public API documentation, see `where_object_changes_from` in
+      # `paper_trail/version_concern.rb`.
+      # @api private
+      class WhereObjectChangesFrom
+        # - version_model_class - The class that VersionConcern was mixed into.
+        # - attributes - A `Hash` of attributes and values. See the public API
+        #   documentation for details.
+        # @api private
+        def initialize(version_model_class, attributes)
+          @version_model_class = version_model_class
+          @attributes = attributes
+        end
+
+        # @api private
+        def execute
+          if PaperTrail.config.object_changes_adapter&.respond_to?(:where_object_changes_from)
+            return PaperTrail.config.object_changes_adapter.where_object_changes_from(
+              @version_model_class, @attributes
+            )
+          end
+
+          case @version_model_class.columns_hash["object_changes"].type
+          when :jsonb, :json
+            json
+          else
+            text
+          end
+        end
+
+        private
+
+        # @api private
+        def json
+          predicates = []
+          values = []
+          @attributes.each do |field, value|
+            predicates.push(
+              "(object_changes->>? ILIKE ?)"
+            )
+            values.concat([field, "[#{value.to_json},%"])
+          end
+          sql = predicates.join(" and ")
+          @version_model_class.where(sql, *values)
+        end
+
+        # @api private
+        def text
+          arel_field = @version_model_class.arel_table[:object_changes]
+
+          @attributes.map do |field, value|
+            ::PaperTrail.serializer.where_object_changes_from_condition(arel_field, field, value)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/paper_trail/queries/versions/where_object_changes_from.rb
+++ b/lib/paper_trail/queries/versions/where_object_changes_from.rb
@@ -52,9 +52,12 @@ module PaperTrail
         def text
           arel_field = @version_model_class.arel_table[:object_changes]
 
-          @attributes.map do |field, value|
+          where_conditions = @attributes.map do |field, value|
             ::PaperTrail.serializer.where_object_changes_from_condition(arel_field, field, value)
           end
+
+          where_conditions = where_conditions.reduce { |a, e| a.and(e) }
+          @version_model_class.where(where_conditions)
         end
       end
     end
--- a/lib/paper_trail/serializers/json.rb
+++ b/lib/paper_trail/serializers/json.rb
@@ -41,6 +41,14 @@ module PaperTrail
           discussion at https://github.com/paper-trail-gem/paper_trail/issues/803
         STR
       end
+
+      # Raises an exception as this operation is not allowed from text columns.
+      def where_object_changes_from_condition(*)
+        raise <<-STR.squish.freeze
+          where_object_changes_from does not support reading JSON from a text
+          column. The json and jsonb datatypes are supported.
+        STR
+      end
     end
   end
 end
--- a/spec/models/version_spec.rb
+++ b/spec/models/version_spec.rb
@@ -273,6 +273,85 @@ module PaperTrail
               end
             end
           end
+
+          describe "#where_object_changes_from", versioning: true do
+            it "requires its argument to be a Hash" do
+              expect {
+                PaperTrail::Version.where_object_changes_from(:foo)
+              }.to raise_error(ArgumentError)
+              expect {
+                PaperTrail::Version.where_object_changes_from([])
+              }.to raise_error(ArgumentError)
+            end
+
+            context "with object_changes_adapter configured" do
+              after do
+                PaperTrail.config.object_changes_adapter = nil
+              end
+
+              it "calls the adapter's where_object_changes_from method" do
+                adapter = instance_spy("CustomObjectChangesAdapter")
+                bicycle = Bicycle.create!(name: "abc")
+                bicycle.update!(name: "xyz")
+
+                allow(adapter).to(
+                  receive(:where_object_changes_from).with(Version, name: "abc")
+                ).and_return([bicycle.versions[1]])
+
+                PaperTrail.config.object_changes_adapter = adapter
+                expect(
+                  bicycle.versions.where_object_changes_from(name: "abc")
+                ).to match_array([bicycle.versions[1]])
+                expect(adapter).to have_received(:where_object_changes_from)
+              end
+
+              it "defaults to the original behavior" do
+                adapter = Class.new.new
+                PaperTrail.config.object_changes_adapter = adapter
+                bicycle = Bicycle.create!(name: "abc")
+                bicycle.update!(name: "xyz")
+
+                if column_datatype_override
+                  expect(
+                    bicycle.versions.where_object_changes_from(name: "abc")
+                  ).to match_array([bicycle.versions[1]])
+                else
+                  expect do
+                    bicycle.versions.where_object_changes_from(name: "abc")
+                  end.to raise_error(/does not support reading YAML/)
+                end
+              end
+            end
+
+            # Only test json and jsonb columns. where_object_changes_from does
+            # not support text columns.
+            if column_datatype_override
+              it "locates versions according to their object_changes contents" do
+                widget.update!(name: name, an_integer: 0)
+                widget.update!(name: "foobar", an_integer: 100)
+                widget.update!(name: FFaker::Name.last_name, an_integer: int)
+
+                expect(
+                  widget.versions.where_object_changes_from(name: name)
+                ).to eq([widget.versions[1]])
+                expect(
+                  widget.versions.where_object_changes_from(an_integer: 100)
+                ).to eq([widget.versions[2]])
+                expect(
+                  widget.versions.where_object_changes_from(an_integer: int)
+                ).to eq([])
+                expect(
+                  widget.versions.where_object_changes_from(an_integer: 100, name: "foobar")
+                ).to eq([widget.versions[2]])
+              end
+            else
+              it "raises error" do
+                expect {
+                  widget.versions.where_object_changes_from(name: "foo").to_a
+                }.to(raise_error(/does not support reading YAML from a text column/))
+              end
+            end
+          end
         end
       end
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Added
 
-- None
+- `where_object_changes_from` queries for versions where the object's attributes
+  changed from one set of known values to any other set of values.
 
 ### Fixed
 
--- a/lib/paper_trail/serializers/yaml.rb
+++ b/lib/paper_trail/serializers/yaml.rb
@@ -38,6 +38,14 @@ module PaperTrail
           discussion at https://github.com/paper-trail-gem/paper_trail/pull/997
         STR
       end
+
+      # Raises an exception as this operation is not allowed with YAML.
+      def where_object_changes_from_condition(*)
+        raise <<-STR.squish.freeze
+          where_object_changes_from does not support reading YAML from a text
+          column. The json and jsonb datatypes are supported.
+        STR
+      end
     end
   end
 end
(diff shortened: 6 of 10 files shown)


lib/paper_trail/queries/versions/where_object_changes_from.rb
```diff
@@ -0,0 +1,62 @@
...
+          @version_model_class.where(sql, *values)
+        end
+
+        # @api private
+        def text
+          arel_field = @version_model_class.arel_table[:object_changes]
+
+          @attributes.map do |field, value|
+            ::PaperTrail.serializer.where_object_changes_from_condition(arel_field, field, value)
+          end
```
kevin-j-m, line 57: This is honestly relying on `where_object_changes_from_condition` in either serializer raising an exception. I personally feel we might be better off explicitly raising here, but I also wanted to mirror some of the implementation in [WhereObjectChanges](https://github.com/paper-trail-gem/paper_trail/blob/865f069e091d69ef9c136a27f3372f9d24eb37e9/lib/paper_trail/queries/versions/where_object_changes.rb#L64-L71), though I'm not sure if that's valuable.

lib/paper_trail/serializers/json.rb
```diff
@@ -41,6 +41,14 @@ def where_object_changes_condition(*)
...
         STR
       end
+
+      # Raises an exception as this operation is not allowed from text columns.
+      def where_object_changes_from_condition(*)
+        raise <<-STR.squish.freeze
+          where_object_changes_from does not support reading JSON from a text
+          column. The json and jsonb datatypes are supported.
+        STR
+      end
```
kevin-j-m, line 51: If we do explicitly raise in the [query class](https://github.com/paper-trail-gem/paper_trail/pull/1286/files#r553058607) itself, we can remove this method from each serializer, unless there's any other location I'm not aware of where this might get called via meta-programming in any other corners of the library I'm not familiar with.

tlynam:
Thanks @kevin-j-m, this is really interesting.  Hi @jaredbeck, what do you think?

jaredbeck:
Nice work, Kevin. It seems like an obviously useful API. However, `where_object` and `where_object_changes` already represent a lot of complexity, so I'm hesitant to take on yet more.

What do you think about extracting a `paper_trail-queries` gem? I could do the extraction (into the `paper-trail-gem` organization), and Kevin could maintain it? The new gem would probably contain all of `PaperTrail::VersionConcern::ClassMethods`. Other than asking end-users to add a line to their Gemfile, there would be no changes to the API.

As I know from extracting the `paper_trail-association_tracking` gem, there is a bit of overhead in having separate gems, but it's been a great way to share the maintenance, and the credit.

kevin-j-m:
I'll defer to members of the org for architecture and organization. I will say I feel unqualified to maintain a section of what is now the API, particularly solo, given that I haven't committed any code to the repo prior to this.

From an organizational perspective of splitting it out, my main concern is how the extraction makes sure to have an early-warning system on how changes in the "main" repo may affect it. 

I have some [history](https://github.com/kevin-j-m/clockwork-test) playing a bit of a cat and mouse game with a "main" dependency, and it's always been tougher to have that feedback loop consider the affect changes in the "main" dependency would have on this downstream, until it's already merged in, and perhaps released.

From a consumer perspective, would the expectation be that in order to utilize:

* item_subtype_column_present?
* with_item_keys
* creates
* updates
* destroys
* not_creates
* between
* timestamp_sort_order
* where_object
* where_object_changes
* primary_key_is_int?
* primary_key_is_int?
* object_changes_col_is_json?
* preceding
* subsequent

on some future release forward, one would need to add in this new gem? And that gem would, I suppose, reopen the `Version` class and add those methods on to there? Sorry about the long list, I just wanted to be exhaustive/clear.

Or would the paper_trail gem have this new gem as a dependency in its gemspec? Users would then have it even if they didn't need it, but...they'd have it - as they do today.

I'm happy to continue discussing how we can best guide this work into the hands of users, if it's desirable.

jaredbeck:
Kevin, Thanks for your quick response, and thoughtful analysis.

@tlynam, @aried3r, and @batter What do you think of the idea, at a high level? Extracting PT-AT was a success, in my opinion, reducing the complexity of PT core and attracting new maintainers.

> I will say I feel unqualified to maintain a section of what is now the API, particularly solo ..

I understand. However, this PR shows that you've acquired significant knowledge of this part of the codebase. If you don't want to maintain the proposed gem alone, we'd be happy to have you involved in any capacity.

> .. it's always been tougher to have that feedback loop consider the affect changes in the "main" dependency would have on this downstream ..

Yes, I think this is perhaps the greatest downside of having a separate gem. Upstream must follow SemVer, have a thorough changelog, and prefer deprecation. Downstream should have sensible dependency constraints.

> And that gem would, I suppose, reopen the `Version` class and add those methods on to there?

Yes, reopening the `VersionConcern` is probably simplest.

Let's give some of the other collaborators a chance to weigh in on the high-level idea before we tackle the other good questions you've raised.

tlynam:
I'm pretty new to open source / helping maintain this gem so I don't have a good sense of when to add a feature.  Do we know if users have requested this before?  Would we be confident users would use it?  Releasing it as a separate gem seems like the conservative approach considering how much effort it is to maintain the existing code.  Then we could try to gauge usage and the maintenance burden.  It would seem like it would be easier to merge it in later if problems arose and it made sense.  What does everyone think?

kevin-j-m:
Hi folks - I'm not familiar with the governance structure here, but I'm just curious where we stand on this or what we need to do to make a decision here. 

Please let me know how I can help move this forward, or bring this towards a conclusion one way or another. My apologies if this just needs more time to simmer or go through the proper channels. Without knowing that myself, I'm just looking to stay on top of this. Thanks.

jaredbeck:
Hi Kevin. Sounds like we have a consensus so far on extracting a `paper_trail-queries` gem. Unless we hear back from @aried3r or @batter saying they're opposed to the plan, I'll begin that extraction.

jaredbeck:
Here's a proof-of-concept: https://github.com/paper-trail-gem/paper_trail-queries

Kevin, I'd appreciate your help moving methods over there, following the example I've set with `.creates`.

Once we've moved all the query methods to the new gem, we can remove them here. It's a good time to be making breaking changes, because our next release is a major version anyway (12).

kevin-j-m:
Thanks @jaredbeck. Hopefully I'll have some time in the next few days to chip away at something.

I may start with getting some integration-level testing and working on how we hook into PaperTrail proper, so we can ensure that these continue to work across databases and such, like already exists in this repo. Let me know if you have thoughts around that or aspirations to handle that yourself.

seanlinsley:
>Hi Kevin. Sounds like we have a consensus so far on extracting a `paper_trail-queries` gem. Unless we hear back from @aried3r, @batter, or @seanlinsley saying they're opposed to the plan, I'll begin that extraction.

I'm very much in the "if it ain't broke, don't fix it" camp. Splitting association tracking into its own repo may have encouraged more people to contribute, but at the time that feature was quite buggy and I expect very few projects were using it. Paper Trail query methods are in a very different place: the features that are supported are well-tested, and I'd guess the majority of projects use at least one of them.

Even if it's as simple as adding a new gem to the Gemfile, I would try to avoid any downstream breaking change unless there's a clear benefit. I only see downsides here (as @kevin-j-m [mentioned](https://github.com/paper-trail-gem/paper_trail/pull/1286#issuecomment-762537261) there are recurring issues that contributors will run into when dealing with two codebases that are tightly coupled).

That said don't let me have too much influence over the decision! 😅 It's still y'all that are putting in the time.

kevin-j-m:
I'll again couch my position from the perspective of being a "relative" outsider who has been a happy user of the gem for a while, but never contributed. As such, I don't have the historical context. 

From that outsider's perspective, it would be my preference to keep these queries as part of the gem (and that's not just me trying to avoid additional responsibility - I promise!).

I wonder if this conversation could benefit from a reframing to understand what benefits we're looking to derive from splitting this off into a separate gem. We can then evaluate that vs. the costs or downsides of other approaches to determine where to go.

Are we looking to:
1. Minimize the footprint of the gem/public API because users may not want or need to take on these additional features - but we can allow them to opt in?
2. Limit the functionality in the core gem because of maintenance concerns/complexity?
3. Any other items that I'm not thinking of.

If it's item 2, I wonder if that could be achieved through additional maintainers or some different construction of the architecture of the gem (which I do not have an immediate suggestion for, if you're asking 😀 - I think that's a separate conversation, that at least for me would require more homework/investigation before I have an informed opinion, that we can go down if we get to that point). 

If it's item 1, as long as the code coming along for the ride isn't the concern, then we could still allow an opt-in mechanism by not immediately `require`ing this optional code. That would require some reorganization in how the gem is structured currently, I believe, but I think would be possible.

If it's item 1, and we're looking to minimize the actual number of bytes the gem carries with it by itself, then splitting it seems the only way to support that, if we want to provide this functionality at all (another option!).

Splitting it up does have a number of costs (both for maintainers and users). Keeping it in the repo is not without cost either, to be clear. 

While I'm not entirely opposed to splitting it up (I'm just hoping to move this forward), I think being intentional about accepting that cost because of the expected benefits seems like a valuable exercise so everyone's on the same page.

Again, I hope I'm not being too forward here; I'm looking to gather information and hopefully leave the ecosystem a little better after the fact by contributing.

jaredbeck:
Thanks for the feedback Sean, and thanks Kevin for the excellent analysis. Without a clear consensus in support of extracting a gem, let's put that idea on the back burner for now.

My primary goal in extracting a gem was to attract new maintainers and give them recognition as lead maintainers of their "component". If maintenance of the queries API becomes a problem in the future, we can reconsider this option. For now, Kevin has joined the org. Assuming they're willing to maintain at least this feature (and hopefully help with the rest of the queries API), then I'm happy to move towards merging this in to PT core.

lib/paper_trail/queries/versions/where_object_changes_from.rb
kevin-j-m: I pushed another commit which includes the same full implementation as exists in `WhereObjectChanges`. The reason for that is [Custom Serializers](https://github.com/paper-trail-gem/paper_trail#6b-custom-serializer).

However, I don't think I have a great testing story to ensure this works as expected. The [custom serializer](https://github.com/paper-trail-gem/paper_trail/blob/c3c58b739a62100b83ff79762c91f3dbd857761d/spec/support/custom_json_serializer.rb) available for specs already extends the JSON Serializer, which will raise the exception (unless we modify `CustomJsonSerializer` to *not* do that.

@jaredbeck, do you have any thoughts on if there's more work we can or should do to ensure custom serializers will play nicely with this?

README.md
```diff
@@ -1338,6 +1338,7 @@ An adapter can implement any or all of the following methods:
 1. diff: Returns the changeset in the desired format given the changeset in the original format
 2. load_changeset: Returns the changeset for a given version object
 3. where_object_changes: Returns the records resulting from the given hash of attributes.
+4. where_object_changes_from: Returns the records resulting from the given hash of attributes where the attributes changed *from* the provided value(s).
```
kevin-j-m, line 1341: I pushed a commit adding this information. @jaredbeck, do you think there's more we need to do to ensure custom adapters will work? Is there a list of known custom adapters (beyond the one mentioned in the README) that we know of?

Should we drop issues in those repos, alerting them to the additional surface area to the API being added here?

spec/models/version_spec.rb
```diff
@@ -273,6 +273,85 @@ module PaperTrail
...
           end
+
+          describe "#where_object_changes_from", versioning: true do
+            it "requires its argument to be a Hash" do
+              expect {
+                PaperTrail::Version.where_object_changes_from(:foo)
+              }.to raise_error(ArgumentError)
+              expect {
+                PaperTrail::Version.where_object_changes_from([])
+              }.to raise_error(ArgumentError)
```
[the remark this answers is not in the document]
kevin-j-m: I agree we could, and a string may be the next most likely type of argument passed in, but I'm not sure where we draw the line for an exhaustive iteration of object types to test passing in. I also don't check ints, floats, nil, or other primitive data types or ruby classes.

This was constructed to mirror the test cases in `where_object_changes`, where I have not done the archaeology on, but I imagine if this was an issue, was discussed at the point those tests were introduced? That may be an unfair assumption to make on my part - but I was relying on the exhaustiveness of the existing specs for `where_object_changes` to be the bar.

Happy to discuss if and how we should improve upon this and the tests in `where_object_changes` so everyone has the comfort and confidence needed in this feature.

kevin-j-m:
@jaredbeck, I hope you're well. Do you have any context or opinion on the two questions I have here:

1. [Custom Adapters](https://github.com/paper-trail-gem/paper_trail/pull/1286#discussion_r571616784)
2. [Custom Serializers](https://github.com/paper-trail-gem/paper_trail/pull/1286#discussion_r571616583)

Any other feedback, support, or suggestions to move this forward towards a merge would be greatly appreciated. I'm also happy to hear from anyone else!

jaredbeck:
Looks good, except I don't understand why we need `rails-controller-testing`. I'd rather not add new dev. deps. if it can be avoided.

>  Is there a list of known custom adapters (beyond the one mentioned in the README) that we know of? Should we drop issues in those repos, alerting them to the additional surface area to the API being added here?

Not that I know of. I wouldn't bother.

> .. relying on where_object_changes_from_condition in either serializer raising an exception. I personally feel we might be better off explicitly raising [in the query object] but I also wanted to mirror some of the [existing] implementation ..

Let's discuss as a subsequent PR.

kevin-j-m:
> Looks good, except I don't understand why we need `rails-controller-testing`. 

The following specs fail after you pull the repo locally:

```
paper_trail|master ⇒ rake
...
389 examples, 8 failures

Failed examples:

rspec ./spec/controllers/articles_controller_spec.rb:22 # ArticlesController PaperTrail.request.enabled? PaperTrail.enabled? == false returns false
rspec ./spec/controllers/articles_controller_spec.rb:12 # ArticlesController PaperTrail.request.enabled? PaperTrail.enabled? == true returns true
rspec ./spec/controllers/widgets_controller_spec.rb:54 # WidgetsController#destroy stores information like IP address in version
rspec ./spec/controllers/widgets_controller_spec.rb:45 # WidgetsController#destroy can be disabled
rspec ./spec/controllers/widgets_controller_spec.rb:33 # WidgetsController#create PT disabled does not save a version, and metadata is not set
rspec ./spec/controllers/widgets_controller_spec.rb:12 # WidgetsController#create PT enabled stores information like IP address in version
rspec ./spec/controllers/widgets_controller_spec.rb:67 # WidgetsController#update stores information like IP address in version
rspec ./spec/controllers/widgets_controller_spec.rb:78 # WidgetsController#update can be disabled
```

They all fail with the error message:

```
     NoMethodError:
       assigns has been extracted to a gem. To continue using it,
               add `gem 'rails-controller-testing'` to your Gemfile.
```

The Github issue to deprecate `assigns` in rails is: https://github.com/rails/rails/issues/18950. Some more information on it being removed is available [here](https://bigbinary.com/blog/changes-to-test-controllers-in-rails-5).

The extracted gem lives [here](https://github.com/rails/rails-controller-testing).

If we don't want to pull it in, then we'll need to modify those existing tests to not test i-vars in the controller directly. 

This was the first thing I did upon forking the repo, so I made sure I had a green test suite before moving forward with my change.

This is also done in the in-progress PR #1260 from @ckhall.

jaredbeck:
>> Looks good, except I don't understand why we need rails-controller-testing.

> The following specs fail after you pull the repo locally ..

Sorry, I forgot that we *do* use `rails-controller-testing` already. You can find it listed in our `Appraisals`. If tests aren't passing for you, please make sure you're using `bundle exec appraisal .... rake` as described in `CONTRIBUTING.md`.

So, I think you can remove all the `rails-controller-testing` stuff in this PR.

kevin-j-m:
I can rebase to drop the commit that adds `rails-controller-testing`. Considering that'll rewrite the SHAs anyway, shall I squash down to a single commit? Or do you prefer seeing & keeping all the commits as this changed through the PR review process?

jaredbeck:
> .. do you prefer seeing & keeping all the commits as this changed through the PR review process?

In this case, yes please, to enable incremental review. We can squash when it comes time to merge.

kevin-j-m:
In that case, I pushed an additional commit, effectively undoing the first commit, to remove rails-controller-testing. Dropping the commit via a rebase would rewrite history, which I think confuses the incremental reviews.

jaredbeck:
Thanks Kevin, nice work!

PR #1294, opened 2021-03-17
Docs: add instructions on how to configure the item association

wout:
- [ ] Added tests.

Here's a PR as [suggested on Stackoverflow](https://stackoverflow.com/questions/66641277/adding-a-counter-cache-column-on-the-version-model-for-the-paper-trail-gem/66659718#66659718). I chose to reorganise the sub-sections of section 5 by inserting this item as 5.c. If that's not what you want, I can change them back to how it was and inserting it at the bottom for example. Or rename 5.b. and make sub-sub-sections 5.b.1. and 5.b.2, if that makes sense.


jaredbeck (changes requested):
Hi @wout, thanks for this.

> I chose to reorganise the sub-sections of section 5 by inserting this item as 5.c. If that's not what you want, I can change them back to how it was and inserting it at the bottom for example. Or rename 5.b. and make sub-sub-sections 5.b.1. and 5.b.2, if that makes sense.

Let's try to avoid breaking links, if we can. I often link to README sections from S.O. answers (or mention sections by number) and I don't want to break those. Your suggestion of 5.b.1. and 5.b.2 sounds great, thanks.

wout:
@jaredbeck I've added the sub-sub-sections as you suggested. There is a bit of a duplication now in the title of 5.b. and 5.b.1., but it's clear enough I think. Let me know if you think there is a better way to structure it.

jaredbeck (changes requested):
The organization under 5.b is great, thanks. But, some links remain broken. For example, the `#5c-generators` link has been changed to `#5d-generators`.

wout:
ONG, can't believe I missed that. I'm so sorry. Should be fixed now.

jaredbeck (approved):

jaredbeck:
Thanks for the contribution!

PR #1303, opened 2021-04-05
Queries: object_changes: Simplify error instantiation

jaredbeck:
- Introduces a uniform error class, UnsupportedColumnType
- Simplifies the built-in serializers (paper_trail/serializers)

Since 9.2.0, when `object_changes_adapter` was introduced, if someone must use a
text column, and still wants to use these queries, they must write an
`object_changes_adapter`. AFAIK, no one has ever done this. The only public
adapter I know of, paper_trail-hashdiff, only supports json/b columns.

It's also theoretically possible that, after `where_object_changes` dropped
support for text columns, someone wrote a custom serializer (see
`PaperTrail.serializer=`). AFAIK, no one has done that either. Such a technique
was never documented under [6.b. Custom
Serializer](https://github.com/paper-trail-gem/paper_trail#6b-custom-serializer)

### DIFF
--- a/lib/paper_trail/queries/versions/where_attribute_changes.rb
+++ b/lib/paper_trail/queries/versions/where_attribute_changes.rb
@@ -23,12 +23,16 @@ module PaperTrail
               @version_model_class, @attribute
             )
           end
-
-          case @version_model_class.columns_hash["object_changes"].type
+          column_type = @version_model_class.columns_hash["object_changes"].type
+          case column_type
           when :jsonb, :json
             json
           else
-            text
+            raise UnsupportedColumnType.new(
+              method: "where_attribute_changes",
+              expected: "json or jsonb",
+              actual: column_type
+            )
           end
         end
 
@@ -39,15 +43,6 @@ module PaperTrail
           sql = "object_changes -> ? IS NOT NULL"
 
           @version_model_class.where(sql, @attribute)
-        end
-
-        # @api private
-        def text
-          arel_field = @version_model_class.arel_table[:object_changes]
-
-          @version_model_class.where(
-            ::PaperTrail.serializer.where_attribute_changes(arel_field, @attribute)
-          )
         end
       end
     end
--- a/lib/paper_trail/queries/versions/where_object_changes.rb
+++ b/lib/paper_trail/queries/versions/where_object_changes.rb
@@ -28,13 +28,18 @@ module PaperTrail
               @version_model_class, @attributes
             )
           end
-          case @version_model_class.columns_hash["object_changes"].type
+          column_type = @version_model_class.columns_hash["object_changes"].type
+          case column_type
           when :jsonb
             jsonb
           when :json
             json
           else
-            text
+            raise UnsupportedColumnType.new(
+              method: "where_object_changes",
+              expected: "json or jsonb",
+              actual: column_type
+            )
           end
         end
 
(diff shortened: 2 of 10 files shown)


lib/paper_trail/queries/versions/where_attribute_changes.rb
```diff
@@ -23,12 +23,16 @@ def execute
...
-          case @version_model_class.columns_hash["object_changes"].type
+          column_type = @version_model_class.columns_hash["object_changes"].type
+          case column_type
           when :jsonb, :json
             json
           else
-            text
+            raise UnsupportedColumnType.new(
+              method: "where_attribute_changes",
+              expected: "json/b",
```
kevin-j-m, line 33: Is this a canonical way to say either json or jsonb? I looked quickly through the Postgres docs, just as an example, and couldn't find any reference to it, but I don't know if we refer to that combination elsewhere in this repo in that way.

I think being explicit here may help readers of the exception's message:

```suggestion
              expected: "json or jsonb",
```

lib/paper_trail/queries/versions/where_attribute_changes.rb
```diff
@@ -40,15 +44,6 @@ def json
...
         end
-
-        # @api private
-        def text
-          arel_field = @version_model_class.arel_table[:object_changes]
-
-          @version_model_class.where(
-            ::PaperTrail.serializer.where_attribute_changes(arel_field, @attribute)
-          )
-        end
```
kevin-j-m, line 51: 🎉

lib/paper_trail/queries/versions/where_object_changes.rb
```diff
@@ -28,13 +28,18 @@ def execute
...
+          case column_type
           when :jsonb
             jsonb
           when :json
             json
           else
-            text
+            raise UnsupportedColumnType.new(
+              method: "where_object_changes",
+              expected: "json/b",
```
kevin-j-m, line 40: Some suggestion as above.

```suggestion
              expected: "json or jsonb",
```

lib/paper_trail/queries/versions/where_object_changes_from.rb
```diff
@@ -23,12 +23,16 @@ def execute
...
-          case @version_model_class.columns_hash["object_changes"].type
+          column_type = @version_model_class.columns_hash["object_changes"].type
+          case column_type
           when :jsonb, :json
             json
           else
-            text
+            raise UnsupportedColumnType.new(
+              method: "where_object_changes_from",
+              expected: "json/b",
```
kevin-j-m, line 33:
```suggestion
              expected: "json or jsonb",
```

lib/paper_trail/queries/versions/where_object_changes_to.rb
```diff
@@ -23,12 +23,16 @@ def execute
...
-          case @version_model_class.columns_hash["object_changes"].type
+          column_type = @version_model_class.columns_hash["object_changes"].type
+          case column_type
           when :jsonb, :json
             json
           else
-            text
+            raise UnsupportedColumnType.new(
+              method: "where_object_changes_to",
+              expected: "json/b",
```
kevin-j-m, line 33:
```suggestion
              expected: "json or jsonb",
```

kevin-j-m (commented):
I think this'll be great to not have to carry this around anymore.

I have one suggestion that touches a few places in how the error message is conveyed to users for clarity, but I'm on board otherwise!

jaredbeck:
Thanks for the quick review, Kevin.

PR #1304, opened 2021-04-05
Change default object* column type from text to json

jaredbeck:
- It's hard to change _post facto_
- Our object_changes query methods do not support text columns
- MySQL has some support for json columns now
  (https://dev.mysql.com/doc/refman/8.0/en/json.html)
- If you have to use text, you can edit your migration

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Experts: to install incompatible versions of activerecord, see
 1. Add a `versions` table to your database:
 
     ```
-    bundle exec rails generate paper_trail:install [--with-changes]
+    bundle exec rails generate paper_trail:install --with-changes
     ```
 
     For more information on this generator, see [section 5.c.
@@ -1105,10 +1105,20 @@ Be advised that redefining an association is an undocumented feature of Rails.
 ### 5.c. Generators
 
 PaperTrail has one generator, `paper_trail:install`. It writes, but does not
-run, a migration file.
-The migration adds (at least) the `versions` table. The
-most up-to-date documentation for this generator can be found by running `rails
-generate paper_trail:install --help`, but a copy is included here for
+run, a migration file. The migration creates the `versions` table.
+
+```
+rails generate paper_trail:install --with-changes
+```
+
+The `--with-changes` option is recommended. It adds an `object_changes` column,
+which enables [diffing](#3c-diffing-versions) and [queries](#3e-queries). This
+optional column can be deleted later, if necessary.
+
+#### Reference
+
+The most up-to-date documentation for this generator can be found by running
+`rails generate paper_trail:install --help`, but a copy is included here for
 convenience.
 
 ```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1105,10 +1105,20 @@ Be advised that redefining an association is an undocumented feature of Rails.
...
-generate paper_trail:install --help`, but a copy is included here for
+run, a migration file. The migration creates the `versions` table.
+
+```
+rails generate paper_trail:install --with-changes
+```
+
+The `--with-changes` option is recommended. It adds an `object_changes` column,
+which enables [diffing](#3c-diffing-versions) and [queries](#3e-queries). This
+optional column can be deleted later, if necessary.
```
kevin-j-m, line 1116: Do we need to discuss or make clear any limitations for MySQL users for the json data type? 

I *did* notice in my last few PRs that when we are testing the JSON `object_changes_*` methods, we only, I believe, do so only in Postgres, not MySQL, so I don't know if we have a full picture of what the MySQL compatibility story is here - unless I was misreading the test suite setup.

jaredbeck: > I _did_ notice in my last few PRs that when we are testing the JSON `object_changes_*` methods, we only, I believe, do so only in Postgres ..

That's correct. I'll look into enabling those tests for MySQL.

gurgelrenan (commented):
@jaredbeck Great improvement! 

How about create some topic on the docs to users that want to migrate their columns? I know that is up to the user this upgrade, but maybe we should post a example about how to do.

jaredbeck:
> How about create some topic on the docs to users that want to migrate their columns?

I think we already have this: https://github.com/paper-trail-gem/paper_trail#convert-existing-yaml-data-to-json

gurgelrenan (changes requested):
LGTM! Just need to solve conflicts on `README.md`

jaredbeck:
Preliminary testing indicates that changing the default column type is premature, assuming we want MySQL support.

MySQL does support JSON, but I believe our queries are Postgres-specific. For example, in Postgres vs. MySQL, the `->>` operator takes semantically-different arguments.

### MariaDB

MariaDB presents additional challenges. It has the JSON functions, but no operators have been implemented [yet](https://jira.mariadb.org/browse/MDEV-13594). We may be able to rewrite our queries using functions like `json_extract` instead. 

Worse, the `json` type is just an alias for `longtext`. We may have to explicitly set `attribute :object, :json` in our `VersionConcern` (and our conditionals would then have to check `type_for_attribute` instead of `columns_hash`)

### References

- https://www.postgresql.org/docs/13/functions-json.html
- https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html
- https://mariadb.com/kb/en/json-functions/
- https://mariadb.com/kb/en/jsonpath-expressions/

Issue #1307, opened 2021-04-09
Unable to load schema with limit option


ryanb:
If I add the `limit` option to `has_paper_trail` it will cause commands such as `rails db:schema:load` to fail. This prevents setting up the app in a new environment or running the test suite since the schema isn't loaded. I assume this is because the limit option depends on the `versions` table which is not created until after the schema is loaded.

---

- [x] This is not a usage question, this is a bug report
- [x] This bug can be reproduced with the script I provide below
- [x] This bug can be reproduced in the latest release of the `paper_trail` gem

```ruby
# frozen_string_literal: true

# Use this template to report PaperTrail bugs.
# Please include only the minimum code necessary to reproduce your issue.
require "bundler/inline"

# STEP ONE: What versions are you using?
gemfile(true) do
  ruby "2.5.5"
  source "https://rubygems.org"
  gem "activerecord", "6.0.3.6"
  gem "minitest", "5.11.3"
  gem "paper_trail", "12.0.0", require: false
  gem "sqlite3", "1.4.2"
end

require "active_record"
require "minitest/autorun"
require "logger"

# Please use sqlite for your bug reports, if possible.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = nil
ActiveRecord::Schema.define do
  # STEP TWO: Define your tables here.
  create_table :users, force: true do |t|
    t.text :first_name, null: false
    t.timestamps null: false
  end

  # versions table created after AR model is defined
end
ActiveRecord::Base.logger = Logger.new(STDOUT)
require "paper_trail"

# STEP FOUR: Define your AR models here.
class User < ActiveRecord::Base
  # Removing limit option works
  has_paper_trail limit: 5
end

# Define versions table after model is loaded to simulate db:schema:load command
ActiveRecord::Schema.define do
  create_table :versions do |t|
    t.string :item_type, null: false
    t.integer :item_id, null: false
    t.string :item_subtype
    t.string :event, null: false
    t.string :whodunnit
    t.text :object, limit: 1_073_741_823
    t.text :object_changes, limit: 1_073_741_823
    t.datetime :created_at
  end
  add_index :versions, %i[item_type item_id]
end

# STEP FIVE: Please write a test that demonstrates your issue.
class BugTest < ActiveSupport::TestCase
  def test_1
    assert_difference(-> { PaperTrail::Version.count }, +1) {
      User.create(first_name: "Jane")
    }
[...]

PR #1309, opened 2021-04-14
Removes item_subtype requirement for model-specific limits.

PhilCoggins:
* When enforcing limits for a model, it was previously required to
  have the item_subtype column on the versions table to obtain the
  configured limit for a given model type, especially for models
  subclassed via STI.

* This change instead references the item on the version record,
  which is already available via the association cache, to determine
  the properly configured limit. This change drops the requirement of
  having item_subtype to configure model-specific limits.

Assuming there are no issues with simply referencing `item.class` to determine the configured limit, this should be a better setup, as it doesn't require `item_subtype`.

Note: The example that I removed assumes that in the absence of `item_subtype`, the global limit would be used - even if a model had its own configured limit. This spec was irrelevant because it was not possible to have simultaneously defined `limit` with `has_paper_trail` and not have `item_subtype` in your `versions` schema, because an exception would be thrown.

Fixes https://github.com/paper-trail-gem/paper_trail/issues/1307

### DIFF
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -18,11 +18,6 @@ module PaperTrail
       `abstract_class`. This is fine, but all application models must be
       configured to use concrete (not abstract) version models.
     STR
-    E_MODEL_LIMIT_REQUIRES_ITEM_SUBTYPE = <<~STR.squish.freeze
-      To use PaperTrail's per-model limit in your %s model, you must have an
-      item_subtype column in your versions table. See documentation sections
-      2.e.1 Per-model limit, and 4.b.1 The optional item_subtype column.
-    STR
     DPR_PASSING_ASSOC_NAME_DIRECTLY_TO_VERSIONS_OPTION = <<~STR.squish
       Passing versions association name as `has_paper_trail versions: %{versions_name}`
       is deprecated. Use `has_paper_trail versions: {name: %{versions_name}}` instead.
@@ -124,7 +119,6 @@ module PaperTrail
       @model_class.send :include, ::PaperTrail::Model::InstanceMethods
       setup_options(options)
       setup_associations(options)
-      check_presence_of_item_subtype_column(options)
       @model_class.after_rollback { paper_trail.clear_rolled_back_versions }
       setup_callbacks_from_options options[:on]
     end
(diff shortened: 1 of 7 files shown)


PhilCoggins:
@jaredbeck Any thoughts on this approach?

jaredbeck:
@aried3r Do you have time to look at this? I've been busy with some other stuff lately.

aried3r:
I had a quick look. It seems reasonable and all specs pass. Are we perhaps lacking a spec that proves that it works without the `item_subtype` column now? Since the removed spec just documented that it was not possible before.

lib/paper_trail/version_concern.rb
```diff
@@ -379,16 +379,22 @@ def sibling_versions
...
     def version_limit
-      if self.class.item_subtype_column_present?
-        klass = (item_subtype || item_type).constantize
-        if klass&.paper_trail_options&.key?(:limit)
-          return klass.paper_trail_options[:limit]
-        end
+      if limit_option?(item.class)
+        item.class.paper_trail_options[:limit]
+      elsif base_class_limit_option?(item.class)
+        item.class.base_class.paper_trail_options[:limit]
```
jaredbeck, line 386: This is cool, I had forgotten about `base_class`.

It seems this branch is untested, according to SimpleCov.

lib/paper_trail/model_config.rb
```diff
@@ -151,16 +145,6 @@ def cannot_record_after_destroy?
       ::ActiveRecord::Base.belongs_to_required_by_default
     end
 
-    # Some options require the presence of the `item_subtype` column. Currently
-    # only `limit`, but in the future there may be others.
-    #
-    # @api private
-    def check_presence_of_item_subtype_column(options)
-      return unless options.key?(:limit)
-      return if version_class.item_subtype_column_present?
```
jaredbeck, line 160: I think the `item_subtype_column_present?` method can be deleted now? That was the only call site, right?

jaredbeck (changes requested):

jaredbeck:
@PhilCoggins sorry for the delayed review. Please let us know if you have trouble making the requested changes.

jaredbeck:
> jaredbeck force-pushed the PhilCoggins:remove_item_subtype_req_for_limit branch from d79d6d6 to bca9aa5

Rebasing on master as development dependencies have been updated. Also, I've added a minimum test coverage requirement.

lib/paper_trail/model_config.rb
PhilCoggins: Correct 👍

lib/paper_trail/version_concern.rb
PhilCoggins: Yes, I'll get something added.

PhilCoggins:
@jaredbeck I'll get this updated early this week, today or tomorrow evening.

PhilCoggins:
@jaredbeck Updated

jaredbeck (approved):

jaredbeck:
Thanks Phil, nice work

jaredbeck:
Released in 12.1.0

PR #1314, opened 2021-05-07
Docs: defunct `item_id`s

jaredbeck:
Just some thoughts on id reuse. Any obvious preventative measures that I'm missing?


README.md
```diff
@@ -803,6 +805,31 @@ be a `json` or `jsonb` column if possible. If you must use a `text` column,
 you'll have to write a [custom
 `object_changes_adapter`](#6c-custom-object-changes).
 
+### 3.f. Defunct `item_id`s
+
+The `item_id`s in your `versions` table can become defunct over time,
+potentially causing application errors when `id`s in the foreign table are
+reused. `id` reuse can be an explicit choice of the application, or implicitly
+caused by sequence cycling. The chance of `id` reuse is reduced (but not
+eliminated) with `bigint` `id`s, `no cycle` sequences, and/or when `versions`
```
tlynam, line 814: Should we define what a `no cycle` sequence is?  Is it like a `uuid`?

tlynam: Sounds good!  Should we list `uuid` as an option to avoid this id reuse issue?

PR #1327, opened 2021-06-24
Problem: wrong schema created by the generator

yrashk:
'rails generate paper_trail:install' generates a subtly wrong
schema.

```
$ rails generate paper_trail:install
      create  db/migrate/20210624155730_create_versions.rb
$ db:migrate
== 20210624155730 CreateVersions: migrating ===================================
-- create_table(:versions)
   -> 0.0010s
-- add_index(:versions, [:item_type, :item_id])
   -> 0.0003s
== 20210624155730 CreateVersions: migrated (0.0013s) ==========================
```

When you look into the created table, it has this funny "{:null=>false}"
column:

```
$ sqlite3 db/development.sqlite3
SQLite version 3.32.3 2020-06-18 14:16:19
Enter ".help" for usage hints.
sqlite> pragma table_info(versions);
0|id|integer|1||1
1|item_type|varchar|0||0
2|{:null=>false}|varchar|0||0
3|item_id|bigint|1||0
4|event|varchar|1||0
5|whodunnit|varchar|0||0
6|object|text(1073741823)|0||0
7|created_at|datetime|0||0
```

Solution: ensure proper column syntax is used

This comes from the first line in the `versions` table migration:

```ruby
t.string   :item_type, { null: false }
```

When `{ null: false }` is replaced with `null: false`, th
re-created schema looks like this:

```
$ sqlite3 db/development.sqlite3
SQLite version 3.32.3 2020-06-18 14:16:19
Enter ".help" for usage hints.
sqlite> pragma table_info(versions);
0|id|integer|1||1
1|item_type|varchar|1||0
2|item_id|bigint|1||0
3|event|varchar|1||0
4|whodunnit|varchar|0||0
5|object|text(1073741823)|0||0
6|created_at|datetime|0||0
```

This seems to come from this subtlety of how shorthand methods are implemented in
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb#L261-L270

Specifically, I think the reason for the anove bug is that it'll hapilly iterate
over given names, whether they happen to be strings/symbols or not.

In the ideal world, perhaps `define_column_methods`' generated methods should have
verified if a given name is a `Hash` or not. But the world isn't ideal,
so I propose to address this issue in paper_trail. I may open a separate issue with
the Rails team to see if anything should be done on their end going further.

---

- [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 - [#1285](https://github.com/paper-trail-gem/paper_trail/pull/1285) -
   For ActiveRecord >= 6.0, the `touch` callback will no longer create a new
   `Version` for skipped or ignored attributes.
+- Newly generated migrations won't generate a "{:null=>false}" column. Existing
+  tables will need to drop this column and make item_type non-nullable.
 
 ## 12.0.0 (2021-03-29)
 
--- a/lib/generators/paper_trail/install/install_generator.rb
+++ b/lib/generators/paper_trail/install/install_generator.rb
@@ -41,9 +41,9 @@ module PaperTrail
     # See https://github.com/paper-trail-gem/paper_trail/issues/651
     def item_type_options
       if mysql?
-        ", { null: false, limit: 191 }"
+        ", null: false, limit: 191"
       else
-        ", { null: false }"
+        ", null: false"
       end
     end
 


CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 - [#1285](https://github.com/paper-trail-gem/paper_trail/pull/1285) -
   For ActiveRecord >= 6.0, the `touch` callback will no longer create a new
   `Version` for skipped or ignored attributes.
+- Newly generated migrations won't generate a "{:null=>false}" column. Existing
```
tlynam, line 26: Should we specify that this is for Ruby 3?  Are you able to confirm this as well?

tlynam:
Thank you for opening this PR @yrashk!  This fix looks good to me.  It looks like this may be a result of the Ruby 3 keyword argument change:
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

When I run this locally with 2.7, I receive the warning: `Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`.  The existing code worked when I was using Ruby 2.7 and was able to reproduce the issue with Ruby 3.

What do you think @jaredbeck?

jaredbeck:
> When I run this locally with 2.7, I receive the warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call. The existing code worked when I was using Ruby 2.7 and was able to reproduce the issue with Ruby 3.

Thanks for testing manually, Todd! PT supports all living rubies (currently 2.6 - 3.0) so, let's take care not to introduce a warning like this.

What is the plan for automated testing, @yrashk ? How about adding something in `spec/generators`?

kinduff:
Able to reproduce with PostgreSQL and Ruby 3.0.1

duleorlovic:
Also got strange string column
```
    t.string "{:null=>false}"
```
in my `db/schema.rb`.
When I use this fix
```
# Gemfile
gem 'paper_trail', github: 'HackerIntro/paper_trail', branch: 'migration-item-type'
```
than it works fine for Ruby 3.0.2 and postgresql

Issue #1316, opened 2021-06-09
Add a last_version association to help against N+1


Hugo-Hache:
### Problem description

Solve the N+1 when serializing attributes of the last version is not trivial.

```ruby
class Widget < ActiveRecord::Base
  has_paper_trail
end

class WidgetsController < ApplicationController
  def index
    @widgets = Widget.all
    render json: @widgets.map { |w| {id: w.id, last_editor: w.versions.last.whodunnit} } # Triggers N+1
  end
end
```

### Proposed solution

Add a `has_one :last_version` relation in `model_config` along `versions` :
```ruby
def define_has_one_last_version(options)
  options = ensure_versions_option_is_hash(options)
  check_version_class_name(options)
  check_versions_association_name(options)
  scope = -> { order(created_at: :desc) }
  @model_class.has_one(
    @model_class.versions_association_name,
    scope,
    class_name: @model_class.version_class_name,
    as: :item,
    **options[:versions].except(:name, :scope)
  )
end
```

On our example we can then simply include last_version to fix the N+1:
```ruby
  def index
    @widgets = Widget.all.includes(:last_version)
    render json: @widgets.map { |w| {id: w.id, last_editor: w.last_version.whodunnit} } # N+1 solved
  end
```

### Alternative solution

Introduce a README section "Solving N+1" explaining how a `has_one` can help to solve the issue.

tlynam:
Thank you for opening this issue!  It's an interesting idea to define this association.  I feel like I'm missing something.  How does having a `has_one` association prevent n+1 queries?  I would think we would need to use `includes` or use multiple queries?

Something like using `includes`?
`@widgets = Widget.includes(:versions).all`

Or use multiple queries like:
`PaperTrail::Version.where(item_id: @widgets.map(&:id), item_type: "Widget").pluck(:whodunnit)`

Hugo-Hache:
Exactly this has_one association is only a help to solve the N+1 with the `includes` method. You can spot the difference in the first line of the index method in the examples of the issue description.  

Without last_version
```
  def index
    @widgets = Widget.all
    render json: @widgets.map { |w| {id: w.id, last_editor: w.versions.last.whodunnit} } # Triggers N+1
  end
```

With last_version
```
  def index
    @widgets = Widget.all.includes(:last_version)
    render json: @widgets.map { |w| {id: w.id, last_editor: w.last_version.whodunnit} } # N+1 solved
  end
```

[...]

PR #1330, opened 2021-07-05
Use an app specific name for abstract version class to avoid autoloading conflict

### LINKED ISSUE #1320: Custom version classes documentation does not work with Rails eager autoloading
## Description

Current README paragraph [Custom version classes](https://github.com/paper-trail-gem/paper_trail#6a-custom-version-classes) does not eager load properly with Rails current usage of Zeitwerk. It produces the following log when starting the server in production mode 

```
Zeitwerk@rails.main: file [REDACTED]/app/models/paper_trail/version.rb is ignored because PaperTrail::Version is already defined
```

We detected this issue when method defined in this `app/models/paper_trail/version.rb` worked in development but not in production.
 
## Suggestion 

I envision two ways to solve this:

1. Set `abstract_class` on `PaperTrail::Version` in an initializer (however I'm not sure how to do it properly with autoloading)
2. Suggest a model with a different name like `ApplicationVersion` to inherit in your classes
```ruby
class PostVersion < ApplicationVersion
  # custom behaviour, e.g:
  self.table_name = :post_versions
end

# app/models/application_version.rb
class ApplicationVersion < ActiveRecord::Base
  self.abstract_class = true
end
```

## Script 

The following file should be put in `my_models/paper_trail` folder beside the script: 
```ruby
module PaperTrail
    self.abstract_class = true
  end
end
```

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  ruby "3.0.1"
  source "https://rubygems.org"
  gem "activerecord", "6.1.3.2"
  gem "paper_trail", "12.0.0", require: false
  gem "sqlite3", "1.4.2"
end

require "active_record"
require "minitest/autorun"
require "logger"

    t.timestamps null: false
  end

  create_table :user_versions do |t|
    t.string :event, null: false
    t.string :whodunnit
    t.datetime :created_at
  end
end
require "paper_trail"

require "zeitwerk"
loader = Zeitwerk::Loader.new
loader.push_dir('my_models')
loader.setup
loader.log!
loader.eager_load

puts "Eager loaded"

class UserVersion < PaperTrail::Version
  self.table_name = :user_versions
  self.sequence_name = :user_versions_id_seq
end

  has_paper_trail versions: {
    class_name: 'UserVersion'
  }
end

  def test_1
    assert PaperTrail::Version.abstract_class
  end
end

```

Hugo-Hache:

Check the following boxes:

- [x] Wrote [good commit messages](http://chris.beams.io/posts/git-commit/).
- [ ] Added tests.

Fix #1320

### DIFF
--- a/README.md
+++ b/README.md
@@ -1216,17 +1216,20 @@ class PostVersion < PaperTrail::Version
 end
 ```
 
-If you only use custom version classes and don't have a `versions` table, you
-must let ActiveRecord know that the `PaperTrail::Version` class is an
-`abstract_class`.
+If you only use custom version classes and don't have a `versions` table, you must
+let ActiveRecord know that your base version class (eg. `ApplicationVersion` below)
+class is an `abstract_class`.
 
 ```ruby
-# app/models/paper_trail/version.rb
-module PaperTrail
-  class Version < ActiveRecord::Base
-    include PaperTrail::VersionConcern
-    self.abstract_class = true
-  end
+# app/models/application_version.rb
+class ApplicationVersion < ActiveRecord::Base
+  include PaperTrail::VersionConcern
+  self.abstract_class = true
+end
+
+class PostVersion < ApplicationVersion
+  self.table_name = :post_versions
+  self.sequence_name = :post_versions_id_seq
 end
 ```
 


jaredbeck (changes requested):
The paragraph above the proposed change should refer to `ApplicationRecord` instead, right?

```
If you only use custom version classes and don't have a `versions` table, you
must let ActiveRecord know that the `PaperTrail::Version` class is an
`abstract_class`.
```

Otherwise, it makes sense to me.

Hugo-Hache:
> The paragraph above the proposed change should refer to the proposed `ApplicationVersion` instead, right?

The current paragraph is still valid, what we want to achieve is set the property abstract_class to true. Maybe I could append to it something like that ?

`A simple way to do that for all your custom versions classes is to create a parent ApplicationVersion class.`

jaredbeck:
> A simple way to do that for all your custom versions classes is to create a parent ApplicationVersion class.

I like this sentence, let's add it. Also, how about amending the paragraph above with something like the following?

```diff
- you must let ActiveRecord know that the `PaperTrail::Version` 
- class is an `abstract_class`.
+ you must let ActiveRecord know that your base version 
+ class (eg. `ApplicationVersion` below) is an `abstract_class`.
```

tlynam:
Hi @jaredbeck, how about I make these changes and we can merge this?

jaredbeck:
Thanks Todd, go for it

Hugo-Hache:
Mea culpa, took me an infinite time to handle this nice suggestion. Just updated it, let me know if it's ok for you.

tlynam (approved):
Looks great, thank you so much!

Hugo-Hache:
Thanks @tlynam 👍 I hope I will be more reactive if you take interest in #1316

PR #1333, opened 2021-07-23
Improve reification of STI classes

### LINKED ISSUE #1332: Improve fetching reified class names for STI models

We use PaperTrail on various models. One of them is a legacy model, where we use (abuse?) single table inheritance to separate callbacks, validations, and so forth into separate submodels. Let's call that model `Setting`, which has a column `name`. Name contains values such as `user_setting`, `team_setting`, and so on. 

We follow an approach similar to [the one outlined here](https://stackoverflow.com/a/23295546/2553104) to use STI using the names column to map underscored names (`user_setting`) to the actual class (`UserSetting`). Note that usually, STI columns  contain carmelized names.   

When destroying such a model and reifying it, the existing mechanism does not account for our  ̶a̶w̶e̶f̶u̶l̶ ̶h̶a̶c̶k̶ workaround, and we get an error: 

```ruby
setting = Setting.create(name: 'user_setting')
setting.destroy
setting.versions.last.reify # <- This fails with the wrong constant name 'user_setting'
```

The underlying issue is the implementation of `PaperTrails::Reifier`: 

```ruby
def version_reification_class(version, attrs)
  inheritance_column_name = version.item_type.constantize.inheritance_column
  inher_col_value = attrs[inheritance_column_name]
  class_name = inher_col_value.blank? ? version.item_type : inher_col_value
  class_name.constantize # <- This does not account for non-standard class names, such as underscored instead of camelized
end
```

Instead of simply using `class_name.constantize` I suggest to take advantage of the mechanism Rails provides - calling `find_sti_class`: 

```ruby
def version_reification_class(version, attrs)
  clazz = version.item_type.constantize

  inheritance_column_name = clazz.inheritance_column
  inher_col_value = attrs[inheritance_column_name]
  return clazz if inher_col_value.blank?

  clazz.send(:find_sti_class, inher_col_value)
end
```

This allows users who use some non-standard form of STI to use PaperTrail, while keeping the existing functionality intact.

I realize that this is a _very_ specific problem to have. The obvious alternative is to just not have that problem and properly use STI. Unfortunately, we don't really have that option :shrug: 

I'm of course perfectly fine with just monkey-patching `PaperTrail::Reifier` to suit our needs. However, I can't see any obvious downside to using `find_sti_class`, and this change makes reifying more flexible. 

Let me know what you think :)

hschne:
Checklist

- [X] Added tests.
  
This fixes #1332. 

As discussed, two commits were created, one with failing test cases and one to fix them. Some details about this solution: 

I opted to add a new test model, rather than an modify an existing one (e.g. `Animal`) to better represent the specific edge case described in the issue. 

Futhermore, as `sti_class_for` was only added recently (Rails 6.1), reification falls back to calling the private method `find_sti_class`. I could not find another way to keep tests for older versions of Rails passing. 

I left the commits unsquashed for transparency, feel free to squash when merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 - [#1285](https://github.com/paper-trail-gem/paper_trail/pull/1285) -
   For ActiveRecord >= 6.0, the `touch` callback will no longer create a new
   `Version` for skipped or ignored attributes.
+- [#1333](https://github.com/paper-trail-gem/paper_trail/pull/1333) -
+  Improve reification of STI models that use `find_sti_class`/`sti_class_for`
+  to customize single table inheritance.
 
 ## 12.0.0 (2021-03-29)
 
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -109,14 +109,22 @@ module PaperTrail
       end
 
       # Given a `version`, return the class to reify. This method supports
-      # Single Table Inheritance (STI) with custom inheritance columns.
+      # Single Table Inheritance (STI) with custom inheritance columns and
+      # custom inheritance column values.
       #
       # For example, imagine a `version` whose `item_type` is "Animal". The
       # `animals` table is an STI table (it has cats and dogs) and it has a
       # custom inheritance column, `species`. If `attrs["species"]` is "Dog",
       # this method returns the constant `Dog`. If `attrs["species"]` is blank,
-      # this method returns the constant `Animal`. You can see this particular
-      # example in action in `spec/models/animal_spec.rb`.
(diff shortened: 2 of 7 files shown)


jaredbeck (changes requested):
This is an interesting STI feature that I had not heard of before. As of 6.1, it is a feature of Rails' public API, so I'm happy to support it.

Your [description of the issue](https://github.com/paper-trail-gem/paper_trail/issues/1332#issue-949506346) was excellent. Please boil that down into a few paragraphs, and add them to the comments above `#version_reification_class`.

hschne:
I updated the comments, let me know what you think :)

jaredbeck (approved):

jaredbeck:
Thanks Hans, nice work!

jaredbeck:
Released in 12.1.0

PR #1338, opened 2021-09-05
Support Psych version 4

### LINKED ISSUE #1335: Support for Psych version 4
I think Psych 4  uses `safe_load` by default

https://github.com/ruby/psych/pull/487

This seems to cause a problem when retrieving previous versions with papertrail
Check the following boxes:

- [X] This is not a usage question, this is a bug report
- [X] This bug can be reproduced with the script I provide below
- [X] This bug can be reproduced in the latest release of the `paper_trail` gem

```ruby

# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  ruby "2.7.4"
  source "https://rubygems.org"
  gem "activerecord", "6.1.0"
  gem "minitest", "5.11.3"
  #gem "psych", ">=3", "<4"
  gem "psych", ">=4"
  gem "paper_trail", ">=12", require: false
  gem "sqlite3",  "1.4.2"
end

require "active_record"
require "minitest/autorun"
require "logger"

    t.timestamps null: false
  end

  create_table :versions do |t|
    t.string :event, null: false
    t.string :whodunnit
    t.datetime :created_at
  end
end
require "paper_trail"

  has_paper_trail
end

  def test_1
    assert_nothing_raised {
      u = User.create(first_name: "Jane")
      u.first_name = "John"
      u.save
      u.paper_trail.previous_version
    }
  end
end


```

t27duck:
Fixes #1335

Psych's `.load` method uses `.safe_load` by default which is not compatible with papertail needs to load.

This implements the suggested fix in the issue which is a similar fix Rails uses throughout its codebase when it needs to load YAML content.

The test case outlined in the issue passes with this change (after I pointed the paperclip gem to the forked branch).

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1338](https://github.com/paper-trail-gem/paper_trail/pull/1338) -
+  Support Psych version 4
 
 ## 12.1.0 (2021-08-30)
 
--- a/lib/paper_trail/serializers/yaml.rb
+++ b/lib/paper_trail/serializers/yaml.rb
@@ -9,7 +9,7 @@ module PaperTrail
       extend self # makes all instance methods become module methods as well
 
       def load(string)
-        ::YAML.load string
+        ::YAML.respond_to?(:unsafe_load) ? ::YAML.unsafe_load(string) : ::YAML.load(string)
       end
 
       # @param object (Hash | HashWithIndifferentAccess) - Coming from
--- a/spec/paper_trail/serializers/yaml_spec.rb
+++ b/spec/paper_trail/serializers/yaml_spec.rb
@@ -22,6 +22,19 @@ module PaperTrail
           expect(described_class.load(hash.to_yaml)).to eq(hash)
           expect(described_class.load(array.to_yaml)).to eq(array)
         end
+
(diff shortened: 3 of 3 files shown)


gurgelrenan:
Hi @t27duck , thanks for your P.R. Looks good to me. 

@jaredbeck do you think that they should add some additional tests? I can't figure out a good way to test this cenario.

jaredbeck:
> @jaredbeck do you think that they should add some additional tests? I can't figure out a good way to test this cenario. Perhaps some mock in `spec/paper_trail/serializers/yaml_spec.rb`?

Yes, a mock should work, e.g. 

```
allow(YAML).to receive(:unsafe_load)
serializer.load
expect(YAML).to have_received(:unsafe_load)
```

gurgelrenan:
@t27duck Could you add the required tests?

t27duck:
@gurgelrenan - Spec added.

Rspec doesn't allow stubbing methods that don't exist on the object that's being stubbed, so based on whichever version of Psych is running with the tests, I had to have the spec check what `YAML` does implement and set stubs and expectations based on that.

jaredbeck:
Thanks Tony. This is good enough. 

> Rspec doesn't allow stubbing methods that don't exist on the object that's being stubbed, ..

For future reference, I wonder if `without_partial_double_verification` would help ..? I haven't tried.

gurgelrenan:
Thanks @t27duck 😄

PR #1350, opened 2021-10-22
Add specific instructions for MySQL migration

koffeinfrei:
The currently generated migration mentions that MySQL users should explicitly enable "fractional seconds precision", but doesn't provide specific steps to enable this. The user has to click the links to figure out what to do.
This change adds the specific migration statement that can be used.

Check the following boxes:

- [x] ~~Added tests.~~
- [x] ~~Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new 
  code introduces user-observable changes.~~

### DIFF
--- a/lib/generators/paper_trail/install/templates/create_versions.rb.erb
+++ b/lib/generators/paper_trail/install/templates/create_versions.rb.erb
@@ -28,7 +28,9 @@ class CreateVersions < ActiveRecord::Migration<%= migration_version %>
       # MySQL users should also upgrade to at least rails 4.2, which is the first
       # version of ActiveRecord with support for fractional seconds in MySQL.
       # (https://github.com/rails/rails/pull/14359)
-      #
+      # 
+      # MySQL users should use the following line for `created_at`
+      # t.datetime :created_at, limit: 6                                        
       t.datetime :created_at
     end
     add_index :versions, %i(item_type item_id)


lib/generators/paper_trail/install/templates/create_versions.rb.erb
```diff
@@ -28,7 +28,9 @@ class CreateVersions < ActiveRecord::Migration<%= migration_version %>
       # MySQL users should also upgrade to at least rails 4.2, which is the first
       # version of ActiveRecord with support for fractional seconds in MySQL.
       # (https://github.com/rails/rails/pull/14359)
-      #
+      # 
+      # MySQL users should use the following line for `created_at`
+      # t.datetime :created_at, limit: 6
```
tlynam, line 33: Hi @jaredbeck, do you use MySQL to verify this?

jaredbeck: `limit` is the correct option, IIRC. I don't remember what the limit should be though. I think `6` will give microseconds precision. That's probably fine. In rare cases, some apps may choose a different number, but it's probably fine if we recommend `6`.

> Hi @jaredbeck, do you use MySQL to verify this?

Yes, the CI runs an actual MySQL instance. You can also run MySQL locally if you want.

jaredbeck (approved):

tlynam:
Sounds good @jaredbeck, thank you!

PR #1365, opened 2022-01-15
Rails 7.0 Compatibility

dfurber:
Makes PaperTrail work with Rails 7. The specs now pass, with underlying issues addressed and comments added. You'll have to add it to the GitHub Actions. 

Most of this PR is actually just adding specs unrelated to Rails 7 to get the coverage threshold up over the line.


lib/paper_trail/version_concern.rb
```diff
@@ -19,6 +19,20 @@ module VersionConcern
...
+      # effects result:
+      # 1. If the reifier uses version.item to reify, then it also modifies the original item
+      #    object. If you just deleted that item and it is frozen, then you'll get an error that
+      #    it can't modify a frozen object.
+      # 2. If you delete an item and then get the "last" version's item, you'll get the item you
+      #    just deleted rather than nil. That is because the belongs_to association's stale target
+      #    checking now thinks that the association hasn't gone stale and thus returns the
+      #    original item object.
+      if PaperTrail::RAILS_GTE_7_0
+        after_find { association(:item).instance_variable_set(:@stale_state, nil) }
```
pedrocarmona, line 34: why not use the `inverse_of: nil` discussed in the other PR? it uses the public api of rails to fix it...

dfurber: Public api preferred of course. But doesn't work. At least, I just tried removing the `after_find` and adding `inverse_of: nil` to the `belongs_to :item`, and the specs failed that led me to this hack.

dfurber: Which may well indicate that this new Rails feature is broken, and until we can fix it...

mathieujobin (approved):
:100:

mathieujobin:
Thank you very much for this, I am using this branch now and its working well.

lib/paper_trail.rb
```diff
@@ -26,6 +26,8 @@ module PaperTrail
     named created_at.
   EOS
 
+  RAILS_GTE_7_0 = ::ActiveRecord.gem_version >= ::Gem::Version.new("7.0.0")
```
jaredbeck, line 29: Nice use of `Gem::Version`

lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
```diff
@@ -32,6 +32,10 @@ def deserialize(attr, val)
         if defined_enums[attr] && val.is_a?(::String)
           # Because PT 4 used to save the string version of enums to `object_changes`
           val
+        elsif PaperTrail::RAILS_GTE_7_0 && val.is_a?(ActiveRecord::Type::Time::Value)
+          # Because Rails 7 time attribute throws a delegation error when you deserialize
+          # it with the factory.
```
jaredbeck, line 37: Should we add a link to https://github.com/rails/rails/issues/43966 to this comment?

lib/paper_trail/version_concern.rb
jaredbeck: This `instance_variable_set` makes me pretty nervous. Pedro, if you're aware of a public-api alternative, can you please push a branch? If you can open a PR in David's fork, against this branch, that would be ideal. Thanks.

lib/paper_trail/version_number.rb
```diff
@@ -7,8 +7,8 @@ module PaperTrail
   # because of this confusion, but it's not worth the breaking change.
   # People are encouraged to use `PaperTrail.gem_version` instead.
   module VERSION
-    MAJOR = 12
-    MINOR = 1
+    MAJOR = 13
+    MINOR = 0
```
jaredbeck, line 11: Please don't change version number. First, I like to change it during release process. Second, the next version may be 12.2, as there are no breaking changes yet, AFAIK.

jaredbeck (commented):
Hi @dfurber long time no see! Thanks for the PR. I love to see all the new tests, and the great comments.

The only thing I'm concerned about is the `instance_variable_set`. I'm hoping @pedrocarmona's research will lead to solution that uses public API.

lib/paper_trail/version_number.rb
jaredbeck: Oh, that reminds me, can you please add an entry to the changelog under Unreleased -> Added?

lib/paper_trail/version_concern.rb
dfurber: `inverse_of: nil` didn't work but `inverse_of: false` did. Coming soon!

lib/paper_trail/version_number.rb
dfurber: Done!

lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
dfurber: Done!

lib/paper_trail.rb
dfurber: Thanks. Just trying to stick to the gem's dialect - got it from `ModelConfig`.

lib/paper_trail/version_concern.rb
dfurber: And done.

dfurber: I guess now it's a question of how to get back those hours between setting `inverse_of` to `nil` (first thing I tried because it was the only obvious change from 6.1.4 to 7.0), and trying `false` instead.

jaredbeck (approved):
LGTM, thanks David!

@pedrocarmona , any concerns?

lib/paper_trail/version_concern.rb
pedrocarmona: Nice!

pedrocarmona:
> LGTM, thanks David!
> 
> @pedrocarmona , any concerns?

LGTM

Thank you @dfurber

pedrocarmona (approved):

mathieujobin:
oups, I think we are missing the update to the .github/workflows "script"

baarkerlounger:
Now that this is merged would it be possible to cut a new release?

t27duck:
> Now that this is merged would it be possible to cut a new release?

12.2.0 is that release. Been out for a couple weeks.

baarkerlounger:
@t27duck thanks, missed that as Github releases still shows 12.0.0.

PR #1366, opened 2022-01-18
Fix `create_versions` migration to support Ruby3

joeytepp:
The `create_versions` migration created by doing `bundle exec rails generate paper_trail:install` is not compatible with Ruby3. This PR updates the migration so that it will be compatible

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1366](https://github.com/paper-trail-gem/paper_trail/pull/1366) -
+  Fixed a bug where the `create_versions` migration lead to a broken `db/schema.rb` for Ruby 3
 
 ### Dependencies
 
--- a/lib/generators/paper_trail/install/install_generator.rb
+++ b/lib/generators/paper_trail/install/install_generator.rb
@@ -41,9 +41,9 @@ module PaperTrail
     # See https://github.com/paper-trail-gem/paper_trail/issues/651
     def item_type_options
       if mysql?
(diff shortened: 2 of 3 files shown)


batter:
Does this remove compatibility with Ruby 2 in any capacity?

joeytepp:
@batter No, keyword arguments were introduced in [Ruby 2.0](https://thoughtbot.com/blog/ruby-2-keyword-arguments)

mathieujobin:
I am curious why the tests were not failing on ruby 3 before?

joeytepp:
@mathieujobin they weren't failing because the migration still succeeds, however in the `schema.rb` file we end up with
```ruby
    t.string "{:null=>false, :limit=>191}"
```
instead of
```ruby
    t.string :item_type, null: false, limit: 191
```
This is due to the fact that Ruby 3 interprets the hash `{ null: false, limit: 191 }` as a second argument instead of keyword arguments. So it needs to be explicitly passed as keyword arguments.

mathieujobin (approved):

jaredbeck:
Thanks for the contribution. Please add an entry to the changelog under Unreleased -> Fixed.

joeytepp:
@jaredbeck updated the changelog 👍

jaredbeck:
Thanks, Joey

tinacious:
In existing projects, what do we need to do to get these changes applied retroactively? I've updated the gem to 12.3.0 but am still seeing the mangled schema.rb file after running migrations. `bundle exec rails g paper_trail:install` fails as there's already a version created, and `--force` doesn't do anything (this is expected as the table exists). I'm on Rails 6.1.0 and Ruby ruby 3.0.2. Any help would be appreciated, thanks! 🙏

PR #1371, opened 2022-01-31
Added in_after_callback argument to save_with_version

denisahearn:
My application calls `PaperTrail::RecordTrail#save_with_version` in an `after` callback. The version record that this method creates does not record the current changes on the object being tracked. I found the cause in the `PaperTrail::Events::Base#load_changes_in_latest_version` method, which only uses `@record.saved_changes` when `@in_after_callback` is true. Unfortunately I don't currently see a way to make `@in_after_callback == true` when using `save_with_version`.

This pull request adds an `in_after_callback` argument to the `PaperTrail::RecordTrail#save_with_version` method, to allow the caller to indicate if this method is being called during an `after` callback. The argument defaults to `false` when not provided.

Check the following boxes:

- [x] ~~Squashed related commits together.~~
- [x] Added tests.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -50,6 +50,7 @@ Security/YAMLLoad:
     - 'spec/models/book_spec.rb'
     - 'spec/models/gadget_spec.rb'
     - 'spec/models/no_object_spec.rb'
+    - 'spec/models/not_on_update_spec.rb'
     - 'spec/models/person_spec.rb'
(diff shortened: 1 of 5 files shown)


jaredbeck:
Hi Denis, thanks for the contribution. I'm not sure how many people will use this new feature, but there aren't really any good alternatives that I can think of. It's a good PR too :) You've ticked all the boxes.

Issue #1373, opened 2022-02-17
Add instruction to use uuid type for `item_id` in migration.


Nuzair46:
Thank you for your contribution!

Due to limited volunteers, issues that do not follow this template will be
closed without comment.

**Is your feature suggestion related to a problem? Please describe.**

* By default, create_versions migration use `bigint` type for `item_id`. This causes issues with tables using `uuid` where the `item_id` will be saved as `0` sometimes.

**Describe the solution you'd like to build**

* Add instruction in the readme to set the type `uuid` for `item_id` in the migration.

PR #1374, opened 2022-02-17
Added ability to set uuid as item_id type while generating migration

Nuzair46:
Added option `--uuid` when generating new migration. This can be used to set the type on `item_id` column to uuid.

Use case: 

* If a user uses paper_trail on a database that uses `uuid` as primary key, there is a chance where `item_id` becomes `0` causing the versioning to not work. Changing the type of `item_id` from `bigint` to `string` fixes this issue.
* This PR makes it possible for the user to use `uuid` while generating the migration.

```
bin/rails generate paper_trail:install [--uuid]
```

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,6 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ## Unreleased
 
-- Added option `--uuid` when generating new migration. This can be used to set the type of item_id column to uuid for use with paper_trail on a database that uses uuid as primary key.
-
 ### Breaking Changes
 
 - None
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1152,176 +1152,177 @@ candidates.
...
-  - Updated deprecation warnings for `Model.paper_trail_on` and
-    `Model.paper_trail_off` to have display correct version number the methods
-    will be removed (`4.0`)
+- [#479](https://github.com/paper-trail-gem/paper_trail/issues/479) - Deprecated
+  `originator` method in favor of `paper_trail_originator` Deprecation warning
+  informs users that the `originator` of the methods will be removed in
+  version `4.0`. (Backported from v4)
+- Updated deprecation warnings for `Model.paper_trail_on` and
+  `Model.paper_trail_off` to have display correct version number the methods
+  will be removed (`4.0`)
```
jaredbeck, line 1161: Please do not change unrelated parts of the CHANGELOG. I'm guessing your IDE reformatted it automatically, but I don't want to read through 300 lines of diffs to confirm that.

spec/generators/paper_trail/install_generator_spec.rb
```diff
@@ -102,4 +102,27 @@
...
     end
   end
+
+  describe "`--uuid` option set to `true`" do
+    before do
+      prepare_destination
+      run_generator %w[--uuid]
+    end
+
+    expected_item_id_type = "string"
```
jaredbeck, line 112: Thanks for writing a test, but it's unusual IME to define a local variable in a `describe` block. Please move this variable into the `it` block.

jaredbeck (changes requested):
Thanks Nuzair! Generally this looks good, just a few minor comments.

Also, please update the install documentation in the README.

Nuzair46:
@jaredbeck , Suggested changes are made. Please review. Thanks.

closes #1373

jaredbeck:
Thanks for the contribution!

PR #1376, opened 2022-03-01
Try to create a version record when associated object is touched

### LINKED ISSUE #1339: The versions doesn't create when associated object deleted since v12.1.0.
Check the following boxes:

Our application have a model that have an association with `touch: true` option. Before v12.1.0, when deleting the model, `touch` event triggered and created an version of an  association. But v12.1.0, association's version isn't created.  

Reproduction script is following. 

```ruby 
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "activerecord", "6.0.4.1"
  gem "paper_trail", "12.1.0", require: false
  # gem "paper_trail", "12.0.0", require: false
  gem "sqlite3"
  gem "minitest"
end

require "active_record"
require "minitest/autorun"
require "logger"

    t.string :name, null: false
    t.timestamps
  end

  create_table :avatars, force: true do |t|
    t.string :name, null: false
    t.belongs_to :user, null: false
    t.timestamps
  end

  create_table :versions do |t|
    t.string :event, null: false
    t.string :whodunnit
    t.datetime :created_at
  end
end

require "paper_trail"

  has_paper_trail
end

class Avatar < ActiveRecord::Base
  has_paper_trail

  belongs_to :user, touch: true
end

  # This test passed with v12.0.0, but doesn't pass with v12.1.0.
  def test_1
    user = User.create(name: "Jane")
    avatar = Avatar.create(name: "Dummy1", user: user)

    assert_difference(-> { PaperTrail::Version.count }, +2) {
      Avatar.destroy_all
    }
  end
end
```

Is this intentional change? 

Thanks.

y-yagi:
Before #1285, a version record was created when associated object is touched.
But because of Rails don't track implicit touch mutation[1], now PaperTail don't create a version record was created when associated object is touched.

This patch try to keep creating a version record in that case by doing checking an event and changed value.

Fixes #1339.

[1]: https://github.com/rails/rails/commit/dcb825902d79d0f6baba956f7c6ec5767611353e

---

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1376](https://github.com/paper-trail-gem/paper_trail/pull/1376) -
+  Create a version record when associated object was changed the same as
+  PaperTrail < v12.1.0.
 
 ## 12.2.0 (2022-01-21)
 
--- a/lib/paper_trail/events/update.rb
+++ b/lib/paper_trail/events/update.rb
@@ -40,6 +40,20 @@ module PaperTrail
         merge_metadata_into(data)
       end
 
+      # If it is a touch event, and changed are empty, it is assumed to be
+      # implicit `touch` mutation, and will a version is created.
+      #
+      # See https://github.com/rails/rails/commit/dcb825902d79d0f6baba956f7c6ec5767611353e
+      #
+      # @api private
+      def changed_notably?
+        if @is_touch && changes_in_latest_version.empty?
+          true
+        else
+          super
+        end
+      end
+
       private
 
       # @api private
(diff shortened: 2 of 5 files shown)


jaredbeck:
LGTM. WDYT @westonganger @quainjn @tlynam ?

westonganger:
Sure looks good I guess, not sure I should be considered an expert here though.

jaredbeck:
Thanks @y-yagi !

y-yagi:
Thanks for your review!

KevinBongart:
Thank @y-yagi for the patch! I noticed the same issue, identified the root cause and found this PR before I attempted to fix it :D

@jaredbeck Any ETA for when this patch will make its way into a new release?

steve21168:
I'm actually looking to do the opposite of this.  I don't want a version record when the association is touched.  In my case it's creating extraneous version records showing no object changes.

To fix this, I added `has_paper_trail ignore: [:updated_at]`.  However this change doesn't take into account `ignore` or `skip`.  So now when associated objects are touched there's no way to avoid these version records.

Any ideas to work around this?

PR #1381, opened 2022-03-31
Fix: n+1 when loading a changeset

HashNotAdam:
When producing a changeset, the item is loaded from the database in
order to check it's class. This came about because of issues that
emerged when supporting STI.

Unfortunately, the result is that the record is retrieved once for every
version despite the fact that the item class will be the same.

The documented reason for moving away from using item_type to know the
class was that it returns the wrong class when working with STI but
there is an item_subtype which returns the correct class.

Check the following boxes:

- [N/A] Squashed related commits together.
- [N/A] Added tests.
- [N/A] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new 

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -320,18 +320,10 @@ module PaperTrail
       # and appears to be responsible for custom attribute serializers. For an
       # example of a custom attribute serializer, see
       # `Person::TimeZoneSerializer` in the test suite.
-      #
-      # Is `item.class` good enough? Does it handle `inheritance_column`
-      # as well as `Reifier#version_reification_class`? We were using
-      # `item_type.constantize`, but that is problematic when the STI parent
-      # is not versioned. (See `Vehicle` and `Car` in the test suite).
-      #
-      # Note: `item` returns nil if `event` is "destroy".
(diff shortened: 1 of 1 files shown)


jaredbeck:
Thanks, Adam. This is something I'd like to investigate further.

Loading the changeset for multiple versions is a common use-case, so this could be a common performance performance problem.

## 1. Non-breaking change?

`item.class` and `item_subtype.constantize` are always the same, right?

Assuming correct library usage, can you think of any situations where they wouldn't be the same?

## 2. Missing tests?

Looking at our SimpleCov report, It seems we are missing a test for when `item_subtype` is _absent_? That seems important because `item_subtype` is an optional column. The final term of our new ternary would be uncovered.

```
type = respond_to?(:item_subtype) ? item_subtype : item_type
```

## 3. Potential `LoadError`

Finally, how should we handle the potential `LoadError` from `type.constantize`? 

This error could happen when an STI child's source files are deleted, and the programmer neglects to update the db records referenced by that `item_subtype`. That'd be the developer's fault, not the library, of course, but we might want to rescue the `LoadError` and raise something more helpful? Or, maybe just print a warning and re-raise?

PR #1396, opened 2022-08-01
Fix invalid version created_at value when using update_columns

### LINKED ISSUE #1395: paper_trail.update_columns is setting an incorrect Version#created_at value
- `PaperTrail::RecordTrail#update_columns` is setting `Version#created_at` as the `recod.updated_at` value.
- since `update_columns` does not update timestamps - the previous update date is used for the `Version#created_at` instead of the actual update date (current)

I think this was actually addressed in the original PR, but the [adjustment that fixed the value got removed during the code review](https://github.com/paper-trail-gem/paper_trail/pull/1037#discussion_r162418391)

------

------
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  ruby "3.0.2"
  source "https://rubygems.org"
  gem "activerecord", "6.1.4.1"
  gem "minitest", "5.11.3"
  gem "paper_trail", "12.3.0", require: false
  gem "sqlite3", "1.4.2"
end

require "active_record"
require "minitest/autorun"
require "logger"

    t.timestamps null: false
  end

  create_table :versions do |t|
    t.string :event, null: false
    t.string :whodunnit
    t.datetime :created_at
  end
end
require "paper_trail"

  has_paper_trail
end

class UpdateColumnsCreatedAtTest < ActiveSupport::TestCase
  def test_version_created_at
    freeze_time
    user = User.create!(first_name: 'John', created_at: '2015-01-01 14:00', updated_at: '2015-01-01 15:00')

    user.paper_trail.update_columns(first_name: 'Jack')
    assert_equal Time.now, user.versions.last.created_at
  end
end


```

iiwo:
Use the default(AR generated) timestamp for the `Version#created_at` value when using `PaperTrail::RecordTrail#update_columns`.

`PaperTrail::RecordTrail#update_columns` was previously setting `Version#created_at` as the recod `updated_at` value.
Since `update_columns` does not update timestamps - the previous update date was being used for the `Version#created_at` instead of the actual update date.

Fixes #1395

---------

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -148,11 +148,14 @@ module PaperTrail
     # paper_trail-association_tracking
     def record_update_columns(changes)
       return unless enabled?
-      event = Events::Update.new(@record, false, false, changes)
+      data = Events::Update.new(@record, false, false, changes).data
+
+      # use AR default timestamp value instead of record.updated_at value
+      data.delete(:created_at)
 
       # Merge data from `Event` with data from PT-AT. We no longer use
       # `data_for_update_columns` but PT-AT still does.
-      data = event.data.merge(data_for_update_columns)
+      data.merge!(data_for_update_columns)
 
       versions_assoc = @record.send(@record.class.versions_association_name)
       version = versions_assoc.create(data)
--- a/spec/models/widget_spec.rb
+++ b/spec/models/widget_spec.rb
@@ -955,15 +955,19 @@ RSpec.describe Widget, type: :model, versioning: true do
   end
 
   describe ".paper_trail.update_columns", versioning: true do
-    let(:widget) { described_class.create! name: "Bob", an_integer: 1 }
-
     it "creates a version record" do
-      widget = described_class.create
+      widget = described_class.create(updated_at: "2015-01-01 15:00")
       expect(widget.versions.count).to eq(1)
-      widget.paper_trail.update_columns(name: "Bugle")
+      now = nil
+      freeze_time do
+        now = Time.zone.now
+        widget.paper_trail.update_columns(name: "Bugle")
+      end
+
       expect(widget.versions.count).to eq(2)
       expect(widget.versions.last.event).to(eq("update"))
       expect(widget.versions.last.changeset[:name]).to eq([nil, "Bugle"])
+      expect(widget.versions.last.created_at.to_i).to eq(now.to_i)
     end
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1395](https://github.com/paper-trail-gem/paper_trail/issues/1395) Fix invalid
+  `Version#created_at` value when using `PaperTrail::RecordTrail#update_columns`
 
 ## 12.3.0 (2022-03-13)
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -10,6 +10,7 @@ end
 SimpleCov.minimum_coverage(ENV["DB"] == "postgres" ? 97.3 : 92.4)
 
 require "byebug"
+require "active_support/testing/time_helpers"
 require_relative "support/pt_arel_helpers"
 
 unless ENV["BUNDLE_GEMFILE"].match?(/rails_\d\.\d\.gemfile/)
@@ -41,6 +42,7 @@ RSpec.configure do |config|
   end
   config.order = :random
   config.include PTArelHelpers
+  config.include ActiveSupport::Testing::TimeHelpers
   Kernel.srand config.seed
 end
 


marcelotoledo5000 (commented):
👍 Good call!!!

gurgelrenan:
Thanks for your P.R @iiwo . SGTM. I think we can merge. Do you see any problems @jaredbeck ?

lib/paper_trail/record_trail.rb
```diff
@@ -148,11 +148,14 @@ def data_for_update
     # paper_trail-association_tracking
     def record_update_columns(changes)
       return unless enabled?
-      event = Events::Update.new(@record, false, false, changes)
+      data = Events::Update.new(@record, false, false, changes).data
+
+      # use AR default timestamp value instead of record.updated_at value
+      data.delete(:created_at)
```
jaredbeck, line 154: Why remove it here instead of removing the original assignment (`data[:created_at] =`) in `events/update.rb`?

spec/models/widget_spec.rb
```diff
@@ -955,15 +955,19 @@
...
+      now = nil
+      freeze_time do
+        now = Time.zone.now
+        widget.paper_trail.update_columns(name: "Bugle")
+      end
+
       expect(widget.versions.count).to eq(2)
       expect(widget.versions.last.event).to(eq("update"))
       expect(widget.versions.last.changeset[:name]).to eq([nil, "Bugle"])
+      expect(widget.versions.last.created_at.to_i).to eq(now.to_i)
```
jaredbeck, line 970: I'm not familiar with `freeze_time`. Can we do something like `expect {}.to_not(change { widget.versions.last.created_at })` instead? Would that achieve the same testing goal?

lib/paper_trail/record_trail.rb
iiwo: I want to specifically address `update_columns` method do not want to affect `build_version_on_update` (so the "regular" update event flow / core functionality).

Doing this in `Events::Update` would require me to add an additional parameter to the constructor. 

This felt like a cleaner and simpler solution that is just an adjustment for this specific method/scenario which I'm addressing, but I can change it if you think that would be better.

spec/models/widget_spec.rb
iiwo: We want to test that `widget.versions.last.created_at`  will change to `Time.now` (current time) instead of `widget.updated_at` (record last update date) (it will always change)

Alternatively (without freezing time and rounding instead - not perfect) I think we could do:

- option 1 (consistent with existing spec):
```diff
     it "creates a version record" do
       widget = described_class.create(updated_at: "2015-01-01 15:00")
       expect(widget.versions.count).to eq(1)
-      now = nil
-      freeze_time do
-        now = Time.zone.now
-        widget.paper_trail.update_columns(name: "Bugle")
-      end
+      widget.paper_trail.update_columns(name: "Bugle")
 
       expect(widget.versions.count).to eq(2)
       expect(widget.versions.last.event).to(eq("update"))
       expect(widget.versions.last.changeset[:name]).to eq([nil, "Bugle"])
-      expect(widget.versions.last.created_at.to_i).to eq(now.to_i)
+      expect(widget.versions.last.created_at).to be_within(1.second).of(Time.zone.now)
     end
end
```

using the `change` matcher with `widget.versions.last` is a bit complicated because the subject (`versions.last`) of the test will be different for comparison, but I guess we could do something like this:

- option 2 (using `change` matcher and a separate scenario):
```diff
   describe ".paper_trail.update_columns", versioning: true do
     it "creates a version record" do
-      widget = described_class.create(updated_at: "2015-01-01 15:00")
+      widget = described_class.create
       expect(widget.versions.count).to eq(1)
-      now = nil
-      freeze_time do
-        now = Time.zone.now
-        widget.paper_trail.update_columns(name: "Bugle")
-      end
+      widget.paper_trail.update_columns(name: "Bugle")
 
       expect(widget.versions.count).to eq(2)
       expect(widget.versions.last.event).to(eq("update"))
       expect(widget.versions.last.changeset[:name]).to eq([nil, "Bugle"])
-      expect(widget.versions.last.created_at.to_i).to eq(now.to_i)
+    end
+
+    it "uses current time for the version created_at" do
+      widget = described_class.create(updated_at: "2015-01-01 15:00")
+
+      expect { widget.paper_trail.update_columns(name: "Bugle") }.to(
+        change {
+          Time.now.to_f - widget.versions.where(event: "update").last&.created_at.to_f < 1
+        }.to(eq(true))
+      )
     end
   end
```

lib/paper_trail/record_trail.rb
jaredbeck: Makes sense, thanks for the helpful explanation :)

spec/models/widget_spec.rb
jaredbeck: Thanks for the options! I think I'll go with option 2.

jaredbeck:
Merged in https://github.com/paper-trail-gem/paper_trail/pull/1405, preserving your authorship.

Thanks @iiwo for the fix, and thanks @gurgelrenan for reviewing.

PR #1401, opened 2022-09-23
explicitly require active_record

isaacbowen:
Rails 7 (and Rails 6 in `zeitwerk` mode) requires active_record to be explicitly required, before the `::ActiveRecord` module becomes available.

I ran into this when using sidekiq-ent's `sidekiqswarm`, which loads the bundle via `Bundler.require(:default)` *before* booting Rails.

Additional context:
* https://github.com/mperham/sidekiq/issues/4766
* https://github.com/mperham/sidekiq/wiki/Ent-Multi-Process#bundler-preload

Note that I did _not_ write any tests for this; I'm not immediately sure what the right way to test this _is_. :)

---

Check the following boxes:

- [ ] Added tests.
- [x] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes.

### DIFF
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# Rails 7 (and Rails 6 with Zeitwerk) requires explicit loading of AR.
+require "active_record"
+
 # AR does not require all of AS, but PT does. PT uses core_ext like
 # `String#squish`, so we require `active_support/all`. Instead of eagerly
 # loading all of AS here, we could put specific `require`s in only the various


isaacbowen:
To be clear, this is addressing a `uninitialized constant ActiveRecord (NameError)` error in `lib/paper_trail.rb:29`:

https://github.com/paper-trail-gem/paper_trail/blob/6eb5c39e7eb6c39e90b50d40d1e0ee766e14d08c/lib/paper_trail.rb#L29

```
$ sidekiqswarm -q run_default
[swarm] Preloading Bundler groups ["default"]
/Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/paper_trail-13.0.0/lib/paper_trail.rb:29:in `<module:PaperTrail>': uninitialized constant ActiveRecord (NameError)

  RAILS_GTE_7_0 = ::ActiveRecord.gem_version >= ::Gem::Version.new("7.0.0")
                  ^^^^^^^^^^^^^^
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/paper_trail-13.0.0/lib/paper_trail.rb:22:in `<top (required)>'
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/bundler-2.3.9/lib/bundler/runtime.rb:60:in `require'
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/bundler-2.3.9/lib/bundler/runtime.rb:60:in `block (2 levels) in require'
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/bundler-2.3.9/lib/bundler/runtime.rb:55:in `each'
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/bundler-2.3.9/lib/bundler/runtime.rb:55:in `block in require'
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/bundler-2.3.9/lib/bundler/runtime.rb:44:in `each'
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/bundler-2.3.9/lib/bundler/runtime.rb:44:in `require'
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/bundler-2.3.9/lib/bundler.rb:176:in `require'
	from /Users/isaac/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/sidekiq-ent-2.5.2/bin/sidekiqswarm:40:in `<top (required)>'
	from /Users/isaac/.rbenv/versions/3.1.2/bin/sidekiqswarm:25:in `load'
	from /Users/isaac/.rbenv/versions/3.1.2/bin/sidekiqswarm:25:in `<main>'
```

jaredbeck:
What are the steps to reproduce this issue, starting with clean installs only Rails and PT? Thanks.

isaacbowen:
Thanks for the reply, @jaredbeck! Here's a repo illustrating the issue:

https://github.com/isaacbowen/pt-demo

See the readme for instructions.

jaredbeck:
Thanks for the demo repo, Isaac.

The demo script, `bin/demo`,

```ruby
#!/usr/bin/env ruby
require 'bundler/setup'
Bundler.require('default')
```

differs from the [normal Rails boot process](https://guides.rubyonrails.org/initialization.html).

```
# normal boot process
require_relative "boot"
require "rails/all" # <--- including, notably for this PR, ActiveRecord
Bundler.require(*Rails.groups)
```

As you can see, in the normal Rails boot process, Rails is loaded before PaperTrail.

PaperTrail is a `Railtie` (please see `paper_trail/frameworks/rails/railtie.rb`) so it ".. must be loaded during the Rails boot process .." (`lib/rails/railtie.rb:33`).

> I ran into this when using sidekiq-ent's sidekiqswarm, which loads the bundle via Bundler.require(:default) before booting Rails.

I'm not familiar with `sidekiqswarm`, so I'm sure they have their reasons, but again, it's not following the normal boot process.

jaredbeck (changes requested):
https://github.com/paper-trail-gem/paper_trail/pull/1401#issuecomment-1279887866

isaacbowen:
This is super interesting. Sidekiq is framework-agnostic, and has _some_ Rails-specific intelligence, but isn't built to be intrinsically Rails-specific. This is a different relationship than PaperTrail has to Rails, obviously, so I'm not immediately sure where the fix belongs.

@mperham, can you speak to this?

mperham:
You are referencing ActiveRecord without requiring it. That's it.

jaredbeck:
> You are referencing ActiveRecord without requiring it. That's it.

Thanks for the feedback, @mperham . Do I understand correctly that you approve this PR? That's it? :)

My concern is, we used to have this `require 'activerecord'` at the top of `paper_trail.rb`. It was removed when we created `PaperTrail::Railtie` (https://github.com/paper-trail-gem/paper_trail/pull/1281). @ericproulx and @tlynam, you worked on our Railtie, what do you think about restoring this `require` now that we use [lazy-load hooks](https://github.com/paper-trail-gem/paper_trail/pull/1281/files#diff-f226b8d714b180354bcadd32e8b0f55afb2e4d9e2c639c28bcf248aba53b38e0R25)?

### Testing

> Note that I did not write any tests for this; I'm not immediately sure what the right way to test this is. :)

Would it be possible to test this branch in your `sidekiqswarm` app? You could e.g. use [the `github:` option](https://bundler.io/guides/git.html) in your Gemfile.

I've played around a bit with a clean install of Rails, with and without this `require`, and I don't see any change to the boot process (as expected, because of the `require 'rails/all'`). But, I don't know how to test this for non-Rails applications.

### Todo

If we move forward with this PR, I guess the `require` in `lib/paper_trail/frameworks/active_record.rb` will become redundant, and the comment should be updated. I think the comment in `paper_trail.rb` can also be improved, to reflect some of the project history I've outlined here.

mperham:
@jaredbeck There's two possible fixes:

1. require before referencing
2. use a Railtie to ensure a Rails subsystem is initialized before referencing constants in that subsystem

This PR does (1). You could use (2) and delay any reference to ActiveRecord to within that railtie. I think (2) is useful if you are using paper_trail in a Rails process which does not use AR. Maybe the user wants to use Sequel; by requiring AR you are forcing the process to use more memory and slowing down the boot process.

mperham:
I just read your comment fully. The lazy-load hooks looks exactly like the Railtie I was imagining.

tlynam:
Sorry @isaacbowen, I'm trying to understand the full picture. Looking at the [reproduction demo file](https://github.com/isaacbowen/pt-demo/blob/main/bin/demo), could there be a workaround by requiring active_record or rails?

```
#!/usr/bin/env ruby

require 'bundler/setup'
require "active_record/railtie" # Or require "rails/all"?
Bundler.require('default')

puts 'Loaded successfully!'
```

ericproulx:
@jaredbeck @tlynam 

The issue rely on the constant `RAILS_GTE_7_0` which reference ActiveRecord too ealrly since ActiveRecord may be not loaded. I think `PaperTrail.compatiblitiy` is the rightful place since it's checking ActiveRecord version. Thus, at the end of `self.check_activerecord` I would store the `ar_version` in an instance variable. Finally, you add a `self.ar_version_gte_7?` function that will replace the purpose of `RAILS_GTE_7_0`. That should solve the issue.

isaacbowen:
@ericproulx That approach makes a bunch of sense to me.

@tlynam Ah, sorry! The wider context is that we're using Sidekiq for our jobs, and are using the `sidekiqswarm` executable (from Sidekiq's enterprise edition) for managing worker processes. This executable preloads our bundle _before_ doing anything else, as a part of its memory optimization strategy. This is sane for Sidekiq because it's not a Rails-specific gem. But it results in PaperTrail seeing something it doesn't expect, on load: an environment in which ActiveRecord hasn't yet been required by anything else. Does that make sense?

jaredbeck:
Fixed by #1404. @isaacbowen please test the `master` branch in your swarm app, and let me know.

Thanks, everyone, for contributing to this discussion.

isaacbowen:
@jaredbeck Yep, your fix works well. :) Thank you!

PR #1406, opened 2022-11-12
Simplify Events::Update by moving the timestamp-copying feature up the stack

jaredbeck:
Also, on a related note, forbid certain metadata keys, such as `created_at`. The metadata feature is mainly intended for _new_ columns. Hopefully no one is using these forbidden metadata keys, because that seems .. dangerous. More apropos to this PR, if we support such keys, then moving the timestamp-copying feature out of `Events::Update` might not be possible.

### DIFF
--- a/lib/paper_trail/events/base.rb
+++ b/lib/paper_trail/events/base.rb
@@ -22,6 +22,19 @@ module PaperTrail
     #
     # @api private
     class Base
+      E_FORBIDDEN_METADATA_KEY = <<-EOS.squish
+        Forbidden metadata key: %s. As of PT 14, the following metadata keys are
+        forbidden: %s
+      EOS
+      FORBIDDEN_METADATA_KEYS = %i[
+        created_at
+        id
+        item_id
+        item_subtype
+        item_type
+        updated_at
+      ].freeze
+
       # @api private
       def initialize(record, in_after_callback)
         @record = record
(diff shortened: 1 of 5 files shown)


jaredbeck:
@iiwo WDYT?

lib/paper_trail/events/base.rb
```diff
@@ -44,6 +57,13 @@ def changed_notably?
 
       private
 
+      # @api private
+      def assert_metadatum_key_is_permitted?(key)
```
iiwo, line 61: technically not a boolean method (if you want to follow the convention)

```suggestion
      def assert_metadatum_key_is_permitted(key)
```

lib/paper_trail/record_trail.rb
```diff
@@ -277,11 +269,22 @@ def build_version_on_create(in_after_callback:)
...
       event = Events::Update.new(@record, in_after_callback, is_touch, nil)
       return unless force || event.changed_notably?
+      data = event.data
+
+      # Copy the (recently set) `updated_at` from the record to the `created_at`
+      # of the `Version`. Without this feature, these two timestamps would
+      # differ by a few milliseconds. To some people, it seems a little
+      # unnatural to tamper with creation timestamps in this way. But, this
+      # feature has existed for a long time, almost a decade now, and some users
+      # may rely on it now.
```
iiwo, line 279: this is a very helpful comment, thanks for adding it 👍

iiwo (commented):
> @iiwo WDYT?

looks great! moving that timestamp manipulation to the `build_version_on_update` method is a nice refactor - that's definitely cleaner

lib/paper_trail/events/base.rb
jaredbeck: Good catch :)

PR #1411, opened 2022-12-03
Change return value of #save_with_version when disabled

richardboehme:
This patch changes the return value behavior of #save_with_version to return a boolean if PaperTrail is disabled.

Previously #save_with_version would return `nil` if version tracking is disabled. After applying this commit it will return the result of calling #save on the model.

The main reason for changing this behavior is to be able to react to failing saves when using #save_with_version in our application - for example when saving with invalid data.

--- 

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -150,9 +150,13 @@ module PaperTrail
     # This is an "update" event. That is, we record the same data we would in
     # the case of a normal AR `update`.
     def save_with_version(in_after_callback: false, **options)
-      ::PaperTrail.request(enabled: false) do
-        @record.save(**options)
-      end
+      save_result =
+        ::PaperTrail.request(enabled: false) do
+          @record.save(**options)
+        end
+
+      return save_result unless enabled?
+
       record_update(force: true, in_after_callback: in_after_callback, is_touch: false)
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Breaking Changes
 
-- None
+- [#1411](https://github.com/paper-trail-gem/paper_trail/pull/1411) - Change the return
+  value of #save_with_version when PaperTrail is disabled. It now returns the result
+  of calling #save on the model.
 
 ### Added
 
--- a/spec/dummy_app/app/models/not_on_update.rb
+++ b/spec/dummy_app/app/models/not_on_update.rb
@@ -3,4 +3,6 @@
 # This model does not record versions when updated.
 class NotOnUpdate < ApplicationRecord
   has_paper_trail on: %i[create destroy]
+
+  validates_presence_of :name
 end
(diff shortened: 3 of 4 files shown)


lib/paper_trail/record_trail.rb
```diff
@@ -150,9 +150,13 @@ def source_version
...
     def save_with_version(in_after_callback: false, **options)
-      ::PaperTrail.request(enabled: false) do
-        @record.save(**options)
-      end
+      save_result =
+        ::PaperTrail.request(enabled: false) do
+          @record.save(**options)
+        end
+
+      return save_result unless enabled?
```
jaredbeck, line 158: I don't think `enabled?` is an expensive method, but it will be called by `record_update`, so it might be wasteful to call it a second time here. What do you think of the following?

```ruby
def save_with_version(in_after_callback: false, **options)
  save_result = ::PaperTrail.request(enabled: false) do
    @record.save(**options)
  end
  record_update(force: true, in_after_callback: in_after_callback, is_touch: false)
  save_result
end
```

Actually, this raises some other questions. 

- Why are we calling `record_update` if the `save` failed? That seems wrong.
- Perhaps `save_with_version` should only return true if both `save` and `record_update` succeed.

jaredbeck (changes requested):
Thanks for the PR. Nice work. I have a few questions (below)

lib/paper_trail/record_trail.rb
jaredbeck: These probably aren't questions you _wanted_ me to ask :), but if we're going to make a breaking change, I think we should address these questions holistically.

richardboehme: No worries about asking any questions 😅 

I actually asked those questions myself. My goal was to minimize the change that I'm doing to the API while achieving the goal I wanted. The problem I see with your proposed solution is that it will always return a boolean even if version updates are enabled (currently it would return the created version object).

A previous version that I had in mind was like this:

```ruby
def save_with_version(in_after_callback: false, **options)
  save_result = ::PaperTrail.request(enabled: false) do
    @record.save(**options)
  end
  save_result && record_update(force: true, in_after_callback: in_after_callback, is_touch: false)
end
```

I did not use this variant because it changes the behavior if saving failed. However I can totally see this being better (as you mentioned as well). I'm just not sure if other people might depend on this. What do you think? I'm also not sure what should be returned if version creation failed. I think it would return `nil` currently but I did not check that yet.

jaredbeck: Yeah, it seems wrong to save a `Version` record when `@record.save` failed. That seems like a bug to me. Do you mind fixing it, please, while we're in here? I know it's a bit out of your original scope.

> The problem I see with your proposed solution is that it will always return a boolean ..

I'd think I'd like it to always return a boolean, because that's how Rails' `save` works. You're right that we should treat this as a breaking change. We can increment the major version number (Not in this PR please, I'll increment it at release-time)

WDYT?

richardboehme: Hey @jaredbeck! Is there anything I can do to move this forward? Would love to get consent on the change so that we can get this merged as soon as possible.

PR #1415, opened 2023-01-16
Allow to set --json while generating migration

tvdeyen:
Starting with PaperTrail 13 json is the preferred `object` and `object_changes` column type, but it needs to be changed manually after the migration has been generated. This is not ideal for ie. engines generating the migration during their install process or if you install paper_trail via an CI.

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/lib/generators/paper_trail/install/install_generator.rb
+++ b/lib/generators/paper_trail/install/install_generator.rb
@@ -26,6 +26,12 @@ module PaperTrail
       default: false,
       desc: "Use uuid instead of bigint for item_id type (use only if tables use UUIDs)"
     )
+    class_option(
+      :json,
+      type: :boolean,
+      default: false,
+      desc: "Use json instead of text for object column (use only if database supports json)"
+    )
 
     desc "Generates (but does not run) a migration to add a versions table." \
          "  See section 5.c. Generators in README.md for more information."
@@ -47,6 +53,11 @@ module PaperTrail
     # To use uuid instead of integer for primary key
     def item_id_type_options
       options.uuid? ? "string" : "bigint"
+    end
+
+    # To use uuid instead of integer for primary key
+    def object_column_type
+      options.json? ? "json" : "text"
     end
 
     # MySQL 5.6 utf8mb4 limit is 191 chars for keys used in indexes.
--- a/lib/generators/paper_trail/install/templates/add_object_changes_to_versions.rb.erb
+++ b/lib/generators/paper_trail/install/templates/add_object_changes_to_versions.rb.erb
@@ -7,6 +7,6 @@ class AddObjectChangesToVersions < ActiveRecord::Migration<%= migration_version 
   TEXT_BYTES = 1_073_741_823
 
   def change
-    add_column :versions, :object_changes, :text, limit: TEXT_BYTES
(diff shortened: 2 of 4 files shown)


spec/generators/paper_trail/install_generator_spec.rb
```diff
@@ -124,4 +125,50 @@
...
+    end
+  end
+
+  describe "`--with-changes` and `--json` option set to `true`" do
+    before do
+      prepare_destination
+      run_generator %w[--with-changes --json]
+    end
+
+    # rubocop:disable Layout/LineLength
```
jaredbeck, line 156: Please do not disable rubocop. In this case, please shorten the test title. Feel free to add comments above the test if necessary.

lib/generators/paper_trail/install/install_generator.rb
```diff
@@ -49,6 +55,11 @@ def item_id_type_options
       options.uuid? ? "string" : "bigint"
     end
 
+    # To use uuid instead of integer for primary key
```
jaredbeck, line 58: Was this comment intentional? If so, can you please explain it further?

lib/generators/paper_trail/install/templates/add_object_changes_to_versions.rb.erb
```diff
@@ -7,6 +7,6 @@ class AddObjectChangesToVersions < ActiveRecord::Migration<%= migration_version
   TEXT_BYTES = 1_073_741_823
 
   def change
-    add_column :versions, :object_changes, :text, limit: TEXT_BYTES
+    add_column :versions, :object_changes, :<%= object_column_type %>, limit: TEXT_BYTES
```
jaredbeck, line 10: Can we still use `limit:` with a `:json` column? What effect does it have?

spec/generators/paper_trail/install_generator_spec.rb
```diff
@@ -124,4 +125,50 @@
...
+                contains "t.json     :object, limit: TEXT_BYTES"
+              }
+            }
+          }
+        }
+      )
+    end
+  end
+
+  describe "`--with-changes` and `--json` option set to `true`" do
```
jaredbeck, line 150: Thank you for testing all three contexts!

lib/generators/paper_trail/install/install_generator.rb
```diff
@@ -26,6 +26,12 @@ class InstallGenerator < MigrationGenerator
       default: false,
       desc: "Use uuid instead of bigint for item_id type (use only if tables use UUIDs)"
     )
+    class_option(
+      :json,
+      type: :boolean,
+      default: false,
+      desc: "Use json instead of text for object column (use only if database supports json)"
+    )
```
jaredbeck, line 34: What do you think about making JSON the default?

It's my understanding that all of Rails' built-in [connection adapters](https://github.com/rails/rails/tree/main/activerecord/lib/active_record/connection_adapters) support JSON columns. 

Our default for Postgres (only) should be JSONB.

We can have a --text option, or not. I'm fine either way.

jaredbeck (changes requested):
Thanks Thomas, for getting the ball rolling on this.

Please add documentation (the README) and a changelog entry.

PR #1422, opened 2023-03-24
Fix the issue that unencrypted plaintext values are versioned with ActiveRecord encryption (since Rails 7)

### LINKED ISSUE #1392: Store encrypted values on object_changes column

Rails 7 encrypted fields have their changes on the object stored in plain text

The serializer stores the encrypted value for the change set.

I can't think of one really.  I am assuming you are encrypting values so they won't show up as plain text in the database.

FunnyHector:
Since Rails 7, ActiveRecord introduces a built-in encryption mechanism. When versioning encrypted attributes for JSON columns on PostgreSQL, currently the unencrypted values are saved. This makes the encryption meaningless, and stops people from upgrading to Rails 7. 

This PR is mainly from the patch that @vccoffey posted at https://github.com/paper-trail-gem/paper_trail/issues/1392#issuecomment-1326737082. Thanks for the original idea.

Fixes #1392.

Check the following boxes:

- [ ] Wrote [good commit messages][1].
- [ ] Added tests.
- [ ] The PR relates to *only* one subject with a clear title

### DIFF
--- a/lib/paper_trail/attribute_serializers/object_changes_attribute.rb
+++ b/lib/paper_trail/attribute_serializers/object_changes_attribute.rb
@@ -8,6 +8,12 @@ module PaperTrail
     class ObjectChangesAttribute
       def initialize(item_class)
         @item_class = item_class
+
+        # ActiveRecord since 7.0 has a built-in encryption mechanism
+        @encrypted_attributes =
+          if PaperTrail.active_record_gte_7_0?
+            @item_class.encrypted_attributes&.map(&:to_s)
+          end
       end
 
       def serialize(changes)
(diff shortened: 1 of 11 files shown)


FunnyHector:
@jaredbeck Sorry for pinging, could this be looked at please? I imagine this would be stopping lots of users from upgrading to Rails 7.

spec/support/shared_examples/active_record_encryption.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+      end
+
+      it "reifies encrypted values to decrypted values" do
+        record.update!(supplier: "XYZ", name: "Avocado")
+        expect(record.versions.last.reify.supplier).to eq "ABC"
+      end
+    end
+  end
+end
```
jaredbeck, line 58: Nice tests

lib/paper_trail/attribute_serializers/object_changes_attribute.rb
```diff
@@ -8,6 +8,12 @@ module AttributeSerializers
     class ObjectChangesAttribute
       def initialize(item_class)
         @item_class = item_class
+
+        # ActiveRecord since 7.0 has a built-in encryption mechanism
+        @encrypted_attributes =
+          if PaperTrail.active_record_gte_7_0?
+            @item_class.encrypted_attributes&.map(&:to_s)
```
jaredbeck, line 15: Surprisingly, it [seems](https://github.com/search?q=repo%3Arails%2Frails%20encrypted_attributes&type=code) that rails does not initialize `encrypted_attributes` to an empty array, and so the safe-navigation operator is warranted.

jaredbeck (approved):
Nice work, Hector.

FunnyHector:
@jaredbeck Thanks for reviewing and merging this! Could I ask about the plan for the new release please? We have been using a monkey patch in production. It'd be good if we could get off the monkey patch and use a released version.

TastyPi:
@FunnyHector @jaredbeck could we get a new release including this fix? We need it, I could reference the git commit we want temporarily but I'd prefer official releases :slightly_smiling_face:

jaredbeck:
Released in 15.0.0

PR #1424, opened 2023-04-12
Use polymorphic_name on destroy event

rporrasluc:
In Rails it's common to override the polymorphic names used in polymorphic associations. The item association implemented by paper-trail is a polymorphic association and so should use polymorphic methods.

Destroy event is creating versions of types without considering their polymorphic_name but its base name.

Check the following boxes:

- [ ] Added tests.

### DIFF
--- a/lib/paper_trail/events/destroy.rb
+++ b/lib/paper_trail/events/destroy.rb
@@ -14,7 +14,7 @@ module PaperTrail
       def data
         data = {
           item_id: @record.id,
-          item_type: @record.class.base_class.name,
+          item_type: @record.class.base_class.polymorphic_name,
           event: @record.paper_trail_event || "destroy",
           whodunnit: PaperTrail.request.whodunnit
         }


aramgre:
@rporrasluc why was this PR closed? This would be helpful to us as well.

lib/paper_trail/events/destroy.rb
```diff
@@ -14,7 +14,7 @@ class Destroy < Base
       def data
         data = {
           item_id: @record.id,
-          item_type: @record.class.base_class.name,
+          item_type: @record.class.base_class.polymorphic_name,
```
aramgre, line 17: Should this be just `@record.class.polymorphic_name`? The [default implementation of `polymorphic_name` already considers the `base_class`](https://github.com/rails/rails/blob/e9f38a164e36104cc495d5ac5859e57f6be75f36/activerecord/lib/active_record/inheritance.rb#L210), and if you have overridden it you may have a reason to do so only in the child class.

rporrasluc:
> @rporrasluc why was this PR closed? This would be helpful to us as well.

I think I closed it because it was opened for 2 months without receiving any attention. I am happy to reopen if it will be merged.

PR #1426, opened 2023-05-03
Removes an unnecessary query when checking version limit config.

parkeryoung:
- [ ] Added tests. (N/A)
  code introduces user-observable changes. (N/A)

Due to some monkey patching adding an `on_view` feature in an application I am working on, this query resulted in an infinite loop in our specific case. As this query is unnecessary I thought I'd try to get a slight performance improvement in for the whole project and be able to remove some monkeypatching code from my project's codebase.

I'd also add, this extra query seems to be a result of some upstream change in Rails. I wasn't able to identify the actual source of it in Rails, but this query started happening between version 12 and version 13 of PaperTrail.

If there's anything else I need to add (tests or Changelog entry) or change, let me know and I'll follow up.

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -385,7 +385,7 @@ module PaperTrail
     #
     # @api private
     def version_limit
-      klass = item.class
+      klass = (item_subtype || item_type).constantize
       if limit_option?(klass)
         klass.paper_trail_options[:limit]
       elsif base_class_limit_option?(klass)


jaredbeck:
> Due to some monkey patching adding an on_view feature in an application I am working on, this query resulted in an infinite loop in our specific case.

Thanks for the contribution. Can this issue be reproduced using our [template](https://github.com/paper-trail-gem/paper_trail/blob/master/.github/ISSUE_TEMPLATE/bug-report.md)?

parkeryoung:
The specific issue I had is not a bug in PaperTrail, so I'm not certain the template applies. If you wanted me to use the template, I would also need to put in the monkeypatch code I have written on top of PaperTrail for a feature PaperTrail doesn't have.

This will fix a particular bug for me because of additional features I have added in my project to PaperTrail, it is not a bug in PaperTrail. From the perspective of PaperTrail, this MR is a slight performance improvement as it will no longer make an unnecessary query.

If you would like for me to add the monkeypatch code into the template and attempt to re-create the infinite loop, I could try. However I don't think it would add very much to the discussion. I only mentioned it as context.

PR #1427, opened 2023-05-08
Add config to enable storing the object after the update

sander-deryckere:
I recently started implementing paper-trail in our project, and was pleasantly surprised with the provided features.

But one thing struck me as odd:

> PaperTrail stores the pre-change version of the model, unlike some other auditing/versioning plugins, so you can retrieve the original version. [1](https://github.com/paper-trail-gem/paper_trail/tree/master#1c-basic-usage)

Indeed, when using other audit systems, it's usually the new version that gets saved. And for certain reasons IMO:
* When displaying the versions in a UI, it also makes more sense to show the result of what the user inputted next to who made the change when, thus it also belongs in the same record IMO
* For all non-destroyed records, all version's objects will be non-nil
* The original version is just `widget.versions.first.reify`, a lot more logical than `widget.versions.second&.reify || widget`
* The last version equals the current state. This also means if something happens and doesn't get logged (like an accidental delete directly in the database), you can still restore the most recent version.

As for the main advantage: tracking objects that existed before, that's also possible by writing a migration and touching all the objects you want to track. Or perhaps even write a bulk-insert migration directly to the database.

I didn't want to change the default behaviour, as apparently a lot of people happily use it with the current working. So I chose to add a separate configuration to switch between both ways of working.

I didn't write tests or documentation yet, if you'd like to add this configuration, I'd be happy to write that too.

Check the following boxes:

- [ ] Added tests.

### DIFF
--- a/lib/paper_trail.rb
+++ b/lib/paper_trail.rb
@@ -100,6 +100,18 @@ module PaperTrail
       PaperTrail.config.serializer
     end
 
+    # Set the storage method for storing the object after change instead of before change
+    # @api public
+    def store_after_change=(value)
+      PaperTrail.config.store_after_change= value
+    end
+
+    # Get the storage method: storage before or after applying the changes
+    # @api public
+    def store_after_change
+      PaperTrail.config.store_after_change
+    end
+
     # Returns PaperTrail's global configuration object, a singleton. These
     # settings affect all threads.
     # @api public
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -13,6 +13,7 @@ module PaperTrail
       :association_reify_error_behaviour,
       :object_changes_adapter,
       :serializer,
+      :store_after_change,
       :version_limit,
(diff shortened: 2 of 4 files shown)


jaredbeck:
> Indeed, when using other audit systems, it's usually the new version that gets saved. And for certain reasons .. [etc]

Hi Sander, you're not the first to raise these points, and I agree with some of them! There are [other auditing libraries in Ruby](https://github.com/paper-trail-gem/paper_trail#10-related-libraries-and-ports) which have made different design decisions. 

I'm concerned it would be too difficult to support two different data-models. PT has many methods for querying and traversing versions which would require changes to support both models. Worse, people might want to switch between models halfway through the life of their project. I don't think we can maintain that level of complexity. So, I'm not interested in adding this feature. 

I appreciate your recent contributions, and if you wanted to add something to our README about this, perhaps comparing PT to other Ruby libraries, I'd be happy to review that.

PR #1450, opened 2023-11-16
Add version_error_behavior config option

### LINKED ISSUE #1449: errors are not raised when update, update_columns, or destroy versions fail to create
`paper_trail` is inconsistent when `Version` records cannot be created. in some cases (`create`) an error is raised, in others (`update`, `update_columns`, and `destroy`) an error is logged but not raised. this can result in history silently being lost which is certainly unexpected behavior. 

Currently when creating a new versioned record an error is raised if the `Version` cannot be created - this is because `#save!` is used:

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L62

However, when updating, using `update_columns`, or destroying a versioned record no error is raised if the `Version` cannot be created:

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L111

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L301

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L84

## proposal
i think the behavior should be consistent - all failures should either raise errors or not. personally i don't see any advantage to _not_ raising an error in these cases - it likely means that `whodunnit` isn't defined which i would expect to be an oversight rather than intentional.  i realize this is not backwards compatible so i propose a two step process:

1. an opt-in configuration parameter is added to the current major version which causes `update` and `destroy` failures to raise an error.
2. in the next major version bump this becomes the default behavior.

does this seem reasonable? i'm happy to take this on asap.

modosc:
NOTE: all the specs pass locally but rubocop is failing on a complexity check - i'm unsure exactly how to simplify this code  to satisfy rubocop without making the intent less clear. happy to hear suggestions? for now i've disabled the check around the offending method

Currently when creating a new versioned record an error is raised if the `Version` cannot be created - this is because `#save!` is used:

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L62

However, when updating, using `update_columns`, or destroying a versioned record no error is raised if the `Version` cannot be created:

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L111

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L301

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L84

This pr adds a `always_raise_on_error` global config option. When set, all of the above failures will raise an error. 

Fixes #1449 

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Added
 
-- [#1422](https://github.com/paper-trail-gem/paper_trail/pull/1450) - Add `always_raise_on_error`
-  config option to consistently surface errors when creating `Version` records.
+- [#1422](https://github.com/paper-trail-gem/paper_trail/pull/1450) - Add `version_error_behavior` config
+  config option to control error handling when creating/updating/deleting `Version` records.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Added
 
-- None
+- [#1422](https://github.com/paper-trail-gem/paper_trail/pull/1450) - Add `always_raise_on_error`
+  config option to consistently surface errors when creating `Version` records.
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -289,6 +289,7 @@ Global configuration options affect all threads.
 - object_changes_adapter
 - serializer
 - version_limit
+- always_raise_on_error
 
 Syntax example: (options described in detail later)
 
--- a/README.md
+++ b/README.md
@@ -289,7 +289,7 @@ Global configuration options affect all threads.
 - object_changes_adapter
 - serializer
 - version_limit
-- always_raise_on_error
+- version_error_behavior
 
 Syntax example: (options described in detail later)
 
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -14,7 +14,8 @@ module PaperTrail
       :object_changes_adapter,
       :serializer,
       :version_limit,
-      :has_paper_trail_defaults
+      :has_paper_trail_defaults,
+      :always_raise_on_error
     )
 
     def initialize
--- a/lib/paper_trail/config.rb
+++ b/lib/paper_trail/config.rb
@@ -15,7 +15,7 @@ module PaperTrail
       :serializer,
       :version_limit,
       :has_paper_trail_defaults,
-      :always_raise_on_error
+      :version_error_behavior
     )
 
     def initialize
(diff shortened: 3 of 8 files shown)


paper_trail.gemspec
```diff
@@ -52,7 +52,7 @@ has been destroyed.
   # PT supports request_store versions for 3 years.
   s.add_dependency "request_store", "~> 1.4"
 
-  s.add_development_dependency "appraisal", "~> 2.4.1"
+  s.add_development_dependency "appraisal", "~> 2.5.0"
```
modosc, line 55: this was necessary because i couldn't run `appraisal` locally - i think [this fix](https://github.com/thoughtbot/appraisal/commit/97079f000840338edde73ded08ed6687e0b5223a) is what got it working again

modosc:
is there any chance of getting this looked at? silently not saving version data is a pretty serious issue imho.

modosc:
still not stale, still waiting for maintainers to chime in. 

or alternately i'm happy to help out if someone wants to add permissions.

modosc:
hi @jaredbeck 
> Instead of a boolean configuration option, how about a Symbol, like version_error_behavior: :log # or :exception, :silent

this would be a breaking change because of how the existing code works since currently an error is raised when creating but not updating. if that's ok then i'm happy to make the change.

modosc:
> * Please do not use `rubocop: disable` comments
> 
> > i'm unsure exactly how to simplify this code to satisfy rubocop without making the intent less clear.
> 
> How about extracting private method(s)?

done.

modosc:
not stale, waiting for reviewers

jrodden1:
> not stale, waiting for reviewers

+ 1 
I think I've run into this issue with it silently failing and not raising an error in the past.  It would be rather nice to have this option merged.

jaredbeck:
>> Instead of a boolean configuration option, how about a Symbol, like version_error_behavior: :log # or :exception, :silent

> this would be a breaking change because of how the existing code works since currently an error is raised when creating but not updating. if that's ok then i'm happy to make the change.

How about `version_error_behavior: :legacy # or :log, :exception, :silent`? I'd like to have a single, unified configuration option.

jaredbeck (approved):
Looks great. I appreciate the thorough tests.

jjb:
nice - should there be documentation for how to use this in the readme?

PR #1456, opened 2024-01-27
Suggest using numeric `whodunnit` column type for performance

fatkodima:
One of the easy improvements that could have been done back then was using numeric type for `whodunnit` column. We currently plan to switch to `integer` which is 4 bytes, while using strings to store integer values can easily require much more space, like for `1000000` it will require 8 bytes (padding included), so twice the space. And also integers have statis types, which is easier for databases to work with.

So, I think it is a good idea at least to suggest people to consider switching to numeric types in the migration's comment.

- [ ] Added tests.


fatkodima:
This is valid.

fatkodima:
Ping.

jaredbeck (approved):

jaredbeck:
Good suggestion, thanks

PR #1458, opened 2024-01-27
Fix serializing of time/date columns using yaml serializer

fatkodima:

I noticed, that for yaml serializer date/time objects are serialized as ruby objects. Something like
```yml
--- !ruby/object:ActiveSupport::TimeWithZone
utc: 2024-01-27 18:01:54.627764000 Z
zone: !ruby/object:ActiveSupport::TimeZone
  name: Etc/UTC
time: 2024-01-27 18:01:54.627764000 Z
```

This generates 179 bytes per field.

But that should be serialized into the string. Something like
```yml
--- 2024-01-27 18:03:07 UTC
```

That is 28 bytes long, so 150 bytes difference.

Considering that most people have at least 2 datetime columns (`created_at` and `updated_at`) for each table, that saves 300 bytes per row in the table. 
For example, if we have a table with 4 billion rows, that is `4 * 10^9 * 300 / 10^9` = 1200 GB 😱 saved.

- [x] Added tests.

### DIFF
--- a/spec/paper_trail/attribute_serializers/object_attribute_spec.rb
+++ b/spec/paper_trail/attribute_serializers/object_attribute_spec.rb
@@ -40,6 +40,27 @@ module PaperTrail
           end
         end
       end
+
+      describe "#serialize" do
+        it "serializes a time object into a plain string" do
+          time = Time.zone.local(2015, 7, 15, 20, 34, 0)
+          attrs = { "created_at" => time }
+          described_class.new(Widget).serialize(attrs)
+          expect(attrs["created_at"]).not_to be_a(ActiveSupport::TimeWithZone)
+          expect(attrs["created_at"]).to be_a(String)
+          expect(attrs["created_at"]).to match(/2015/)
+        end
+      end
+
+      describe "#deserialize" do
+        it "deserializes a time object correctly" do
+          time = 1.day.ago
+          attrs = { "created_at" => time }
+          described_class.new(Widget).serialize(attrs)
+          described_class.new(Widget).deserialize(attrs)
+          expect(attrs["created_at"].to_i).to eq(time.to_i)
+        end
+      end
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1458](https://github.com/paper-trail-gem/paper_trail/pull/1416) - Fix serializing
+  date/time columns using yaml serializer. Previously, these were serialized as ruby
+  objects which use a lot of space. Now they are serialized into strings.
 
 ## 15.1.0 (2023-10-22)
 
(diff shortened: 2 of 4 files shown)


fatkodima:
Ok, this is not working for MySQL (`type_cast` still returns a Date/Time object instead of String).
But, json serializer in this gem already works as expected and converts it using `.as_json`. Maybe we should use this too instead of `type_cast`?

jaredbeck:
> Ok, this is not working for MySQL ..

It sounds like we have insufficient test coverage, then?

> But, json serializer in this gem already works as expected ...

We've had a few issues with YAML serialization over the years, and I've often thought of changing the default from YAML to JSON. In fact, I think the best choice is a `jsonb` column, which obviates PT-serialization, and enables performant db queries.

You can also write a custom serializer. In fact, it would be great if you could try that first in your own app, and report back after a few months in production.

Please see https://github.com/paper-trail-gem/paper_trail?tab=readme-ov-file#6b-custom-serializer

fatkodima:
> You can also write a custom serializer. In fact, it would be great if you could try that first in your own app, and report back after a few months in production.

We are currently running this patch in production for postgres for 4 months already.

jaredbeck:
> We are currently running this patch in production for postgres for 4 months already.

Oh, great! Which RDBMS are you using in production? I'm trying to understand the above statement "this is not working for MySQL"

spec/paper_trail/attribute_serializers/object_attribute_spec.rb
```diff
@@ -40,6 +40,32 @@ module AttributeSerializers
...
+          attrs = { "created_at" => time }
+          described_class.new(Widget).serialize(attrs)
+
+          if ENV["DB"] == "postgres" || ENV["DB"] == "sqlite"
+            expect(attrs["created_at"]).not_to be_a(ActiveSupport::TimeWithZone)
+            expect(attrs["created_at"]).to be_a(String)
+            expect(attrs["created_at"]).to match(/2015/)
+          else
+            expect(attrs["created_at"].to_i).to eq(time.to_i)
+          end
```
jaredbeck, line 56: Could you please add a comment here explaining the difference between RDBMS?

lib/paper_trail/attribute_serializers/cast_attribute_serializer.rb
```diff
@@ -44,7 +44,13 @@ def deserialize(attr, val)
...
 
       def serialize(attr, val)
-        AttributeSerializerFactory.for(@klass, attr).serialize(val)
+        serialized = AttributeSerializerFactory.for(@klass, attr).serialize(val)
+        case serialized
+        when Date, Time
+          @klass.connection.type_cast(serialized)
+        else
+          serialized
+        end
```
jaredbeck, line 53: This seems similar to `TypeSerializers::PostgresArraySerializer`, in that it wraps a particular type (in this case `Date, Time`). Should we follow that pattern, and have a `TypeSerializers::DateTimeSerializer`?

fatkodima:
Updated with the suggestion, please take a look.

PR #1468, opened 2024-04-29
Fixes Rails 7.1 raise_on_assign_to_attr_readonly

### LINKED ISSUE #1467: reify tries to set the value for a read_only attribute
Rails 7.1 modifies raise_on_assign_to_attr_readonly [link](https://blog.kiprosh.com/rails-7-1-raises-on-assignment-to-readonly-attributes/) to be enabled by default. In the past, the code would silently carry on. I'm now fixing our code base to no longer write to a read_only attribute. 

Reify sets a model's readonly attribute here:
```
    if model.has_attribute?(k)
          model[k.to_sym] = v
```

Check the following boxes:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  ruby "3.3.1"
  source "https://rubygems.org"
  gem "activerecord", "7.1.3.2"
  gem "minitest", "5.21.2"
  gem "paper_trail", "15.1.0", require: false
  gem "sqlite3", "1.5"
end

require "active_record"
require "minitest/autorun"
require "logger"

ActiveRecord.raise_on_assign_to_attr_readonly = true
ActiveRecord.use_yaml_unsafe_load = true

    t.timestamps null: false
  end

  create_table :versions do |t|
    t.string :event, null: false
    t.string :whodunnit
    t.datetime :created_at
  end
end
require "paper_trail"

  has_paper_trail
  attr_readonly :id
end

  def test_1
    user = User.create(first_name: "Jane")
    user.update! first_name: "Janet"

    versions = user.versions
    old_user = versions.last.reify
  end
end

```

Here is the output that I see....
```
BugTest#test_1:
ActiveRecord::ReadonlyAttributeError: id
```

professor:
Fixes #1467

Check the following boxes:

- [x] Added tests.

I manually tested this PR with these commands:
```
DB=sqlite  bundle exec appraisal rails-7.1 rspec /Users/todd.sedano/workspace/paper_trail/spec/models/wotsit_spec.rb
DB=sqlite  bundle exec appraisal rails-7.0 rspec /Users/todd.sedano/workspace/paper_trail/spec/models/wotsit_spec.rb
DB=sqlite  bundle exec appraisal rails-6.1 rspec /Users/todd.sedano/workspace/paper_trail/spec/models/wotsit_spec.rb
```

### DIFF
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -92,7 +92,7 @@ module PaperTrail
       # @api private
       def reify_attribute(k, v, model, version)
         if model.has_attribute?(k)
-          model[k.to_sym] = v
+          model[k.to_sym] = v unless model.class.readonly_attribute?(k.to_s)
         elsif model.respond_to?("#{k}=")
           model.send("#{k}=", v)
         elsif version.logger
--- a/spec/models/wotsit_spec.rb
+++ b/spec/models/wotsit_spec.rb
@@ -3,12 +3,39 @@
 require "spec_helper"
 
 RSpec.describe Wotsit, versioning: true do
-  it "update! records timestamps" do
-    wotsit = described_class.create!(name: "wotsit")
-    wotsit.update!(name: "changed")
-    reified = wotsit.versions.last.reify
-    expect(reified.created_at).not_to(be_nil)
-    expect(reified.updated_at).not_to(be_nil)
+  context "when handling attr_readonly attributes" do
+    context "without raise_on_assign_to_attr_readonly" do
+      before do
+        # Rails 7.1 first introduces this setting, and framework_defaults 7.0 has it as false
+        if ActiveRecord.respond_to?(:raise_on_assign_to_attr_readonly)
+          ActiveRecord.raise_on_assign_to_attr_readonly = false
+        end
+      end
+
+      it "update! records timestamps" do
+        wotsit = described_class.create!(name: "wotsit")
+        wotsit.update!(name: "changed")
+        reified = wotsit.versions.last.reify
+        expect(reified.created_at).not_to(be_nil)
+        expect(reified.updated_at).not_to(be_nil)
+      end
+    end
+
+    if ActiveRecord.respond_to?(:raise_on_assign_to_attr_readonly)
+      context "with raise_on_assign_to_attr_readonly enabled" do
+        before do
+          ActiveRecord.raise_on_assign_to_attr_readonly = true
+        end
+
+        it "update! records timestamps" do
+          wotsit = described_class.create!(name: "wotsit")
+          wotsit.update!(name: "changed")
+          reified = wotsit.versions.last.reify
+          expect(reified.created_at).not_to(be_nil)
+          expect(reified.updated_at).not_to(be_nil)
+        end
+      end
+    end
   end
 
   it "update! does not raise error" do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,10 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1467](https://github.com/paper-trail-gem/paper_trail/issues/1467) - Rails 7.1
+  enables ActiveRecord.raise_on_assign_to_attr_readonly so that writing to a
+  attr_readonly raises an exception. Fixes paper trail to gracefully handle this
+  situation.
 
 ### Dependencies
 
--- a/spec/dummy_app/app/models/wotsit.rb
+++ b/spec/dummy_app/app/models/wotsit.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class Wotsit < ApplicationRecord
+  attr_readonly :id
   has_paper_trail
 
   belongs_to :widget, optional: true


lib/paper_trail/reifier.rb
```diff
@@ -91,14 +91,17 @@ def init_unversioned_attrs(attrs, model)
...
+          if model.has_attribute?(k)
+            model[k.to_sym] = v
+          elsif model.respond_to?("#{k}=")
+              model.send("#{k}=", v)
+          elsif version.logger
+            version.logger.warn(
+              "Attribute #{k} does not exist on #{version.item_type} (Version id: #{version.id})."
+            )
+          end
+        rescue ActiveRecord::ReadonlyAttributeError
```
technicalpickles, line 104: Is it worth logging a warning here, the line above for when there isn't an attribute?

If there isn't, it'd be good have a comment noting when that can happen.

lib/paper_trail/reifier.rb
```diff
@@ -91,14 +91,17 @@ def init_unversioned_attrs(attrs, model)
...
-          model.send("#{k}=", v)
-        elsif version.logger
-          version.logger.warn(
-            "Attribute #{k} does not exist on #{version.item_type} (Version id: #{version.id})."
-          )
+        begin
+          if model.has_attribute?(k)
+            model[k.to_sym] = v
+          elsif model.respond_to?("#{k}=")
+              model.send("#{k}=", v)
```
technicalpickles, line 98: https://github.com/paper-trail-gem/paper_trail/issues/1467 notes that this is the line that can raise the error. If that is the case, should the begin/rescue wrap just this line?

lib/paper_trail/reifier.rb
technicalpickles, line 104: Found where this was implemented in https://github.com/rails/rails/pull/46105/files#diff-55ee06def7a48911bcbe1a6204576b805a842db2bac44ac66d1732d8c1f44db0R4  ... and doesn't exist before 7.1 unfortunately 😞 

https://github.com/rails/rails/blob/1060cec88bfc5d45a6563d6966f6cb7533057b2b/activerecord/lib/active_record/readonly_attributes.rb#L3-L9

So we need a different way to handle it 🤔  Maybe a `handle_readonly_error` helper that takes a block, except it's defined to yield rescue this only on 7.1 (or when it's defined?), otherwise it's an no-op to just yield.

spec/models/wotsit_spec.rb
```diff
@@ -11,6 +11,40 @@
     expect(reified.updated_at).not_to(be_nil)
   end
 
+  context 'with rails 7.1 attr_readonly' do
+    context 'with rails 7.1 config, default is true' do
```
technicalpickles, line 15: If I am reading it correctly, this is the behavior pre-7.1. I think that means the context names can be generalized.

lib/paper_trail/reifier.rb
professor: Thanks. I've double checked our Datadog logs and have confirmed that the bug report is correct.

lib/paper_trail/reifier.rb
professor: Prior to Rails 7.x, active record would silently "fail" when writing to a read only attribute. I felt that paper trail mimicking this behavior (and not logging) would be fine. 

In one production system, logging would generate 43k of log entries.

professor: Thank you for this feedback, I've updated the code.

spec/models/wotsit_spec.rb
professor: Thank you for the feedback, I've updated the tests.

lib/paper_trail/reifier.rb
```diff
@@ -92,7 +92,7 @@ def init_unversioned_attrs(attrs, model)
       # @api private
       def reify_attribute(k, v, model, version)
         if model.has_attribute?(k)
-          model[k.to_sym] = v
+          model[k.to_sym] = v unless model.class.readonly_attribute?(k.to_s)
```
ngan, line 95: I think this is pretty dangerous since it silently throws away assignments of readonly attributes without the developer knowing about it. This is the exact problem that [config.active_record.raise_on_assign_to_attr_readonly](https://guides.rubyonrails.org/configuring.html#config-active-record-raise-on-assign-to-attr-readonly) was looking to address.

I see some viable options:
1. We consider paper_trail's reify operation as a "dangerous" operation that automatically disable `attr_readonly` declarations (somehow) and forces the attributes to be rewritten.
2. We take the stance you can't have both worlds. As in, you can't reify a record if it's going to set a readonly attribute. This is basically not changing anything.
3. We only set the value _iff_ it's a different value. This allows the reify to work in some cases and fail (loudly) in others. This seems a little odd to me, but still, it's an option.

jaredbeck: > We only set the value iff it's a different value. This allows the reify to work in some cases and fail (loudly) in others. This seems a little odd to me, but still, it's an option.

This is the approach in https://github.com/paper-trail-gem/paper_trail/pull/1473. Any concerns?

jaredbeck:
Should we close this in favor of https://github.com/paper-trail-gem/paper_trail/pull/1473?

ngan:
> Should we close this in favor of #1473?

Yes, I agree.

lib/paper_trail/reifier.rb
ngan: I talked about this with someone today and came to the conclusion that option 1 is actually the best. Option 3 happens to work sometimes for folks. However, if you consider the business logic decision to add `attr_readonly` to an attribute at a certain/later point in time now affects your ability to `reify` a record.

In my opinion, `reify`ing a record is a break-the-glass operation that should just always work (similar to `update_column`). At the very least, the assignment should just always work. Devs can then call `save(validate: false)` if they want to persist the an invalid record.

As for the implementation of option 1, I think papertrail will have to do some tricks to temporarily empty out (`[]`) the readonly array for a model while it's doing it assignment for the reify. Or, it can use private apis to set the attributes directly, bypassing the `raise`. Or, maybe Rails will take a patch to allow a bypass block similar to [`no_touching`](https://api.rubyonrails.org/classes/ActiveRecord/NoTouching/ClassMethods.html#method-i-no_touching).

PR #1473, opened 2024-05-16
Only set the attribute value in reify when it's different

tubaxenor:
- [ ] Added tests.

This is related to https://github.com/paper-trail-gem/paper_trail/pull/1468/files#r1601766553

We shouldn't set something again if the value is no difference from the original value. In some cases such like read only attributes might already set with a default value and will not change over time, this is helpful and will not violate Rails 7.1's new default config `raise_on_assign_to_attr_readonly`.

### DIFF
--- a/lib/paper_trail/reifier.rb
+++ b/lib/paper_trail/reifier.rb
@@ -92,7 +92,7 @@ module PaperTrail
       # @api private
       def reify_attribute(k, v, model, version)
         if model.has_attribute?(k)
-          model[k.to_sym] = v
+          model[k.to_sym] = v unless model.attributes[k] == v
         elsif model.respond_to?("#{k}=")
           model.send("#{k}=", v)
         elsif version.logger
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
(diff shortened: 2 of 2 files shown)


lib/paper_trail/reifier.rb
```diff
@@ -92,7 +92,7 @@ def init_unversioned_attrs(attrs, model)
       # @api private
       def reify_attribute(k, v, model, version)
         if model.has_attribute?(k)
-          model[k.to_sym] = v
+          model[k.to_sym] = v unless model.attributes[k] == v
```
jaredbeck, line 95: Why not `[k.to_sym]`?

jaredbeck:
Looks good, please add an entry to the Changlog under "Fixed"

jaredbeck:
@ngan @professor @technicalpickles Thoughts?

ngan:
Yeah, I think this is a reasonable approach. @tubaxenor can make the request changeling updates when he's back online.

fynsta:
We also ran into this problem when upgrading the Rails version. What's the status on this PR?

ngan:
I wanted to x-post this thread:
[paper-trail-gem/paper_trail#1468 (files)](https://github.com/paper-trail-gem/paper_trail/pull/1468/files#r1842848566)

I'm not fully convinced that this is the right approach to solving this problem. See thread for details.

pgiblock:
This is still broken as far as I can tell

PR #1479, opened 2024-06-15
Do not allow multiple `has_paper_trail` definitions for models

### LINKED ISSUE #1478: Calling has_paper_trail on STI models causes 2 inserts to versions

closed without comment.

Check the following boxes:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  ruby "3.3.0"
  source "https://rubygems.org"
  gem "activerecord"
  gem "minitest"
  gem "paper_trail", "12.1.0", require: false
  gem "sqlite3", "~> 1.4"
end

require "active_record"
require "minitest/autorun"
require "logger"

    t.string :type
    t.timestamps null: false
  end

  create_table :versions do |t|
    t.string :event, null: false
    t.string :whodunnit
    t.datetime :created_at
  end
end
require "paper_trail"

  has_paper_trail
end

class Admin < User
  has_paper_trail on: [:create]
end

  def test_1
    assert_difference(-> { PaperTrail::Version.count }, +1) {
    }
  end

  def test_2
    assert_difference(-> { PaperTrail::Version.count }, +1) {
      Admin.create(first_name: "Jane")
    }
  end
end

```

Our real scenario is actually like this:

```ruby
class ApplicationRecord < ActiveRecord::Base
  NON_PAPER_TRAIL_MODELS = %w[
    X
    Y
    Z
  ]

  def self.inherited(other)
    super

    other.has_paper_trail unless NON_PAPER_TRAIL_MODELS.include? other.name
  end
end
```

When the root class is created `class MyModel < ApplicationRecord`, it calls `has_paper_trail` and when `class MyOtherModel < MyModel` is created, it calls it again.

Current workaround:

```ruby
  def self.inherited(other)
    super

    other.has_paper_trail unless other.respond_to?(:paper_trail_options) || NON_PAPER_TRAIL_MODELS.include?(other.name)
  end
```

I would expect the gem to raise an error if that's not allowed or simple ignore the call if nothing has changed, or even better, respect the options for the STI models in the case it is different from the root model.

fatkodima:
Fixes #1478.

There were 3 variants on how to proceed:
1. ignore multiple `has_paper_trail` calls - this can be confusing, for example
```ruby
class Parent < ApplicationRecord
  has_paper_trail on: [:create]
end

class Child < Parent
  has_paper_trail on: [:update]
end
```
For which events we should create `version` records? 🤷 Same without STI (just when we include some module and have the `has_paper_trail` called on the class) or just `has_paper_trail` defined on the `ApplicationRecord` and then in the concrete class too.

2. detect if we use STI and do something smart about it (as suggested in the issue) - this won't work, because when we detect that we use STI, the callbacks are already defined on the parent model and we can't "undefine" them
3. raise if we have multiple calls - implemented in this PR. Looks the most sane to me, but this is a breaking change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Breaking Changes
 
-- None
+- [#1478](https://github.com/paper-trail-gem/paper_trail/issues/1478) Do not allow
+  multiple `has_paper_trail` definitions for models. Previously, when `has_paper_trail`
+  was called on a parent and a child from STI, then possibly multiple `version` records
+  will be created per event (`create`, `destroy` etc).
 
 ### Added
 
--- a/lib/paper_trail/has_paper_trail.rb
+++ b/lib/paper_trail/has_paper_trail.rb
@@ -68,6 +68,8 @@ module PaperTrail
       #
       # @api public
       def has_paper_trail(options = {})
+        raise Error, "has_paper_trail must be called only once" if self < InstanceMethods
+
         defaults = PaperTrail.config.has_paper_trail_defaults
         paper_trail.setup(defaults.merge(options))
       end
--- a/spec/dummy_app/app/models/callback_modifier.rb
+++ b/spec/dummy_app/app/models/callback_modifier.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
-class CallbackModifier < ApplicationRecord
-  has_paper_trail on: []
+module CallbackModifier
+  extend ActiveSupport::Concern
+
+  included do
+    self.table_name = "callback_modifiers"
+  end
 
   def test_destroy
     transaction do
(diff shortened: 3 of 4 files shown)


jaredbeck:
> [option 3] raise if we have multiple calls - implemented in this PR. Looks the most sane to me, but this is a breaking change

I agree. This is the only option that I'm interested in maintaining.

Thanks for the contribution.

Piioo:
And how to handle this:

We have defined for the whole application `has_paper_trail` in `ApplicationRecord`

But One model we need to override `has_paper_trail skip: %i[encrypted_password otp_secret]`

And this is not possible anymore....

Do we have to set for every model separate the has_paper_trail ?

bborn:
👎 pretty sure this introduces a regression. The custom version class example from the docs no longer works for me:

```
class PostVersion < PaperTrail::Version
  # custom behaviour, e.g:
  self.table_name = :post_versions
end

class Post < ActiveRecord::Base
  has_paper_trail versions: {
    class_name: 'PostVersion'
  }
end
```

My example:

```
class Post < ApplicationRecord
  has_paper_trail versions: {
    class_name: "CustomVersion"
  }
```

PR #1498, opened 2024-10-19
Test against rails 8.0

### LINKED ISSUE #1490: Add support for rails 8.0.0.alpha
I'm facing this on `rails db:migrate`

```
PaperTrail 15.1.0 is not compatible with ActiveRecord 8.0.0.alpha. We allow PT
contributors to install incompatible versions of ActiveRecord, and this
warning can be silenced with an environment variable, but this is a bad
idea for normal use. Please install a compatible version of ActiveRecord
instead (>= 6.1, < 7.2). Please see the discussion in paper_trail/compatibility.rb
for details.
```

as obviously it supports rails <7.2
image

fatkodima:
Closes #1490.

- [x] Added tests.


spec/dummy_app/app/models/post_with_status.rb
```diff
@@ -5,5 +5,9 @@
 class PostWithStatus < ApplicationRecord
   has_paper_trail
 
-  enum status: { draft: 0, published: 1, archived: 2 }
+  if ActiveRecord::VERSION::MAJOR >= 8
```
skipkayhil, line 8:
```suggestion
  if ActiveRecord::VERSION::MAJOR >= 7
```

I think the new syntax was added in 7.0 so this would enable removing the condition sooner (when 6.1 is dropped)

fatkodima: Good idea! 👍

skipkayhil (approved):

modosc:
@fatkodima @skipkayhil should we go ahead and merge this?

PR #1504, opened 2025-01-08
Add `reify_original` method.

### LINKED ISSUE #1204: Add `original_version` method to more easily obtain the original version.
- [x] I am committed to building this feature myself
- [x] I will respond to feedback, and build this feature, in a reasonable amount
  of time

Getting the original version of an item is way too difficult. It stems from the fact that calling `reify` on the first `version` returns `nil` and not the original object. That's fine. But it means that you need to do some fancy footwork to get the original version of an object and also handle when an object doesn't have any changes (i.e. only a single `create` version).

You can see the issue in these two spots: 

- https://stackoverflow.com/questions/45353464/how-to-get-original-attribute-for-a-model-with-paper-trail/56258720#56258720

- https://github.com/paper-trail-gem/paper_trail/issues/475

I want to add an `original_version` method to the `has_paper_trail` concern that will return the original version of an object, regardless of how many changes the object has.

So instead of writing something like this `versions.second&.reify || self` you can just write `original_version`.

`versions.second&.reify || self`

### Commit on Our Fork.

We're using `3.0-stable` branch so I forked that and [committed](https://github.com/joshuapinter/paper_trail/commit/b30f006de6b50c9db8af7e22926c9e246b2c18e6) our desired change there. No tests but we can certainly write them if you want this merged into master.

joshuapinter:
Returning the original version is a common use-case and is not as easy as it should be, particularly because calling `reify` on the first version returns `nil`, which is the `object` *before* it was created.

This method simply tries to `reify` from the *second* version, if it exists. If it doesn’t exist (i.e. there has not been any changes to the object), it falls back to `self`, which is the original and *only* version of the object.

Fixes #1204.

Check the following boxes:

- [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
   `rails generate paper_trail:install CommentVersion` created `comment_versions` table
 - `rails generate paper_trail:update_item_subtype` now supports custom version classes via 
   `--version-class-name` option, e.g. `--version-class-name=CommentVersion`
+- [#1204](https://github.com/paper-trail-gem/paper_trail/issues/1204) - Add `reify_original` method for reifying the original object.
 
 ### Fixed
 
--- a/lib/paper_trail/record_trail.rb
+++ b/lib/paper_trail/record_trail.rb
@@ -69,6 +69,13 @@ module PaperTrail
       end
     end
 
+    # Returns the original version of this object or just this object if there has been no changes.
+    #
+    # @api public
+    def reify_original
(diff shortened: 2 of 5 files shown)


joshuapinter:
Okay, I've added specs and an entry into the CHANGELOG. This PR should be ready for review and merging. 

Let me know if you would like to see any changes. 

Thanks.

joshuapinter:
Not Stale. @fatkodima Can we please get this merged in? Tests and everything are done. Working great in our production environment for the last year. Thanks!

joshuapinter:
Not stale. We use this in production. This is ready to go. Please merge @fatkodima .

Issue #222, opened 2013-05-13
Really no way to do eager loading of versions?


coreyward:
I have a collection of objects that I want to inspect at a particular time in the past, e.g.:

```
historical_objects = current_objects.map { |object| object.version_at(date_in_past) }
```

This code is slow due to a lack of “eager loading” (O(n)+1). Is it possible to, without bending paper_trail too far, load versions for a collection from a particular time?

mgraham:
> That's because the `version_at` method uses the `following` scope, which requires an additional query (as it restricts by date). If you wanted you could use Ruby (reduce an additional call to the database) to accomplish the location of the first version after that date while still using the eager association loading like so:
> 
> ```ruby
> Quiz.includes(:versions).first.versions.detect { |q| q.created_at > 30.days.ago }
> ```

This approach is actually very tricky to get correct for a few reasons.  Firstly, the `versions` table does not contain the current state of the model.  Secondly, the first version is typically the `create` version which needs to be excluded.

As near as I can tell, when you call `thing.paper_trail.version_at(date)`:
* if the date is before the `created_at` of the first version, you get `nil`
* if the date is after the `created_at` of the last version, you get the current model

So here's code that takes all that into account (lightly tested):
```
timestamp = DateTime.parse('2023-04-20T04:00:04Z')
records = Quiz.includes(:versions).map do |record|
  result = nil
  if record.updated_at <= timestamp
    result = record
  else
    versions = record.versions.select { |v| v.event == 'update' }.sort_by { |v| v.created_at }.reverse
    if versions.last.created_at <= timestamp
      version = versions.detect { |v| v.created_at <= timestamp }
      result = version.reify if version
    end
  end
  result
end
records.reject! { |record| record.nil? }
```

However, this ends up having a different N+1 problem, because each call to `version.reify` makes another query.

PR #1509, opened 2025-03-13
Disable strict_loading on the version association

searls:
As documented in #222, paper_trail doesn't support eager-loading, which means anyone turning on strict_loading by default:

```ruby
# config/application.rb
config.active_record.strict_loading_by_default = true
config.active_record.strict_loading_mode = :n_plus_one_only
```

Will routinely encounter errors like:

```
ActiveRecord::StrictLoadingViolationError: `Membership` is marked for strict_loading. The ApplicationVersion association named `:versions` cannot be lazily loaded.
```

I'd normally have just configured this on `has_paper_trail_defaults`, but since that only does a `merge` and not a deep-merge (since this is nested under `:versions`), that won't work, and it seems like too much of a foot-gun to expect each and every `has_paper_trail` to remember to set `strict_loading: false` in applications with strict_loading enabled by default

### DIFF
--- a/lib/paper_trail/model_config.rb
+++ b/lib/paper_trail/model_config.rb
@@ -185,6 +185,7 @@ module PaperTrail
         scope,
         class_name: @model_class.version_class_name,
         as: :item,
+        strict_loading: false,
         **options[:versions].except(:name, :scope)
       )
     end


davidrunger:
Specs are failing for reasons not related to this change. I submitted this PR to fix them: https://github.com/paper-trail-gem/paper_trail/pull/1512.

davidrunger:
@searls FYI, #1512 has been merged (which fixed specs being broken on `master`), and so if you rebase with that and push back up, specs should now have a chance of passing.

searls:
@davidrunger cool, passes

modosc:
i'm hesitant to hard-code this here without a way to override it. 

if `strict_loading: false` is set on the association in the model is it overridden when i call `.strict_loading` on the association? eg the example from #222:
```ruby
ContentObject.includes(:versions).strict_loading
```

PR #1511, opened 2025-03-18
Avoid N+1 queries coming from #version_limit when destroying records

### LINKED ISSUE #1510: N+1 query when destroying record with PaperTrail-tracked has_many dependent: :destroy association

closed without comment.

Check the following boxes:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  ruby "3.4.2"
  source "https://rubygems.org"
  gem "activerecord", "8.0.2"
  gem "minitest", "5.25.5"
  gem "sqlite3", "2.6.0"

  # The N+1 queries bug occurs in the latest released version of paper_trail (16.0.0):
  gem "paper_trail", "16.0.0"
  # The bug is fixed in the following branch on my fork.
  # gem 'paper_trail',
  #   github: 'davidrunger/paper_trail',
  #   branch: 'avoid-n+1-queries-in-version_limit-when-destroying'
end

  create_table :customers, force: true do |t|
    t.text :name, null: false
  end

  create_table :orders, force: true do |t|
    t.bigint :customer_id, null: false
    t.string :order_date
  end

  create_table :versions do |t|
    t.string :event, null: false
    t.string :whodunnit
    t.json :object
    t.json :object_changes
    t.datetime :created_at, null: false
    t.index %i[item_type item_id]
  end
end
stdout_logger = Logger.new($stdout)
$active_record_logs = StringIO.new
stringio_logger = Logger.new($active_record_logs)
ActiveRecord::Base.logger = ActiveSupport::BroadcastLogger.new(stdout_logger, stringio_logger)
ActiveRecord::Base.logger.formatter = ActiveSupport::Logger::SimpleFormatter.new

class Customer < ActiveRecord::Base
  has_many :orders, dependent: :destroy
end

class Order < ActiveRecord::Base
  has_paper_trail
end

require "minitest/autorun"

  def test_no_n_plus_1_query_when_destroying_customer
    customer = Customer.create!(name: "Customer")
    customer.orders.create!
    customer.orders.create!

    customer.destroy!
    # NOTE: Explicitly eager-loading the orders (as below) ideally shouldn't be
    # needed, but, anyway, doing so still doesn't avoid the N+1 query.
    # Customer.includes(:orders).find(customer.id).destroy!

    # The assertion below counts SELECT queries for individual orders by id.
    # There should not be any such queries. (There should only be one query for
    # orders that loads together all of the orders of the customer that is being
    # destroyed.) However, with `paper_trail` 16.0.0, this spec will fail, as
    # there _will_ be individual queries by id for each of the orders being
    # destroyed (two total, in this example). These queries are N+1 queries,
    # they are not necessary, and they should not be performed.
    assert_equal(
      0,
      $active_record_logs.string.scan(
        'SELECT "orders".* FROM "orders" WHERE "orders"."id" = ?'
      ).size
    )
  end
end
```

```
$ ruby n_plus_1_queries.rb
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
-- create_table(:customers, {force: true})
   -> 0.0229s
-- create_table(:orders, {force: true})
   -> 0.0011s
-- create_table(:versions)
   -> 0.0054s
Run options: --seed 18890

# Running:

  TRANSACTION (0.1ms)  BEGIN immediate TRANSACTION
  Customer Create (0.6ms)  INSERT INTO "customers" ("name") VALUES (?) RETURNING "id"  [["name", "Customer"]]
  TRANSACTION (0.0ms)  COMMIT TRANSACTION
  TRANSACTION (0.1ms)  BEGIN immediate TRANSACTION
  Order Create (0.5ms)  INSERT INTO "orders" ("customer_id") VALUES (?) RETURNING "id"  [["customer_id", 1]]
  PaperTrail::Version Create (0.2ms)  INSERT INTO "versions" ("item_type", "item_id", "event", "object_changes", "created_at") VALUES (?, ?, ?, ?, ?) RETURNING "id"  [["item_type", "Order"], ["item_id", 1], ["event", "create"], ["object_changes", "{\"id\":[null,1],\"customer_id\":[null,1]}"], ["created_at", "2025-03-18 05:24:47.887680"]]
  TRANSACTION (0.1ms)  COMMIT TRANSACTION
  TRANSACTION (0.1ms)  BEGIN immediate TRANSACTION
  Order Create (0.7ms)  INSERT INTO "orders" ("customer_id") VALUES (?) RETURNING "id"  [["customer_id", 1]]
  PaperTrail::Version Create (0.3ms)  INSERT INTO "versions" ("item_type", "item_id", "event", "object_changes", "created_at") VALUES (?, ?, ?, ?, ?) RETURNING "id"  [["item_type", "Order"], ["item_id", 2], ["event", "create"], ["object_changes", "{\"id\":[null,2],\"customer_id\":[null,1]}"], ["created_at", "2025-03-18 05:24:47.893108"]]
  TRANSACTION (0.1ms)  COMMIT TRANSACTION
  TRANSACTION (0.0ms)  BEGIN immediate TRANSACTION
  Order Load (0.3ms)  SELECT "orders".* FROM "orders" WHERE "orders"."customer_id" = ?  [["customer_id", 1]]
  PaperTrail::Version Create (0.2ms)  INSERT INTO "versions" ("item_type", "item_id", "event", "object", "object_changes", "created_at") VALUES (?, ?, ?, ?, ?, ?) RETURNING "id"  [["item_type", "Order"], ["item_id", 1], ["event", "destroy"], ["object", "{\"id\":1,\"customer_id\":1,\"order_date\":null}"], ["object_changes", "{\"id\":[1,null],\"customer_id\":[1,null],\"order_date\":[null,null]}"], ["created_at", "2025-03-18 05:24:47.897182"]]
  Order Load (0.2ms)  SELECT "orders".* FROM "orders" WHERE "orders"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  Order Destroy (0.1ms)  DELETE FROM "orders" WHERE "orders"."id" = ?  [["id", 1]]
  PaperTrail::Version Create (0.1ms)  INSERT INTO "versions" ("item_type", "item_id", "event", "object", "object_changes", "created_at") VALUES (?, ?, ?, ?, ?, ?) RETURNING "id"  [["item_type", "Order"], ["item_id", 2], ["event", "destroy"], ["object", "{\"id\":2,\"customer_id\":1,\"order_date\":null}"], ["object_changes", "{\"id\":[2,null],\"customer_id\":[1,null],\"order_date\":[null,null]}"], ["created_at", "2025-03-18 05:24:47.901533"]]
  Order Load (0.1ms)  SELECT "orders".* FROM "orders" WHERE "orders"."id" = ? LIMIT ?  [["id", 2], ["LIMIT", 1]]
  Order Destroy (0.1ms)  DELETE FROM "orders" WHERE "orders"."id" = ?  [["id", 2]]
  Customer Destroy (0.1ms)  DELETE FROM "customers" WHERE "customers"."id" = ?  [["id", 1]]
  TRANSACTION (0.0ms)  COMMIT TRANSACTION
F

Finished in 0.062776s, 15.9295 runs/s, 15.9295 assertions/s.

  1) Failure:
BugTest#test_no_n_plus_1_query_when_destroying_customer [n_plus_1_queries.rb:77]:
Expected: 0
  Actual: 2

1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```

Two of the queries seen in the test output above and highlighted in the screenshot below are unnecessary N+1 queries:

Image

Those N+1 queries for `Order` records (just after creating a PaperTrail::Version and just before actually destroying the Order record) occur because of the reference to `item` here in the PaperTrail::VersionConcern#version_limit method: https://github.com/paper-trail-gem/paper_trail/blob/94e9c0daa28c5a60441f5e173f3e6b1d2c555918/lib/paper_trail/version_concern.rb#L388

I believe that this N+1 query was probably introduced here: https://github.com/paper-trail-gem/paper_trail/pull/1309/files#diff-ce609ac9edbdf2df8bfe15246950dce5efa7efac84609883bb7a210bdef3ba39L386

davidrunger:
Fixes #1510

Check the following boxes:

- [x] Added tests.

## Avoiding N+1 queries when destroying a record with a `has_many` `dependent: :destroy` association

This change avoids N+1 queries that otherwise currently occur on the `master` branch when destroying a record with a `has_many` `dependent: :destroy` association where the `has_many` model is tracked by PaperTrail.

## My thoughts about this PR

Some of the things mentioned below are somewhat questionable/hacky. I understand if a PR in this form isn't really desired. But I believe that this change does fix the bug linked above, and I personally think that the changes herein are beneficial improvements to the gem and to its test suite.

## Background context

I believe that the N+1 queries that this PR aims to remove were introduced by [this change](https://github.com/paper-trail-gem/paper_trail/pull/1309/files#diff-ce609ac9edbdf2df8bfe15246950dce5efa7efac84609883bb7a210bdef3ba39L386), and this current change essentially undoes that specific linked change in #1309 (though this current change handles the fact that `item_subtype` is an optional column in a slightly different way than did the original code prior to that change that I believe introduced this N+1 query).

## Adding `n_plus_one_control` gem

In order to test this change, this PR adds the [`n_plus_one_control` gem](https://github.com/palkan/n_plus_one_control) as a development dependency. I understand that adding dependencies (even development-only dependencies) is a somewhat significant change, and that we might not want to add that gem for the sake of (at least for now) a single test. Nevertheless, it seemed to me like an effective way to test this change, and also that having the `n_plus_one_control` gem available might also be useful to write other specs in the future related to N+1 queries (or, rather, to a lack thereof), and so I'm including that gem in this proposed change.

## Using an instance variable in the spec

I think that the nature of how that gem works (maybe specifically in conjunction with the fact that the spec being added herein destroys the `Customer` record that is somewhat the subject of the spec?) necessitates capturing the `Customer` record with an instance variable, so I disabled the `RSpec/InstanceVariable` cop for that spec.

## Hacking `PostVersion` to act like it has no `item_subtype` method

In working on this change, in one iteration, I wrote code that assumes that the object into which the `PaperTrail::VersionConcern` module has been included will respond to an `item_subtype` message (i.e. have such a method). Specifically, I had written `(item_subtype || item_type).constantize`. Although this code will fail in any project that uses PaperTrail but which does not add an `item_subtype` column for the versions table, all specs nevertheless passed (when executed with `DB=sqlite bundle exec appraisal rails-8.0 rspec`). This implied to me that there is not currently any spec coverage/awareness of the fact that `item_subtype` is an optional column that will not be available in all projects.

I believe that this is the observation that was being made by @jaredbeck [here](https://github.com/paper-trail-gem/paper_trail/pull/1381#issuecomment-1153464553):

> Looking at our SimpleCov report, It seems we are missing a test for when `item_subtype` is _absent_? That seems important because `item_subtype` is an optional column.

The reason that the test suite doesn't have coverage of this possibility (not having an `item_subtype` method) is because all of the version models in the test suite inherit from `PaperTrail::Version`, which, in the spec setup, uses the `versions` table, which is [defined with an `item_subtype` column](https://github.com/paper-trail-gem/paper_trail/blob/94e9c0daa28c5a60441f5e173f3e6b1d2c555918/spec/dummy_app/db/migrate/20110208155312_set_up_test_tables.rb#L82). So all of the version models in the test suite respond to `item_subtype`, even if their respective tables (e.g. `post_versions`) do not have an `item_subtype` column.

This differs from the situation of real applications using `paper_trail`, where in many (most?) of those applications the PaperTrail version model(s) will _not_ respond to an `item_subtype` method.

To try to give the test suite some awareness of this possibility, I hacked one of the version models -- `PostVersion` -- to raise an error if #item_subtype is called and to return `false` when queried with `respond_to?(:item_subtype)`. This is ugly/hacky, but it does at least have the beneficial effect that the problematic iteration of my code change that I mentioned above would be caught by some failing specs.

### DIFF
--- a/lib/paper_trail/version_concern.rb
+++ b/lib/paper_trail/version_concern.rb
@@ -385,7 +385,7 @@ module PaperTrail
     #
     # @api private
     def version_limit
-      klass = item.class
+      klass = ((respond_to?(:item_subtype) && item_subtype) || item_type).constantize
       if limit_option?(klass)
         klass.paper_trail_options[:limit]
       elsif base_class_limit_option?(klass)
--- a/spec/dummy_app/app/versions/post_version.rb
+++ b/spec/dummy_app/app/versions/post_version.rb
@@ -2,4 +2,19 @@
 
 class PostVersion < PaperTrail::Version
   self.table_name = "post_versions"
+
+  # BEGIN Hack to simulate a version class that doesn't have the optional `item_subtype` column. >>>
+  def item_subtype
+    raise "This method should not be called!"
+  end
+
+  # We will follow Ruby's `respond_to?` method signature, including an optional boolean parameter.
+  # rubocop:disable Style/OptionalBooleanParameter
+  def respond_to?(method_name, include_private = false)
+    # rubocop:enable Style/OptionalBooleanParameter
+    return false if method_name.to_sym == :item_subtype
+
+    super
+  end
+  # <<< END Hack to simulate a version class that doesn't have the optional `item_subtype` column.
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- [#1511](https://github.com/paper-trail-gem/paper_trail/pull/1511) - Avoid N+1
+  queries when destroying a record with a `has_many` `dependent: :destroy`
+  association where the `has_many` target model is tracked by PaperTrail.
 
 ### Dependencies
 
--- a/paper_trail.gemspec
+++ b/paper_trail.gemspec
@@ -70,6 +70,7 @@ has been destroyed.
   s.add_development_dependency "ffaker", "~> 2.20"
   s.add_development_dependency "generator_spec", "~> 0.9.4"
   s.add_development_dependency "memory_profiler", "~> 1.0.0"
+  s.add_development_dependency "n_plus_one_control", "~> 0.7.2"
 
   # For `spec/dummy_app`. Technically, we only need `actionpack` (as of 2020).
   # However, that might change in the future, and the advantages of specifying a
--- a/spec/models/order_spec.rb
+++ b/spec/models/order_spec.rb
@@ -11,5 +11,22 @@ RSpec.describe Order, :versioning do
 
       expect(customer.versions.count).to(eq(3))
     end
+
+    # We need to use an instance variable to work with the `n_plus_one_control` gem in this spec.
+    # rubocop:disable RSpec/InstanceVariable
+    context "when the record has many associated records (N+1 check)", :n_plus_one do
+      populate do |n|
+        @customer = Customer.create!
+
+        n.times do
+          @customer.orders.create!
+        end
+      end
+
+      it "does not cause an N+1 query" do
+        expect { @customer.destroy! }.to perform_constant_number_of_queries
+      end
+    end
+    # rubocop:enable RSpec/InstanceVariable
   end
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -69,6 +69,7 @@ require "action_controller/railtie"
 # by the `Bundler.require` in `config/application.rb`.
 require "paper_trail"
 require "ffaker"
+require "n_plus_one_control/rspec"
 require "rspec/rails"
 require "rails-controller-testing"
 


davidrunger:
Specs are failing for reasons not related to this change. I submitted this PR to fix them: https://github.com/paper-trail-gem/paper_trail/pull/1512.

davidrunger:
Woops, I'm sorry, this branch and PR are currently in a bad state, because I made some changes to it for the sake of merging them into my own fork, forgetting about the fact that the branch was also being used for this PR. 😥 I will fix the branch up (and pull in the recently merged fix for specs failing on `master`, #1512), and then push it back up.

davidrunger:
> Woops, I'm sorry, this branch and PR are currently in a bad state, because I made some changes to it for the sake of merging them into my own fork, forgetting about the fact that the branch was also being used for this PR. 😥 I will fix the branch up (and pull in the recently merged fix for specs failing on `master`, #1512), and then push it back up.

Done. The diff is back to the intended much, much smaller size, and, having pulled in #1512, all GitHub Actions checks are now passing.

lib/paper_trail/version_concern.rb
```diff
@@ -385,7 +385,7 @@ def sibling_versions
     #
     # @api private
     def version_limit
-      klass = item.class
```
davidrunger, line 388: Calling `item` here is essentially what was causing the N+1 query, since instantiating the version's `item` requires querying the database for the `item`'s attributes.

This change avoids that database query by instead just referencing the `item_subtype` or `item_type` fields of the version record, which does not require any additional database query.

spec/dummy_app/app/versions/post_version.rb
```diff
@@ -2,4 +2,21 @@
...
+  # BEGIN Hack to simulate a version class that doesn't have the optional `item_subtype` column. >>>
+  def item_subtype
+    raise "This method should not be called!"
+  end
+
+  # We will follow Ruby's `respond_to?` method signature, including an optional boolean parameter.
+  # rubocop:disable Style/OptionalBooleanParameter
+  def respond_to?(method_name, include_private = false)
+    # rubocop:enable Style/OptionalBooleanParameter
+    if method_name.to_sym == :item_subtype
```
batter, line 15: Could shorten this syntax with a guard clause:

```suggestion
    # rubocop:enable Style/OptionalBooleanParameter
    return false if method_name.to_sym == :item_subtype
    
    super
```

davidrunger: I made this change. Thank you for the suggestion!

davidrunger (commented):
Thank you for taking a look, @batter ! :bowing_man: I integrated one of your suggestions, and had some pedantic pushback against the other. :upside_down_face:

parkeryoung:
https://github.com/paper-trail-gem/paper_trail/pull/1426 This looks related to this merge I tried to make years ago. I'd be very happy to see this get merged in. Good work!

PR #1514, opened 2025-03-31
Drop rails 6.1/7.0 and ruby 3.0/3.1 support

modosc:
~note: this depends on https://github.com/paper-trail-gem/paper_trail/pull/1515~

this fixes #1513  

this pr removes support for:
 * rails 6.1
 * rails 7.0
 * ruby 3.0
 * ruby 3.1

all of which are will be EOL'd by 2025-04-01. 

the only significant change to the library code was the removal of `PaperTrail.active_record_gte_7_0?`.

it also upgrades `rspec-rails` and `rubocop-rspec` to their most recent versions (both of these were previously blocked by rails 6.1 support). i ran autofixes and updated the `.rubocop.yml` file - each of these were done in a separate change so they should be easily undoable. 

Check the following boxes:

- [x] Wrote [good commit messages][1].
- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [ ] Added tests.
- [x] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new 
  code introduces user-observable changes.
- [x] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.

[1]: http://chris.beams.io/posts/git-commit/

PR #1515, opened 2025-03-31
Upgrade rubocop + rubocop plugins

modosc:

Check the following boxes:

- [ ] Wrote [good commit messages][1].
- [ ] Added tests.
- [ ] The PR relates to *only* one subject with a clear title


.rubocop.yml
```diff
@@ -1,9 +1,11 @@
 require:
+  - rubocop-rspec
```
davidrunger, line 2: I'm pretty sure that the latest version of `rubocop-rspec` (currently 3.5.0) supports being loaded as a plugin. There is a "Pluginfy RuboCop RSpec" in the release notes [here](https://github.com/rubocop/rubocop-rspec/releases/tag/v3.5.0). Is there a reason that you are leaving `rubocop-rspec` under `require`, rather than loading it with the other `plugins`?

.rubocop_todo.yml
```diff
@@ -1,16 +1,115 @@
...
 # versions of RuboCop, may require this file to be generated again.
 
-# It may be possible for us to use safe_load, but we'd have to pass the
-# safelists, like `whitelist_classes` into our serializer, and the serializer
-# interface is a public API, so that would be a breaking change.
+# Offense count: 23
+# Configuration parameters: EnforcedStyle, AllowedGems, Include.
+# SupportedStyles: Gemfile, gems.rb, gemspec
+# Include: **/*.gemspec, **/Gemfile, **/gems.rb
+Gemspec/DevelopmentDependencies:
```
fatkodima, line 13: I think worth fixing all the new offences in this file or disabling some cops if they aren't preferable.

.rubocop.yml
modosc: yes - it requires updating `rspec-rails` to a version that's incompatible with `rails-6.1`.

.rubocop_todo.yml
modosc: sure, any preference here? i was trying to keep this as safe as possible but i'm happy to fix what i can (or just disable the rules).

modosc: i just re-ran the unsafe fixes on everything under `spec` since that should be relatively safe to do. will look at the non-auto-fixable items next

modosc: ok - i've fixed or ignored everything _except_ the last two. both of these could have consequences beyond what our test suite covers so i'd suggest we defer those to separate prs?

Rakefile
```diff
@@ -8,40 +8,40 @@ Bundler::GemHelper.install_tasks
 
 desc "Copy the database.DB.yml per ENV['DB']"
 task :install_database_yml do
-  puts format("installing database.yml for %s", ENV["DB"])
+  puts format("installing database.yml for %s", ENV.fetch("DB", nil))
```
fatkodima, line 11: I think the cop for these ugly `ENV.fetch(..., nil)` should be disabled.

.rubocop_todo.yml
```diff
@@ -1,23 +1,19 @@
...
-    - 'spec/models/no_object_spec.rb'
-    - 'spec/models/person_spec.rb'
-    - 'spec/models/version_spec.rb'
-    - 'spec/paper_trail/events/destroy_spec.rb'
-    - 'spec/paper_trail/serializer_spec.rb'
+    - 'lib/paper_trail/frameworks/active_record.rb'
+
+# Offense count: 1
+# This cop supports unsafe autocorrection (--autocorrect-all).
+Rails/ApplicationRecord:
```
fatkodima, line 17: This should be not in the todo file, because its not something that should be done (fixed). Can be disabled just inline in that file.

.rubocop.yml
```diff
@@ -25,6 +27,14 @@ AllCops:
   # See "Lowest supported ruby version" in CONTRIBUTING.md
   TargetRubyVersion: 3.0
 
+Gemspec/DevelopmentDependencies:
+  Enabled: false
+
+# skipped for now until we've audited all the gem maintainers but we should
+# enable this asap
+Gemspec/RequireMFA:
```
fatkodima, line 35: Can be enabled - maintainers already required to have MFA on rubygems

spec/controllers/articles_controller_spec.rb
```diff
@@ -10,19 +10,19 @@
       after { PaperTrail.enabled = false }
 
       it "returns true" do
-        expect(PaperTrail.enabled?).to eq(true)
+        expect(PaperTrail.enabled?).to be(true)
```
fatkodima, line 13: I would like the cop for these confusing `eq` -> `be` to be disabled. I doubt there is any useful difference between them or that people remember which one to use.

.rubocop.yml
```diff
@@ -140,6 +150,17 @@ RSpec/ExampleLength:
 RSpec/MultipleExpectations:
   Enabled: false
 
+RSpec/Rails/InferredSpecType:
```
fatkodima, line 153: Looks useless for me, I think can be just disabled.

spec/dummy_app/app/models/person.rb
```diff
@@ -40,13 +40,9 @@ def load(value)
       end
     end
 
-    def dump(zone)
-      self.class.dump(zone)
-    end
+    delegate :dump, to: :class
```
fatkodima, line 43: Original code was more clear, imo, would prefer for the cop to be disabled.

spec/models/version_spec.rb
```diff
@@ -3,9 +3,9 @@
 require "spec_helper"
 require "support/shared_examples/queries"
 require "support/shared_examples/active_record_encryption"
-
+require "support/custom_object_changes_adapter"
```
davidrunger, line 6: Very small nit: probably add a blank line to separate the `require`s from the `module` declaration (as before)?

spec/support/shared_examples/queries.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "support/custom_object_changes_adapter"
```
davidrunger, line 3: Very small nit: probably add a blank line to separate the `require`s from the `RSpec.shared_examples` below?

davidrunger (commented):
This looks solid to me. I largely agree with @fatkodima 's comments about some of the cops that might be better to disable rather than to autocorrect and enforce.

modosc:
i'll make one more pass and try to revert the rules mentioned above. if it's not straightforward i'll have to abandon this attempt - i have very little time today to work on this and i don't want bike shedding to block the real reason i did this (#1514).

modosc:
ok, i've addressed + fixed all the previous issues. if i don't hear anything else back in the next day i'll go ahead and merge.

Issue #163, opened 2012-06-21
Compatibility problems with optimistic locking / lock_version


saizai:
Suppose you have a Widget with a lock_version column.

If you try to do `widget.previous_version.save!`, you'll get a `ActiveRecord::StaleObjectError`.

I'm not sure how to override this. Ideally what should happen is that the newly reified object should have the same staleness / lock properties as the original object, so that if `widget` is OK to save then `widget.previous_version` should be too (since it hasn't changed out from under you).

Any ideas for how to implement this?

jaredbeck:
> The only workaround I have found to to change the lock_version attribute after reifying to the current lock_version and then saving the object. -Justin

I'm not sure I'd describe that as a workaround.  I haven't used [optimistic locking][1] personally, but based on the documentation it sounds like you're using it as intended; you want to update a record, so you must increment the lock_version.

I'm adding a note to the readme about this. (https://github.com/airblade/paper_trail/issues/550)

> I did: model.lock_version += 1 if model.respond_to? :lock_version in version_concern.rb, line 125 just before returning 'model' -Walther

I think this would work in a single-threaded app, but without concurrency control would not be sufficient in a multi-threaded app.

> I don't know if it makes sense to .. code that into PaperTrail .. -Ben

I agree.  If PaperTrail ever adds a high-level undo function, this is something we'll have to keep in mind.

[1]: http://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html

PR #1524, opened 2025-06-06
Support rails optimistic locking

yosefbennywidyo:
- Increment lock_version by 1 on the reified object if :lock_version attribute is present.
- Add a test for reifying an object with an optimistic lock.
- Update Rails version to 8.0.0.rc1 in Gemfile generated by Appraisal.
- Add CarWithLockVersion model and corresponding specs
- Rename Naming/PredicateName to Naming/PredicatePrefix in .rubocop.yml
- Fix offenses rubocop
- Update method name to follow predicate naming convention for controller
- Add entry for Rails optimistic locking support in CHANGELOG
- Fix method reference for enabling PaperTrail in controller
- Fix documentation typo in PaperTrail::Request for controller method reference
- Related to [#163 ](https://github.com/paper-trail-gem/paper_trail/issues/163)

Check the following boxes:

- [x] Added tests.


modosc:
based on @jaredbeck's [comment](https://github.com/paper-trail-gem/paper_trail/issues/163#issuecomment-114169503) in  #163 i'm not sure that this makes sense to merge (but maybe he can chime in?).

imho having the reified models fail this check by default seems like a good idea - this way the caller has to explicitly handle this situation when they're trying to save an old version of a model. also - how do you know that incrementing the lock by 1 is all that's required? if `lock_version` is serialized in your version you may be behind more than one. wouldn't you have to pull the current value out of the db and then increment it?

yosefbennywidyo:
Thank you for the feedback!

> imho having the reified models fail this check by default seems like a good idea - this way the caller has to explicitly handle this situation when they're trying to save an old version of a model.

I agree—making the check fail by default for reified models promotes safer handling when attempting to save an old version. This forces the caller to be intentional about dealing with optimistic locking scenarios, which should help prevent accidental overwrites.

> also - how do you know that incrementing the lock by 1 is all that's required? if `lock_version` is serialized in your version you may be behind more than one. wouldn't you have to pull the current value out of the db and then increment it?

Great point. I’ve updated the PR so that before saving, it pulls the current lock_version from the database and increments it, rather than just incrementing from the serialized version. This should ensure that we’re always using the latest value, even if multiple updates have been made since the version was created.

Let me know if you have other suggestions or if there’s a specific scenario you’d like me to cover!

modosc:
> I agree—making the check fail by default for reified models promotes safer handling when attempting to save an old version. This forces the caller to be intentional about dealing with optimistic locking scenarios, which should help prevent accidental overwrites.

sorry, i'm not going spend time responding to bots. 

this change makes things unsafe by default whenever optimistic locking is used and i don't think that's a great idea.

PR #1487, opened 2024-08-18
Add support for Postgres Range Types

sirwolfgang:
Thank you for your contribution!

Check the following boxes:

- [ ] Wrote [good commit messages][1].
- [ ] Feature branch is up-to-date with `master` (if not - rebase it).
- [ ] Squashed related commits together.
- [ ] Added tests.
- [ ] Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new 
  code introduces user-observable changes.
- [ ] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences.

[1]: http://chris.beams.io/posts/git-commit/

Issue #1486, opened 2024-08-15
Postgres Date Ranges not reify-ing correctly


sirwolfgang:
Thank you for your contribution!

Due to limited volunteers, issues that do not follow these instructions will be
closed without comment.

Check the following boxes:

- [x] This is not a usage question, this is a bug report
- [x] This bug can be reproduced with the script I provide below
- [x] This bug can be reproduced in the latest release of the `paper_trail` gem

Due to limited volunteers, we cannot answer *usage* questions. Please ask such
questions on [StackOverflow](https://stackoverflow.com/tags/paper-trail-gem).

Bug reports must use the following template:

```ruby
# frozen_string_literal: true

# Use this template to report PaperTrail bugs.
# Please include only the minimum code necessary to reproduce your issue.
require "bundler/inline"

# STEP ONE: What versions are you using?
gemfile(true) do
  ruby "3.3.0"
  source "https://rubygems.org"
  gem "activerecord", "7.1.3.4"
  gem "minitest", "5.11.3"
  gem "paper_trail", "15.1.0", require: false
  gem "pg", "1.5.7"
end

require "active_record"
require "minitest/autorun"
require "logger"

# Please use sqlite for your bug reports, if possible.
ActiveRecord::Base.establish_connection('postgres://postgres:password@localhost')
ActiveRecord::Base.connection.execute('DROP DATABASE IF EXISTS paper_trail_test')
ActiveRecord::Base.connection.execute('CREATE DATABASE paper_trail_test')
ActiveRecord::Base.establish_connection('postgres://postgres:password@localhost/paper_trail_test')
ActiveRecord::Base.logger = nil
ActiveRecord::Schema.define do
  # STEP TWO: Define your tables here.
  create_table :users, force: true do |t|
    t.string :name
    t.daterange :range
    t.timestamps null: false
  end

  create_table :versions do |t|
    t.string :item_type, null: false
    t.integer :item_id, null: false
    t.string :event, null: false
    t.string :whodunnit
    t.json :object
    t.json :object_changes
    t.datetime :created_at
  end
  add_index :versions, %i[item_type item_id]
end
ActiveRecord::Base.logger = Logger.new(STDOUT)
require "paper_trail"

# STEP FOUR: Define your AR models here.
class User < ActiveRecord::Base
  has_paper_trail
end

# STEP FIVE: Please write a test that demonstrates your issue.
class BugTest < ActiveSupport::TestCase
  def test_1
[...]

PR #1534, opened 2025-09-20
Add support for Postgres Range Types

sirwolfgang:
This Add/Fixes support for Postgres Range types: https://github.com/paper-trail-gem/paper_trail/issues/1486

Check the following boxes:

- [x] Added tests.

Related: https://github.com/paper-trail-gem/paper_trail/pull/1487

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,7 @@ recommendations of [keepachangelog.com](http://keepachangelog.com/).
 
 ### Fixed
 
-- None
+- Fixed errors when deserializing Range types from Ruby style strings to Postgres
 
 ## 15.1.0 (2023-10-22)
 
--- a/lib/paper_trail/attribute_serializers/attribute_serializer_factory.rb
+++ b/lib/paper_trail/attribute_serializers/attribute_serializer_factory.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "paper_trail/type_serializers/postgres_array_serializer"
+require "paper_trail/type_serializers/postgres_range_serializer"
 
 module PaperTrail
   module AttributeSerializers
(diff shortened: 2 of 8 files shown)


mattmenefee (commented):
Thanks for working on this, @sirwolfgang! This is a real bug that affects anyone using PostgreSQL range columns with PaperTrail's JSON version storage. The approach mirrors the existing `PostgresArraySerializer` pattern nicely, and the test coverage across all six range types is solid.

A few observations and suggestions:

## General Feedback

**The approach is sound** — routing range types through a custom serializer that understands Ruby-style string representations (`1..5`) rather than PostgreSQL bracket notation (`[1,5)`) is the right call, since PaperTrail's JSON columns store `to_s` output.

## Specific Suggestions

### `PostgresRangeSerializer#deserialize_with_ar` — edge cases

The `delimiter = string[/\.{2,3}/]` regex works correctly for well-formed Ruby range strings, but consider what happens if the input is somehow in PostgreSQL bracket format (e.g., `[2024-01-01,2024-01-31)`) — the delimiter match would return `nil`, and `string.split(nil)` would raise. A guard clause or fallback to the original Active Record deserializer might make this more robust:

```ruby
def deserialize_with_ar(string)
  return nil if string.blank?

  delimiter = string[/\.{2,3}/]
  return @active_record_serializer.deserialize(string) unless delimiter

  range_start, range_end = string.split(delimiter, 2)
  # ...
end
```

### `attributes_to_serialize` / `changes_to_serialize` — potential double-serialization

When a column is both encrypted _and_ a range type, it would appear in both `encrypted_to_serialize` and `columns_to_serialize`. Since the merge is `encrypted_to_serialize.merge(columns_to_serialize)`, the range version wins, which seems correct — but it's worth adding a brief comment noting this is intentional.

### Test: description typo

```ruby
# Line says "ruby array" but should say "ruby range"
it "serializes a postgres range into a ruby array" do
```

### CHANGELOG entry

The current wording — "Fixed errors when deserializing Range types from Ruby style strings to Postgres" — is a bit ambiguous about the direction. Consider:

```
- [Fixed] Support for PostgreSQL range column types in version object serialization
```

## Missing test coverage (nice-to-haves)

- `nil` range values
- Range with only one bound set (e.g., `1..` in Ruby 2.6+, or unbounded on one side in PostgreSQL)
- A full integration test that round-trips through `reify` (mirroring the reproduction script in #1486)

Overall this is good work and addresses a legitimate gap. Would love to see it merged!

sirwolfgang:
I know I need to update it, but why do we go through this every year github bot?

PR #1546, opened 2026-01-23
Fix spurious versions from touch events with no changes

### LINKED ISSUE #1465: Empty "update" versions when using active storage

closed without comment.

Check the following boxes:

I couldn't use the script template as I needs Rails to reproduce the bug. But I'm attaching a demo repo in which the bug reproduces.

1) Checkout this repository:
https://github.com/mochetts/paper-trail-active-storage-bug

2) Run `rails db:migrate` to run migrations.

3) Run `rake` to run minitests.

4) See that the test fails:

Screenshot 2024-03-13 at 7 34 08 AM

Some analysis I've been doing:

It seems that Rails touches records when storing a `has_one_attached` relationship as per [this](https://github.com/rails/rails/issues/45567) bug report and its consequent [fix](https://github.com/rails/rails/pull/45579).

This means that every time a new record is created with an associated attachment, it touches the model multiple times, generating multiple "update" versions for it with `nil` "object_changes". This seems to happen because the touch is occurring on creation, when there's no "updated_at" to touch.

Screenshot 2024-03-12 at 6 49 20 PM

Now the real question is if this is a Rails bug or a Papertrail bug, as it only happens under this scenario. Worthy to mention that rails forbids you from performing a `touch` on a new record, but under these circumstances it seems to be doing by itself.

Screenshot 2024-03-13 at 7 28 02 AM

I would suggest skipping the insertion of "update" versions when `object_changes = nil`, as these become noise in the versions history of the model.

I can send a PR for this, but I'd need confirmation that this is the right path.

mattmenefee:
## Summary

Fixes #1465 — spurious empty update versions being created when using ActiveStorage.

### The Problem

When `has_one_attached` (ActiveStorage) calls `touch` on the parent record after attaching a blob, Paper Trail creates an empty version with no meaningful `object_changes`. This happens because `Events::Update#changed_notably?` unconditionally returned `true` for touch events when `changes_in_latest_version` was empty.

**Before this fix:**
```ruby
def changed_notably?
  if @is_touch && changes_in_latest_version.empty?
    true  # Always creates a version, even with no changes
  else
    super
  end
end
```

### The Fix

By removing this override, we delegate to the base class implementation which correctly returns `false` when there are no notable changes tracked.

**After this fix:**
- Touch events that update `updated_at` → version IS created (because Rails tracks the change)
- Touch events with no tracked changes → version is NOT created (fixes spurious versions)

### Breaking Change Notice

This is technically a breaking change for users who rely on the old behavior of always creating versions for touch events. If the maintainers prefer a backward-compatible approach, an alternative would be to add an opt-in configuration option:

```ruby
# Alternative approach: add a configuration option
has_paper_trail skip_touch_with_no_changes: true

# Or globally:
PaperTrail.config.has_paper_trail_defaults = {
  skip_touch_with_no_changes: true
}
```

The implementation for the opt-in approach would be:
```ruby
def changed_notably?
  if @is_touch && changes_in_latest_version.empty?
    !@record.paper_trail_options[:skip_touch_with_no_changes]
  else
    super
  end
end
```

Happy to implement this alternative if preferred.

## Test Plan

- [ ] Existing test suite passes (CI will verify)
- [ ] Tests that expect `touch` to create versions may need updating if Rails doesn't track the `updated_at` change
- [ ] Manual testing confirms ActiveStorage attachments no longer create spurious versions

## Related

- Fixes #1465
- Related Rails commit: https://github.com/rails/rails/commit/dcb825902d79d0f6baba956f7c6ec5767611353e

---

## Contribution Checklist

- [x] Wrote [good commit messages](http://chris.beams.io/posts/git-commit/).
- [ ] Added tests.

### Notes on unchecked items

- **Up-to-date with master**: Will rebase if requested by maintainers.
- **Tests**: The existing touch-related tests in the suite cover this behavior. The change removes an override that forced version creation; the base class logic (already tested) now handles touch events consistently. Happy to add a dedicated regression test for the ActiveStorage scenario if desired.
- **Changelog**: This is a behavior change that warrants a changelog entry. Holding off until maintainers confirm whether this should land as a breaking change or behind a configuration option, which would affect the changelog section it belongs in.

### DIFF
--- a/lib/paper_trail/events/update.rb
+++ b/lib/paper_trail/events/update.rb
@@ -37,19 +37,16 @@ module PaperTrail
         merge_metadata_into(data)
       end
 
-      # If it is a touch event, and changed are empty, it is assumed to be
-      # implicit `touch` mutation, and will a version is created.
+      # NOTE: We previously overrode `changed_notably?` here to return `true`
+      # for touch events even when `changes_in_latest_version` was empty.
+      # This caused spurious versions when used with ActiveStorage, which
+      # calls `touch` after attaching blobs without making meaningful changes.
       #
+      # By removing the override, we now delegate to the base class, which
+      # only creates versions when there are actual notable changes.
+      #
+      # See https://github.com/paper-trail-gem/paper_trail/issues/1465
       # See https://github.com/rails/rails/commit/dcb825902d79d0f6baba956f7c6ec5767611353e
-      #
-      # @api private
-      def changed_notably?
-        if @is_touch && changes_in_latest_version.empty?
-          true
-        else
-          super
-        end
-      end
 
       private
 


lib/paper_trail/events/update.rb
```diff
@@ -37,19 +37,16 @@ def data
         merge_metadata_into(data)
       end
 
-      # If it is a touch event, and changed are empty, it is assumed to be
-      # implicit `touch` mutation, and will a version is created.
+      # NOTE: We previously overrode `changed_notably?` here to return `true`
```
mattmenefee, line 40: It's probably not necessary to retain this comment, so I can remove it after the changes are approved and before this is merged.

